Issue #188, opened 2023-02-02
[Feature Request] Adding additional options to "dynatrace_service" data source 


98christianp:
**Is your feature request related to a problem? Please describe.**
Hello guys, I am starting to use the `dynatrace_service` data source and have found it slightly underwhelming, in the sense that the query options are quite limited. 

**Describe the solution you'd like**
To make it more useful I'd like the option to add the `Operator` field i.e when querying by name we can select from various options; `CONTAINS` `BEINGS_WITH`  `EQUALS` etc.

Another nice thing to have would be the option to filter by the the serviceType field that is available in the API schema.

Dynatrace-Reinhard-Pilz:
Hello @98christianp 
You're right, the current data source doesn't offer a lot of functionality. We've introduced it for a very specific use case (migrating settings from one environment to another), and here matching against tags was sufficient enough for a start.

But we ARE working on functionality that's a bit more flexible. The REST endpoint
`https://#####.live.dynatrace.com/rest-api-doc/index.jsp?urls.primaryName=Environment%20API%20v2#/Monitored%20entities/getEntities` allows for specifying a `metricSelector` and for choosing the fields you'd like to get. Eventually this would allow you to narrow down results already via `metricSelector` and later on filter using HCL functionality.

98christianp:
> Hello @98christianp You're right, the current data source doesn't offer a lot of functionality. We've introduced it for a very specific use case (migrating settings from one environment to another), and here matching against tags was sufficient enough for a start.
> 
> But we ARE working on functionality that's a bit more flexible. The REST endpoint `https://#####.live.dynatrace.com/rest-api-doc/index.jsp?urls.primaryName=Environment%20API%20v2#/Monitored%20entities/getEntities` allows for specifying a `entitySelector` and for choosing the fields you'd like to get. Eventually this would allow you to narrow down results already via `entitySelector` and later on filter using HCL functionality.

Sounds promsing! To align my understanding of what you are saying;  you will add the `entitySelector` param that is available in the API to the data sources? 🤔

Dynatrace-Reinhard-Pilz:
Exactly. We "kind" of have that already under the hoods, but not explicitly.
But we were internally already talking about offering the `entitySelector` as an input property and offering a fixed set of properties as results (depending on what kind of entity some of them would be populated, some of them would be empty).

98christianp:
Sweet, I'll close this for now as I got what I need in the release, thank you 😀

PR #189, opened 2023-02-02
Add additional operator to data source "dynatrace_service"

ceng-p:
Hello guys,

I've done my best attempt at adding the operator field to the data source fields, currently only adding `CONTAINS` and keeping EQUALS as the default, allowing backwards compatability with how it was.

I'm very much open for feedback or suggestions. 

Adresses some of #188


Dynatrace-Reinhard-Pilz:
Thanks for contributing!

PR #203, opened 2023-02-27
Update credentials.md with token details

paulmorar:
Perhaps instead of this change, I would re-write the entire intro page and separate the export documentation to avoid confusion and help readers focus on the topic. I don't have the overview necessary to provide the valuable info there. But I might return with an iteration 🙂

### DIFF
--- a/docs/resources/credentials.md
+++ b/docs/resources/credentials.md
@@ -8,7 +8,7 @@ description: |-
 
 # dynatrace_credentials (Resource)
 
-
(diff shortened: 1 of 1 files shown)


Dynatrace-Reinhard-Pilz:
Hello Paul,
Any improvement regarding documentation would be very welcome. 
If you have further ideas here, don't hesitate to create additional Pull Requests. We will make sure that your changes are getting picked up properly when generating the documentation.

PR #482, opened 2024-06-21
feat: create public documents

warber:
This PR adds a `private`field to the document resource to be able to create public documents.

Creating "public" document require (per design) two API calls. 
* First, a private document needs to be created (using the HTTP `POST` endpoint) 
* Second, the crated document needs to be updated (using the HTTP `PATCH` endpoint) to be public.

### DIFF
--- a/dynatrace/api/documents/document/settings/document.go
+++ b/dynatrace/api/documents/document/settings/document.go
@@ -29,6 +29,7 @@ const SchemaVersion = 3
 type Document struct {
 	Name          string `json:"name" maxlength:"200"`
 	Content       string `json:"content,omitempty"`
+	IsPrivate     bool   `json:"isPrivate,omitempty"`
 	Type          string `json:"type"`
 	Actor         string `json:"actor,omitempty" maxlength:"36" format:"uuid"`
 	Owner         string `json:"owner,omitempty" format:"uuid"`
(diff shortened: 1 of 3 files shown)


dynatrace/api/documents/document/settings/document.go
```diff
@@ -44,6 +45,12 @@ func (me *Document) Schema() map[string]*schema.Schema {
 			Required:         true,
 			ValidateDiagFunc: ValidateMaxLength(200),
 		},
+		"private": {
+			Type:        schema.TypeBool,
+			Description: "Specifies whether the document is private or readable by everybody",
+			Optional:    true,
+			Default:     false,
```
arthurpitman, line 52: Considering documents have been private up to now, wondering if this should be `true` to preserve this for any existing documents, or if we need to migrate state somehow. I think we ultimately want documents to be public by default though, correct?

Dynatrace-Reinhard-Pilz: Had similar thoughts. My idea was that we might want to call the attribute `public` and let it default to `false` (essentially because it's more intuitive for users that an absent `boolean` defaults to `false`.
Will reach out to you guys on Monday - most likely I'll come up with fifty other ideas.

Dynatrace-Reinhard-Pilz: Thought about Arthurs comment over the weekend. In the end customers expect to see newly provisioned dashboards to show up for other users - not for the service user they're authenticating with using Terraform.
In order words, like Arthur said, we (and customers) want documents to be public by default anyways. I expect that customers will welcome that we're flagging any existing documents as `public`.

Issue #592, opened 2024-11-18
Platform buckets deletion fails


vlad-preutu:
**Describe the bug**
When attempting to delete a Log Bucket (dynatrace_platform_bucket), the provider always fails with a 404 error.
This is a bug because the provider retries until it gets a 404 and then returns the error.

**To Reproduce**
1. Using the terraform provider, create a resource of type "dynatrace_platform_bucket".
2. Plan and Apply.
3. Remove the resource from terraform.
4. The plan will work and it will say that it is planning to delete the resource (destroy  module.abs.dynatrace_platform_bucket.example)
5. The apply will fail after some retries with the error:
│ Error: API request HTTP GET https://teantID.apps.dynatrace.com/platform/storage/management/v1/bucket-definitions/example failed with status code 404: {"error":{"code":404,"message":"Not Found","errorDetails":[]}}

**Expected behavior**
A status code of 404 - Not found on a GET request for the Log bucket is considered a successful deletion.

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

Reinhard-Pilz-Dynatrace:
Hello @vlad-preutu, I just merged your Pull Request. Thank you for contributing the fix!
We're planning to push our a new release on Dec 5. Your fix will be contained in there.

PR #593, opened 2024-11-18
fix: Fixed platform bucket deletion failing - issue #592

vlad-preutu:
Fixed platform bucket delete failing when GET request returns 404.


Reinhard-Pilz-Dynatrace:
Thank you for contributing!
Coincidentally we DID have failing integration tests for buckets. Turns out that the issue you've found was the root cause for that.

PR #679, opened 2025-05-03
Fix the schema definition after the type change

paulmorar:
I have been too quick to ship an attempt at fixing the schema, and there are no tests for this particular component that could catch the issue.

The error when running the provider clearly states `panic: interface conversion: interface {} is int, not float64` which indicates that there is a misalignment between the schema and the struct.

### DIFF
--- a/dynatrace/api/v2/synthetic/monitors/settings/thresholds.go
+++ b/dynatrace/api/v2/synthetic/monitors/settings/thresholds.go
@@ -75,7 +75,7 @@ func (me *Threshold) Schema() map[string]*schema.Schema {
 			Optional:    true,
 		},
 		"threshold": {
-			Type:        schema.TypeInt,
+			Type:        schema.TypeFloat,
 			Description: "Notify if monitor request takes longer than X milliseconds to execute",
 			Optional:    true,
 		},


paulmorar:
@kishikawa12 Seems I have been too quick on fixing the struct for the threshold without adjusting the schema.

kishikawa12:
Hi @paulmorar,

Thanks for the find and apologies that I missed that on my end. I blame that it was a Friday. :) 

Unfortunately while I was validating this resource, I found another potential issue. When I enabled the tests, I noticed that we were leading to non-empty plans on configured tags due to the cluster taking a long time to propagate. Will be working on a fix on that this afternoon and will publish a minor release.

paulmorar:
I would like to familiarize myself better with the code base and perhaps catch these issues myself.

But the tests are running against an actual tenant as far as I could understand from the comments. Is that tenant public?

kishikawa12:
The tests for this particular resource seemed to have been disabled since it was initially not available on our test tenant, unfortunately the tenant is not publicly available.

One item I forgot to mention while I was testing the `threshold` attribute this morning, the double/float type definition is a bit misleading. It really only accepts whole numbers, eg. 99.0, 200.0, etc. The GUI also implies this by only allowing an integer to be configured (which I believe where the type definition initially came from). Just an FYI, configuring numbers like 200.5, would be accepted but would lead to non-empty plans.

Great to hear you're possibly interested in contributing further, @Reinhard-Pilz-Dynatrace and I would be happy to have an overview session with you to review the code. If you're interested, email us at terraform@dynatrace.com. 👍

paulmorar:
For us, we configured numbers like 100, and suddenly the API started to return 100.0 in the response. 

We changed nothing at all and suddenly, when trying to introduce a small change via terraform, this completely broke as the types would not match. We are using exclusively terraform to manage configuration and also have built self service for teams to create a Dynatrace "Landing-Zone". This is why the provider is quite important for us.

That being said, I did some state gymnastics meanwhile so we get configuration out while waiting for the fix on the network_monitor resource.

PR #731, opened 2025-07-21
feat: Added `SettingPermissions` struct and HCL schema

d0weinberger:
This PR introduces the structure of the `dynatrace_setting_permissions` resource and its HCL representation. Accordingly, `Schema()`, `MarshalHCL()` and `UnmarshalHCL()` functions are implemented, and an example config is given in `example_a.tf`.

In the following example, a permission resource is defined for the settings object ID of another Terraform resource `example_connection` of type `dynatrace_github_connection`.

It specifically grants **no access** to arbitrary users.
The user `user_id_1` gets write access and the user `user_id_2` gets read access.
Additionally, users of group `group_id_2` get write access.

Permissions for groups and users not defined in the resource will be **revoked**!

```
resource "dynatrace_setting_permissions" "github_connection_access" {
  settings_object_id = dynatrace_github_connection.example_connection.id
  all_users = "none" # possible values: none, read, write - default: none
  users { # default: empty
    user {
        user_id = "user_id_1"
        access = "write" # possible values: read, write
    }
    user {
        user_id = "user_id_2"
        access = "read" # possible values: read, write
    }
  }
  groups { # default: empty
    group {
        group_id = "group_id_2"
        access = "write" # possible values: read, write
    }
  }
}
```

### DIFF
--- a/dynatrace/api/v2/settings/objects/permissions/settings/group_accessor.go
+++ b/dynatrace/api/v2/settings/objects/permissions/settings/group_accessor.go
@@ -31,7 +31,7 @@ func (_ *Groups) Schema() map[string]*schema.Schema {
 			Type:        schema.TypeList,
 			Required:    true,
 			MinItems:    1,
-			Description: "",
+			Description: "Group that is to be granted read or write permissions",
 			Elem:        &schema.Resource{Schema: new(GroupAccessor).Schema()},
 		},
 	}
--- a/dynatrace/api/v2/settings/objects/permissions/settings/group_accessor.go
+++ b/dynatrace/api/v2/settings/objects/permissions/settings/group_accessor.go
@@ -28,7 +28,7 @@ type Groups []*GroupAccessor
 func (_ *Groups) Schema() map[string]*schema.Schema {
 	return map[string]*schema.Schema{
 		"group": {
-			Type:        schema.TypeList,
+			Type:        schema.TypeSet,
 			Required:    true,
 			MinItems:    1,
 			Description: "Group that is to be granted read or write permissions",
--- a/dynatrace/api/v2/settings/objects/permissions/settings/group_accessor.go
+++ b/dynatrace/api/v2/settings/objects/permissions/settings/group_accessor.go
@@ -46,13 +46,13 @@ func (g *Groups) UnmarshalHCL(decoder hcl.Decoder) error {
 }
 
 type GroupAccessor struct {
-	GroupID string
-	Access  string
+	ID     string
+	Access string
 }
 
 func (_ *GroupAccessor) Schema() map[string]*schema.Schema {
 	return map[string]*schema.Schema{
-		"group_id": {
+		"id": {
 			Type:        schema.TypeString,
 			Description: "The UUID of the group, conveniently retrieved via the `id` attribute provided by the data source `dynatrace_iam_group",
 			Required:    true,
@@ -68,14 +68,14 @@ func (_ *GroupAccessor) Schema() map[string]*schema.Schema {
 
 func (ga *GroupAccessor) MarshalHCL(properties hcl.Properties) error {
 	return properties.EncodeAll(map[string]any{
-		"group_id": ga.GroupID,
-		"access":   ga.Access,
+		"id":     ga.ID,
+		"access": ga.Access,
 	})
 }
 
 func (ga *GroupAccessor) UnmarshalHCL(decoder hcl.Decoder) error {
 	return decoder.DecodeAll(map[string]any{
-		"group_id": &ga.GroupID,
-		"access":   &ga.Access,
+		"id":     &ga.ID,
+		"access": &ga.Access,
 	})
 }
--- a/dynatrace/api/v2/settings/objects/permissions/settings/group_accessor.go
+++ b/dynatrace/api/v2/settings/objects/permissions/settings/group_accessor.go
@@ -1,0 +1,81 @@
+/**
+* @license
+* Copyright 2025 Dynatrace LLC
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+ */
+
+package settings
+
+import (
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/terraform/hcl"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+type Groups []*GroupAccessor
+
+func (_ *Groups) Schema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"group": {
+			Type:        schema.TypeList,
+			Required:    true,
+			MinItems:    1,
+			Description: "",
+			Elem:        &schema.Resource{Schema: new(GroupAccessor).Schema()},
+		},
+	}
+}
+
+func (g *Groups) MarshalHCL(properties hcl.Properties) error {
+	return properties.EncodeSlice("group", g)
+}
+
+func (g *Groups) UnmarshalHCL(decoder hcl.Decoder) error {
+	return decoder.DecodeSlice("group", g)
+}
+
+type GroupAccessor struct {
+	GroupID string
+	Access  string
+}
+
+func (_ *GroupAccessor) Schema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"group_id": {
+			Type:        schema.TypeString,
+			Description: "The UUID of the group, conveniently retrieved via the `id` attribute provided by the data source `dynatrace_iam_group",
+			Required:    true,
+		},
+		"access": {
+			Type:         schema.TypeString,
+			Description:  "Valid values: read, write",
+			Required:     true,
+			ValidateFunc: validation.StringInSlice([]string{"read", "write"}, false),
+		},
+	}
+}
+
+func (ga *GroupAccessor) MarshalHCL(properties hcl.Properties) error {
+	return properties.EncodeAll(map[string]any{
+		"group_id": ga.GroupID,
+		"access":   ga.Access,
+	})
+}
+
+func (ga *GroupAccessor) UnmarshalHCL(decoder hcl.Decoder) error {
+	return decoder.DecodeAll(map[string]any{
+		"group_id": &ga.GroupID,
+		"access":   &ga.Access,
+	})
+}
--- a/dynatrace/api/v2/settings/objects/permissions/settings/setting_permissions.go
+++ b/dynatrace/api/v2/settings/objects/permissions/settings/setting_permissions.go
@@ -33,11 +33,13 @@ type SettingPermissions struct {
 func (_ *SettingPermissions) Schema() map[string]*schema.Schema {
 	return map[string]*schema.Schema{
 		"settings_object_id": {
-			Type:     schema.TypeString,
-			Required: true,
+			Type:        schema.TypeString,
+			Description: "The ID of the settings object for which access is to be granted. Here, you can use the `id` attribute of the respective settings object resource",
+			Required:    true,
 		},
 		"all_users": {
 			Type:         schema.TypeString,
+			Description:  "Defines the default access level granted to all users in this environment. Allowed values are `read`, `write`, or `none`",
 			Optional:     true,
 			ValidateFunc: validation.StringInSlice([]string{"read", "write", "none"}, false),
 		},
--- a/dynatrace/api/v2/settings/objects/permissions/settings/setting_permissions.go
+++ b/dynatrace/api/v2/settings/objects/permissions/settings/setting_permissions.go
@@ -44,11 +44,13 @@ func (_ *SettingPermissions) Schema() map[string]*schema.Schema {
 		"users": {
 			Type:     schema.TypeList,
 			Optional: true,
+			MaxItems: 1,
 			Elem:     &schema.Resource{Schema: new(Users).Schema()},
 		},
 		"groups": {
 			Type:     schema.TypeList,
 			Optional: true,
+			MaxItems: 1,
 			Elem:     &schema.Resource{Schema: new(Groups).Schema()},
 		},
 	}
--- a/dynatrace/api/v2/settings/objects/permissions/settings/setting_permissions.go
+++ b/dynatrace/api/v2/settings/objects/permissions/settings/setting_permissions.go
@@ -1,0 +1,73 @@
+/**
+* @license
+* Copyright 2025 Dynatrace LLC
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+ */
+
+package settings
+
+import (
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/terraform/hcl"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+type SettingPermissions struct {
+	SettingsObjectID string
+	AllUsers         string
+	Users            Users
+	Groups           Groups
+}
+
+func (_ *SettingPermissions) Schema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"settings_object_id": {
+			Type:     schema.TypeString,
+			Required: true,
+		},
+		"all_users": {
+			Type:         schema.TypeString,
+			Optional:     true,
+			ValidateFunc: validation.StringInSlice([]string{"read", "write", "none"}, false),
+		},
+		"users": {
+			Type:     schema.TypeList,
+			Optional: true,
+			Elem:     &schema.Resource{Schema: new(Users).Schema()},
+		},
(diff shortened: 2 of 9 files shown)


dynatrace/api/v2/settings/objects/permissions/settings/owner_based_access_control.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/terraform/hcl"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+type OwnerBasedAccessControl struct {
+	SettingsObjectID string `json:"settings_object_id"`
+	AllUsers         string `json:"all_users"`
+	Users            Users  `json:"users"`
+	Groups           Groups `json:"groups"`
```
Kirdock: Do we even need JSON tags for this one? We won't marshal this one as these are separate endpoints right?

dynatrace/api/v2/settings/objects/permissions/settings/user.go
```diff
@@ -0,0 +1,79 @@
...
+	return properties.EncodeSlice("user", g)
+}
+
+func (g *Users) UnmarshalHCL(decoder hcl.Decoder) error {
+	return decoder.DecodeSlice("user", g)
+}
+
+type UserAccessor struct {
+	UserID string `json:"user_id"`
+	Access string `json:"access"`
```
Kirdock: Is there something else I'm not seeing at the moment (mapper function or another struct)?
I can't see the direct mapping from HCL to the API, which should be part of CA-15663, right?

This resource should be marshalled to the API request, which is
```json
{
  "accessor": {
    "id": "abc-xy",
    "type": "user"
  },
  "permissions": [
    "r"
  ]
}
```

dynatrace/api/v2/settings/objects/permissions/settings/group.go
```diff
@@ -0,0 +1,79 @@
...
+	return map[string]*schema.Schema{
+		"group_id": {
+			Type:        schema.TypeString,
+			Description: "No documentation available",
+			Required:    true,
+		},
+		"access": {
+			Type:        schema.TypeString,
+			Description: "Valid values: read, write",
+			Required:    true,
```
Kirdock: Should we add a validation function here and also for other resources?
```suggestion
			Required:    true,
			ValidateFunc: validation.StringInSlice([]string{"read", "write"}, false),
```

dynatrace/api/v2/settings/objects/permissions/testdata/terraform/example_a.tf
```diff
@@ -0,0 +1,20 @@
...
+        access = "read"
+    }
+  }
+  groups {
+    group {
+        group_id = "group_id_2"
+        access = "write"
+    }
+  }
+}
```
tomazpu: New line missing at end of file

dynatrace/api/v2/settings/objects/permissions/testdata/terraform/example_a.tf
```diff
@@ -0,0 +1,20 @@
+resource "dynatrace_owner_based_access_control" "ABC" {
+  settings_object_id = "asdf"
```
tomazpu: If this value is in the documentation, I would prefer something like `"#id#"`

dynatrace/api/v2/settings/objects/permissions/testdata/terraform/example_a.tf
```diff
@@ -0,0 +1,20 @@
+resource "dynatrace_owner_based_access_control" "ABC" {
```
tomazpu: Instead of "ABC" i would prefer "#name#"

dynatrace/api/v2/settings/objects/permissions/settings/group.go
```diff
@@ -0,0 +1,79 @@
+/**
```
tomazpu: I would probably prefer the file be named `group_accessor` to prevent confusion about the group's undescriptiveness.

dynatrace/api/v2/settings/objects/permissions/testdata/terraform/example_a.tf
Laubi: Or we could reference some other fake object to show how we'd recommend using it?

d0weinberger: I would like to use `dynatrace_github_connection.example_connection.id` but I am not sure "id" is the id of the settings object here.

EDIT: Okay, I'm a bit more confident now that this works: https://github.com/dynatrace-oss/terraform-provider-dynatrace/blob/3202ca45f15bcb7869a19ae2b6728de0893d7285/dynatrace/settings/services/settings20/service.go#L463

dynatrace/api/v2/settings/objects/permissions/testdata/terraform/example_a.tf
d0weinberger: I will name it `github_connection_access`

dynatrace/api/v2/settings/objects/permissions/settings/group.go
d0weinberger: I will rename `user` to `user_accessor` as well

dynatrace/api/v2/settings/objects/permissions/settings/group.go
d0weinberger: Validation makes sense, yes. I will add that

dynatrace/api/v2/settings/objects/permissions/settings/user.go
d0weinberger: Oh, that is not yet done, no. This is really just the struct representation of the HCL.

dynatrace/api/v2/settings/objects/permissions/settings/owner_based_access_control.go
d0weinberger: I guess you're right. We could get rid of the tags. The structs which will be marshalled aren't defined yet.

d0weinberger: Done

dynatrace/api/v2/settings/objects/permissions/settings/group.go
d0weinberger: Done

dynatrace/api/v2/settings/objects/permissions/testdata/terraform/example_a.tf
d0weinberger: Done

dynatrace/api/v2/settings/objects/permissions/testdata/terraform/example_a.tf
d0weinberger: Done

dynatrace/api/v2/settings/objects/permissions/testdata/terraform/example_a.tf
d0weinberger: Done

dynatrace/api/v2/settings/objects/permissions/settings/group.go
d0weinberger: Done

dynatrace/api/v2/settings/objects/permissions/testdata/terraform/example_a.tf
```diff
@@ -0,0 +1,20 @@
+resource "dynatrace_setting_permissions" "github_connection_access" {
+  settings_object_id = dynatrace_github_connection.example_connection.id
+  all_users = "write"
```
Kirdock: [nit] should we change it to `read` for this one? Not sure if this is still the case but `allUsers: write` would override/delete the others, right?
Or we could change it in a way so that all the other arguments here make sense? Not sure which accessor type has which priority, but I guess the most powerful one is taken?

dynatrace/api/v2/settings/objects/permissions/settings/setting_permissions.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func (_ *SettingPermissions) Schema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"settings_object_id": {
+			Type:     schema.TypeString,
+			Required: true,
+		},
+		"all_users": {
+			Type:         schema.TypeString,
```
Kirdock, line 41: At the moment, the description is a bit inconsistent. Some have one, some haven't, and for some it's `No documentation available`. Should we improve this here?

d0weinberger: I'll mention that it's the UUID in Dynatrace and maybe add an example to use the data source

dynatrace/api/v2/settings/objects/permissions/settings/setting_permissions.go
```diff
@@ -0,0 +1,73 @@
...
+
+type SettingPermissions struct {
+	SettingsObjectID string
+	AllUsers         string
+	Users            Users
+	Groups           Groups
+}
+
+func (_ *SettingPermissions) Schema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
```
Kirdock, line 34: We should also add descriptions for these arguments.
I'm not sure about the group arguments, though (`groups` and `users`)

dynatrace/api/v2/settings/objects/permissions/testdata/terraform/example_a.tf
```diff
@@ -0,0 +1,32 @@
+data "dynatrace_iam_user" "user_a" {
+  email = "a@home.com"
```
Kirdock: Maybe we should change this one to `example.com` or even `dynatrace.com`. We should not forward to any other sites (`home.com` redirects to `fairway.com`)

d0weinberger: Changed this and several other occurrences to `example.com`

dynatrace/api/v2/settings/objects/permissions/settings/setting_permissions.go
d0weinberger: Done

dynatrace/api/v2/settings/objects/permissions/settings/user_accessor.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+type UserAccessor struct {
+	UserID string
+	Access string
+}
+
+func (_ *UserAccessor) Schema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"user_id": {
```
Kirdock: Sry, for the late feedback: Thinking about it again, should we just call it `user` or `user_uid`? The API and our resource return it as `uid`.

When looking at other resources like `policy_binding`, we don't attach any identifier to the arguments and just call them `environment/account/policies`, but I also see that this is a bit inconsistent (different for management zones, dashboard, and application_id).

On the other hand, when comparing different providers, like AWS, they tend to use `_id`.

We could also leave it like that, but I would first like to know the others' opinions.

dynatrace/api/v2/settings/objects/permissions/settings/setting_permissions.go
```diff
@@ -0,0 +1,77 @@
...
+		"settings_object_id": {
```
Kirdock, line 35: Also thinking here again, we should probably discuss this a bit more: As a Terraform user, do I know which settings objects can actually be used here, or what settings objects are in general?
For example, we have the `dynatrace_site_reliability_guardian`, which is in the `Automation` section in the docs. Then we have the `dynatrace_github_connection`, which is part of `Connectors`.
Both of them are settings objects, but is it enough if the documentation of them just links to settings and refers to settings permissions?

Also, other settings schemas may be currently in use, and not all of them support defining permissions.

dynatrace/api/v2/settings/objects/permissions/settings/user_accessor.go
d0weinberger: We could also drop the `user` prefix as it it in a `user` block anyway.
Maybe just `id`? With `uid` it would be a bit weird compared with groups, because from the group data source, you would use the `id`. There is no `uid`. But, if we wanted to align it with the datasources, it could potentially be:

```
user {
  uid = data.dynatrace_iam_user.user_a.uid
  access = "write"
}
```
and 

```
group {
  id = data.dynatrace_iam_group.group_b.id
  access = "write"
}
```

dynatrace/api/v2/settings/objects/permissions/settings/setting_permissions.go
d0weinberger: I get your point. We could add to the description of "settings_object_id" something like `"To understand which resources are supported, check <insert docs here>"` 

Regarding the mentioning of settings objects ... Yeah, that one is interesting. If we only call it a resource without the context that under the hood it's a settings object, then we are also not talking about settings permissions here, but "resouce permissions" maybe?

dynatrace/api/v2/settings/objects/permissions/settings/user_accessor.go
Kirdock: We decided to go with `user_uid`, because we also have the user resource, which returns the `uid` attribute.

dynatrace/api/v2/settings/objects/permissions/settings/setting_permissions.go
Kirdock: Follow up here: We should have good documentation that tells which resource (settings) can be used here.

dynatrace/api/v2/settings/objects/permissions/settings/user_accessor.go
d0weinberger: Okay. Let's wait for other opinions

Kirdock: Good point. I would go for just `uid`/`id` as the rest is just duplicated information from the parent block

dynatrace/api/v2/settings/objects/permissions/settings/user_accessor.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+type UserAccessor struct {
+	UID    string
+	Access string
+}
+
+func (_ *UserAccessor) Schema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"id": {
```
Kirdock: Not sure if this will lead to confusion, if the user resource has `uid` and this one is `id`, whereas in the description we call it again `UUID`. The API endpoint returns for both (user and group) `uid`

d0weinberger: > The API endpoint returns for both (user and group) uid

Okay, I was focusing on the TF resources and `dynatrace_iam_group` has no uid attribute, therefore i chose "id" there

Kirdock (approved):

dynatrace/api/v2/settings/objects/permissions/testdata/terraform/example_a.tf
```diff
@@ -0,0 +1,32 @@
...
+data "dynatrace_iam_user" "user_b" {
+  email = "b@example.com"
+}
+
+data "dynatrace_iam_group" "example_group" {
+  name = "Terraform Example"
+}
+
+resource "dynatrace_setting_permissions" "github_connection_access" {
+  settings_object_id = dynatrace_github_connection.example_connection.id
```
arthurpitman, line 20: Should the example be fully self-contained, i.e., also provide the `dynatrace_github_connection` we reference here?

dynatrace/api/v2/settings/objects/permissions/settings/setting_permissions.go
arthurpitman, line 35: nit: We write `settings_object_id` but call our resource `dynatrace_setting_permissions`. I wonder if `dynatrace_settings_permissions` makes more sense (also considering the underlying API)? If need be we can still change this, so this is not something blocking this PR.

dynatrace/api/v2/settings/objects/permissions/testdata/terraform/example_a.tf
d0weinberger: Hm, I could of course add it.

EDIT: added it

dynatrace/api/v2/settings/objects/permissions/settings/setting_permissions.go
d0weinberger: Mhm, good question. I think we will revisit this at the story where everything is wired together, because at that point, we need to give the resource a name. We would then also need to adapt the example

dynatrace/api/v2/settings/objects/permissions/settings/user_accessor.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/terraform/hcl"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+type Users []*UserAccessor
+
+func (g *Users) Schema() map[string]*schema.Schema {
```
arthurpitman: For consistency:
```suggestion
func (_ *Users) Schema() map[string]*schema.Schema {
```

dynatrace/api/v2/settings/objects/permissions/settings/user_accessor.go
```diff
@@ -0,0 +1,81 @@
...
+			Type:        schema.TypeSet,
+			Required:    true,
+			MinItems:    1,
+			Description: "User that is to be granted read or write permissions",
+			Elem:        &schema.Resource{Schema: new(UserAccessor).Schema()},
+		},
+	}
+}
+
+func (g *Users) MarshalHCL(properties hcl.Properties) error {
```
arthurpitman: nit:
```suggestion
func (u *Users) MarshalHCL(properties hcl.Properties) error {
```

dynatrace/api/v2/settings/objects/permissions/settings/group_accessor.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/terraform/hcl"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+type Groups []*GroupAccessor
+
+func (g *Groups) Schema() map[string]*schema.Schema {
```
arthurpitman: nit:
```suggestion
func (_ *Groups) Schema() map[string]*schema.Schema {
```

arthurpitman (approved):

PR #734, opened 2025-07-30
feat: add settings permission DTO

Kirdock:
This represents the needed request and response bodies for permission CRUD operations.

### Additional notes
I decided to go with two structs for the `PermissionObject` instead of one to prevent unnecessary nil checks and error handling of the accessor for some requests.

The needed requests are
- GetAll (accessor is returned)
- Create (accessor is required)
- UpdateAccessor (accessor is not needed)
- UpdateAllUsersAccessor (accessor is not needed)
- DeleteAccessor (no payload)
- DeleteAllUsersAccessor (no payload)

**Issue:** CA-15669

### DIFF
--- a/dynatrace/api/v2/settings/objects/permissions/settings/dto.go
+++ b/dynatrace/api/v2/settings/objects/permissions/settings/dto.go
@@ -1,0 +1,44 @@
+package settings
+
+type TypePermissions = string
+
+const (
+	Read  TypePermissions = "r"
+	Write TypePermissions = "w"
+)
+
+type TypeAccessor = string
+
+const (
+	AllUsers TypeAccessor = "all-users"
+	Group    TypeAccessor = "group"
+	User     TypeAccessor = "user"
+)
+
+type Accessor struct {
+	Type TypeAccessor `json:"type"`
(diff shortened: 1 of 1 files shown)


arthurpitman:
The structures look good 👍 

Naming-wise, I'd prefer the names used in the OpenAPI spec, so I'd prefer `AccessorPermissions` to `PermissionObject`, `UpdatePermissionsRequest` to `PermissionObjectUpdate` and `AccessorPermissionsList` to `PermissionObjects`.

I think `Type` works better as a suffix, so would prefer `AccessorType` and `PermissionsType`.

Kirdock:
> The structures look good 👍
> 
> Naming-wise, I'd prefer the names used in the OpenAPI spec, so I'd prefer `AccessorPermissions` to `PermissionObject`, `UpdatePermissionsRequest` to `PermissionObjectUpdate` and `AccessorPermissionsList` to `PermissionObjects`.
> 
> I think `Type` works better as a suffix, so would prefer `AccessorType` and `PermissionsType`.

I took the names from Monaco, so should we touch these names too? I would value consistency over slightly more correct.

arthurpitman (approved):
If we want, let's address naming in a follow-up PR.

PR #735, opened 2025-07-31
feat: add settings permission service

Kirdock:
This adds the service for the permission resource. This includes

- The diff check, if it needs to update, delete, or create permissions
- The actual CRUD calls

## Additional adjustments
- Missing unmarshals weren't visible as `d.Set(key, value)` errors weren't checked. In order to now break any existing resource, I added the error check iff the `DYNATRACE_DEBUG` flag is set.
- Moved accessors (read, write, none) into constants, so they can easily be re-used

**Issue:** CA-15664

As Sonar only complains about the test file, I would say we're safe to ignore it, as it makes sense not to overcomplicate the testing, and some duplication is fine.

### DIFF
--- a/dynatrace/api/v2/settings/objects/permissions/reconcile/all_users.go
+++ b/dynatrace/api/v2/settings/objects/permissions/reconcile/all_users.go
@@ -8,24 +8,27 @@ import (
 	permissions "github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/v2/settings/objects/permissions/settings"
 )
 
-func compareAndUpdateAllUsers(ctx context.Context, client permissions.AllUsersClient, objectID string, currentAllUsers, desiredAllUsers string) error {
+func getAllUserUpsert(ctx context.Context, client permissions.AllUsersClient, objectID string, currentAllUsers, desiredAllUsers string) (func(adminAccess bool) error, error) {
 	if currentAllUsers == desiredAllUsers {
-		return nil
+		// No update needed, current and desired are the same
+		return nil, nil
 	}
 
 	if desiredAllUsers == permissions.HCLAccessorNone {
-		_, err := client.DeleteAllUsersAccessor(ctx, objectID, false)
-		return err
+		return func(adminAccess bool) error {
+			_, err := client.DeleteAllUsersAccessor(ctx, objectID, adminAccess)
+			return err
+		}, nil
 	}
 
 	if currentAllUsers == permissions.HCLAccessorNone {
-		return createAllUsers(ctx, client, objectID, desiredAllUsers)
+		return getAllUserCreate(ctx, client, objectID, desiredAllUsers)
 	}
 
-	return updateAllUsers(ctx, client, objectID, desiredAllUsers)
+	return getAllUserUpdate(ctx, client, objectID, desiredAllUsers)
 }
 
-func createAllUsers(ctx context.Context, client permissions.AllUsersClient, objectID string, desiredAllUsers string) error {
+func getAllUserCreate(ctx context.Context, client permissions.AllUsersClient, objectID string, desiredAllUsers string) (func(adminAccess bool) error, error) {
 	body, err := json.Marshal(permissions.PermissionObject{
 		Accessor: permissions.Accessor{
 			Type: permissions.AllUsers,
@@ -33,21 +36,25 @@ func createAllUsers(ctx context.Context, client permissions.AllUsersClient, obje
 		Permissions: convert.HCLToDTOPermission(desiredAllUsers),
 	})
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	_, err = client.Create(ctx, objectID, false, body)
-	return err
+	return func(adminAccess bool) error {
+		_, err = client.Create(ctx, objectID, adminAccess, body)
+		return err
+	}, nil
 }
 
-func updateAllUsers(ctx context.Context, client permissions.AllUsersClient, objectID string, desiredAllUsers string) error {
+func getAllUserUpdate(ctx context.Context, client permissions.AllUsersClient, objectID string, desiredAllUsers string) (func(adminAccess bool) error, error) {
 	body, err := json.Marshal(permissions.PermissionObjectUpdate{
 		Permissions: convert.HCLToDTOPermission(desiredAllUsers),
 	})
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	_, err = client.UpdateAllUsersAccessor(ctx, objectID, false, body)
-	return err
+	return func(adminAccess bool) error {
+		_, err = client.UpdateAllUsersAccessor(ctx, objectID, adminAccess, body)
+		return err
+	}, nil
 }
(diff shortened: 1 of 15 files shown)


dynatrace/api/v2/settings/objects/permissions/reconcile/groups.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		return err
+	}
+
+	_, err = client.UpdateAccessor(ctx, objectID, permissions.Group, group.ID, false, body)
+	return err
+}
+
+func containsGroups(groups permissions.Groups, group *permissions.GroupAccessor) (exists, equals bool) {
+	for _, u := range groups {
```
arthurpitman: nit: could we change this to `g`?

dynatrace/api/v2/settings/objects/permissions/service.go
```diff
@@ -141,8 +141,13 @@ func (me *ServiceImpl) Update(ctx context.Context, _ string, v *permissions.Sett
...
 
 func (me *ServiceImpl) Delete(ctx context.Context, id string) error {
-	// no-op. Permissions are deleted when the object is deleted.
-	return nil
+	_, err := me.Upsert(ctx, &permissions.SettingPermissions{
+		SettingsObjectID: id,
+		Users:            nil,
+		Groups:           nil,
+		AllUsers:         permissions.HCLAccessorNone,
+	})
```
arthurpitman: nit: Consider extracting this into a variable with a descriptive name

dynatrace/api/v2/settings/objects/permissions/reconcile/all_users.go
```diff
@@ -24,29 +27,33 @@ func compareAndUpdateAllUsers(ctx context.Context, client permissions.AllUsersCl
 	return updateAllUsers(ctx, client, objectID, desiredAllUsers)
 }
 
-func createAllUsers(ctx context.Context, client permissions.AllUsersClient, objectID string, desiredAllUsers string) error {
+func createAllUsers(ctx context.Context, client permissions.AllUsersClient, objectID string, desiredAllUsers string) (func(adminAccess bool) error, error) {
```
arthurpitman: consider renaming this function to include `get...` in the name

dynatrace/api/v2/settings/objects/permissions/settings/group_accessor.go
```diff
@@ -37,7 +39,7 @@ func (_ *Groups) Schema() map[string]*schema.Schema {
 	}
 }
 
-func (g *Groups) MarshalHCL(properties hcl.Properties) error {
+func (g Groups) MarshalHCL(properties hcl.Properties) error {
```
arthurpitman, line 42: Shall we move this line to the next commit?

arthurpitman (approved):

dynatrace/api/v2/settings/objects/permissions/service.go
```diff
@@ -0,0 +1,157 @@
...
+	return &permissions.SettingPermissions{
+		SettingsObjectID: objectID,
+		Users:            users,
+		Groups:           groups,
+		AllUsers:         allUsers,
+	}
+}
+
+// convertDTOToHCLPermission converts the permissions from the DTO format to the HCL format.
+func convertDTOToHCLPermission(pms []permissions.TypePermissions) permissions.HCLAccessor {
```
d0weinberger: I guess there is no case where the slice is empty and we would need to return `HCLAccessorNone`, right?

dynatrace/api/v2/settings/objects/permissions/reconcile/reconcile.go
```diff
@@ -0,0 +1,23 @@
...
+
+func CompareAndUpdate(ctx context.Context, cl permissions.PermissionUpdateClient, current *permissions.SettingPermissions, desired *permissions.SettingPermissions) error {
+	groupErr := compareAndUpdateGroups(ctx, cl, current.SettingsObjectID, current.Groups, desired.Groups)
+	userErr := compareAndUpdateUsers(ctx, cl, current.SettingsObjectID, current.Users, desired.Users)
+	allUserEr := compareAndUpdateAllUsers(ctx, cl, current.SettingsObjectID, current.AllUsers, desired.AllUsers)
+
+	return errors.Join(groupErr, userErr, allUserEr)
+}
+
+func convertHCLToDTOPermission(access permissions.HCLAccessor) []permissions.TypePermissions {
```
d0weinberger: [nit] Just noticed, this conversion function is the only one in `reconcile.go`, the others are in `service.go`.

dynatrace/api/v2/settings/objects/permissions/service.go
```diff
@@ -0,0 +1,157 @@
...
+		SettingsObjectID: id,
+		Users:            nil,
+		Groups:           nil,
+		AllUsers:         permissions.HCLAccessorNone,
+	}
+	_, err := me.Upsert(ctx, emptyPermissions)
+	return err
+}
+
+func (me *ServiceImpl) New() *permissions.SettingPermissions {
```
d0weinberger: I have only seen this used in the test. Is it used anywhere else? If not, maybe the function is not necessary and could be removed. Wdyt?

d0weinberger (approved):

tomazpu (approved):

dynatrace/api/v2/settings/objects/permissions/reconcile/reconcile.go
Kirdock: Yes,  because the reconcile needs to map HCL to API, and we only need it there. In the service, we only need to map from API to HCL but not the other way around

dynatrace/api/v2/settings/objects/permissions/service.go
Kirdock: Correct. None is basically "doesn't exist at all". The only cases returned by the API are `["r"]` and `["r", "w"]`. It isn't even possible to save an empty array.

dynatrace/api/v2/settings/objects/permissions/service.go
Kirdock: At first, I thought this was part of the interface, but it doesn't look like it.
I'll remove it

Edit: removed 👍

dynatrace/api/v2/settings/objects/permissions/settings/group_accessor.go
Kirdock: done 👍

dynatrace/api/v2/settings/objects/permissions/reconcile/reconcile.go
Kirdock: I could move both into a new `convert.go` file, though, if you want.

d0weinberger: Okay true. No, that's not necessary, I think. I think it's okay the way it is.

Kirdock: I'll move it to `convert.go`, then it's also reducing the complexity of the service a bit.

Edit: done

PR #746, opened 2025-08-07
feat: add tests for settings permissions

d0weinberger:
This PR introduces some tests for the marshalling and unmarshalling between the settings permissions object and HCL.

### DIFF
--- a/dynatrace/api/v2/settings/objects/permissions/settings/setting_permissions_test.go
+++ b/dynatrace/api/v2/settings/objects/permissions/settings/setting_permissions_test.go
@@ -1,0 +1,173 @@
+package settings
+
+import (
+	"testing"
+
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/terraform/hcl"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSettingPermissions_MarshalHCL(t *testing.T) {
+	cases := []struct {
+		name        string
+		permissions *SettingPermissions
+		expected    hcl.Properties
+	}{
+		{
+			name: "full-permissions",
+			permissions: &SettingPermissions{
+				SettingsObjectID: "obj-123",
+				AllUsers:         HCLAccessorNone,
+				Users: Users{
+					{UID: "user1", Access: HCLAccessorWrite},
+				},
+				Groups: Groups{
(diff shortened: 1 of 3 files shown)


dynatrace/api/v2/settings/objects/permissions/settings/user_accessor_test.go
```diff
@@ -0,0 +1,175 @@
...
+		wantErr  bool
+	}{
+		{
+			name:  "read-access",
+			input: map[string]interface{}{"uid": "my-id", "access": "read"},
+			expected: &UserAccessor{
+				UID:    "my-id",
+				Access: "read",
+			},
+			wantErr: false,
```
Kirdock: As we don't test a `true` case, we can remove this property from the struct, right?

dynatrace/api/v2/settings/objects/permissions/settings/setting_permissions_test.go
```diff
@@ -0,0 +1,209 @@
...
+		})
+	}
+}
+
+func settingPermissionsAreEqual(this SettingPermissions, that SettingPermissions) bool {
+	return this.SettingsObjectID == that.SettingsObjectID &&
+		this.AllUsers == that.AllUsers &&
+		equalUnordered(this.Users, that.Users) &&
+		equalUnordered(this.Groups, that.Groups)
+}
```
Kirdock: You could actually compare the arrays via `assert.ElementsMatch`

```go
func settingPermissionsAreEqual(t *testing.T, this SettingPermissions, that SettingPermissions) {
	assert.Equal(t, this.SettingsObjectID, that.SettingsObjectID)
	assert.Equal(t, this.AllUsers, that.AllUsers)
	assert.ElementsMatch(t, this.Users, that.Users)
	assert.ElementsMatch(t, this.Groups, that.Groups)
}
```

d0weinberger: Ah, nice! Will do that 👍

dynatrace/api/v2/settings/objects/permissions/settings/user_accessor_test.go
d0weinberger: True 👍  will do that

Kirdock (approved):
[nit]: we should use the constants for `read`, `write`, `none`

PR #747, opened 2025-08-08
feat: Settings permissions list

arthurpitman:
#### **Why** this PR?
An implementation for `func (me *ServiceImpl) List(ctx context.Context) (api.Stubs, error)` is required for export of settings permissions.

#### **What** has changed?
The PR provides the implementation of list as well as a basic settings clients for retrieving settings objects with permissions.

#### **How** does it do it?
First, all schemas IDs with owner-based access control are retrieved, then for each of these, all object IDs are retrieved.

#### How is it **tested**?
New tests in `service_test.go` as well as tests for the settings client in `settings_client_test.go`

On its own, it has no effect on the user, however is a prerequisite for the  export functionality.

**Issue:** CA-16047

### DIFF
--- a/dynatrace/api/v2/settings/objects/permissions/service.go
+++ b/dynatrace/api/v2/settings/objects/permissions/service.go
@@ -1,3 +1,19 @@
+/*
+ * @license
+ * Copyright 2025 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package permissions
 
 import (
@@ -21,8 +37,9 @@ func Service(credentials *rest.Credentials) settings.CRUDService[*permissions.Se
 }
 
 type ServiceImpl struct {
-	Credentials *rest.Credentials
-	Client      permissions.PermissionClient
+	Credentials    *rest.Credentials
+	Client         permissions.PermissionClient
+	SettingsClient PlatformSettingsClient
 }
 
 func (me *ServiceImpl) getClient(ctx context.Context) (permissions.PermissionClient, error) {
@@ -36,6 +53,19 @@ func (me *ServiceImpl) getClient(ctx context.Context) (permissions.PermissionCli
 
 	me.Client = permissions2.NewClient(restClient)
 	return me.Client, nil
+}
+
+func (me *ServiceImpl) getSettingsClient(ctx context.Context) (PlatformSettingsClient, error) {
+	if me.SettingsClient != nil {
+		return me.SettingsClient, nil
+	}
+	restClient, err := rest.CreatePlatformClient(ctx, me.Credentials.OAuth.EnvironmentURL, me.Credentials)
+	if err != nil {
+		return nil, err
+	}
+
+	me.SettingsClient = NewPlatformSettingsClient(restClient)
+	return me.SettingsClient, nil
 }
 
 func (me *ServiceImpl) Get(ctx context.Context, objectID string, v *permissions.SettingPermissions) error {
@@ -75,7 +105,30 @@ func (me *ServiceImpl) SchemaID() string {
 }
 
 func (me *ServiceImpl) List(ctx context.Context) (api.Stubs, error) {
-	return api.Stubs{}, nil
+	client, err := me.getSettingsClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+	schemaIds, err := client.GetSchemaIDsWithOwnerBasedAccessControl(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	stubs := api.Stubs{}
+	for _, schemaID := range schemaIds {
+		objectIDs, err := client.ListObjectsIDsOfSchema(ctx, schemaID)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, id := range objectIDs {
+			stubs = append(stubs, &api.Stub{
+				ID:   id,
+				Name: id,
+			})
+		}
+	}
+	return stubs, nil
 }
 
 func (me *ServiceImpl) Upsert(ctx context.Context, v *permissions.SettingPermissions) (*api.Stub, error) {
--- a/dynatrace/api/v2/settings/objects/permissions/service_test.go
+++ b/dynatrace/api/v2/settings/objects/permissions/service_test.go
@@ -1,3 +1,19 @@
+/*
+ * @license
+ * Copyright 2025 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package permissions_test
 
 import (
@@ -6,6 +22,7 @@ import (
 	"sync/atomic"
 	"testing"
 
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api"
 	permissionService "github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/v2/settings/objects/permissions"
 	permissions "github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/v2/settings/objects/permissions/settings"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/rest"
@@ -46,6 +63,19 @@ func (c *clientStub) UpdateAllUsersAccessor(ctx context.Context, objectID string
 
 func (c *clientStub) DeleteAllUsersAccessor(ctx context.Context, objectID string, adminAccess bool) (coreApi.Response, error) {
 	return c.deleteAllUsersAccessor(ctx, objectID, adminAccess)
+}
+
+type settingsClientStub struct {
+	getSchemaIDsWithOwnerBasedAccessControl func(ctx context.Context) ([]string, error)
+	listObjectsIDsOfSchema                  func(ctx context.Context, schemaID string) ([]string, error)
+}
+
+// Implement exporter.SettingsClient interface
+func (c *settingsClientStub) GetSchemaIDsWithOwnerBasedAccessControl(ctx context.Context) ([]string, error) {
+	return c.getSchemaIDsWithOwnerBasedAccessControl(ctx)
+}
+func (c *settingsClientStub) ListObjectsIDsOfSchema(ctx context.Context, schemaID string) ([]string, error) {
+	return c.listObjectsIDsOfSchema(ctx, schemaID)
 }
 
 func TestService(t *testing.T) {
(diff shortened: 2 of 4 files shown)


dynatrace/api/v2/settings/objects/permissions/settings_client.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	return api.NewResponseFromHTTPResponse(response)
+}
+
+func (c *settingsClient) getObjectsNextPage(ctx context.Context, nextPageKey string) (api.Response, error) {
+	queryParams := url.Values{
+		"nextPageKey": []string{nextPageKey},
+	}
+	response, err := c.client.GET(ctx, objectsEndpoint, rest.RequestOptions{QueryParams: queryParams})
```
d0weinberger, line 160: Is there no adminAccess for the subsequent pages?

arthurpitman: I will go and test this, even if the API docs state: "When the nextPageKey is set to obtain subsequent pages, you must omit all other query parameters."

dynatrace/api/v2/settings/objects/permissions/service.go
```diff
@@ -72,7 +86,30 @@ func (me *ServiceImpl) SchemaID() string {
...
+
+	stubs := api.Stubs{}
+	for _, schemaID := range schemaIds {
+		objectIDs, err := client.ListObjectsIDsOfSchema(ctx, schemaID)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, id := range objectIDs {
+			stubs = append(stubs, &api.Stub{
```
d0weinberger, line 125: The information which schema the object belongs to is not preserved in the list, but that doesn't matter, right?

dynatrace/api/v2/settings/objects/permissions/settings_client.go
```diff
@@ -0,0 +1,166 @@
+package permissions
```
tomazpu, line 17: Missing license header?

dynatrace/api/v2/settings/objects/permissions/settings_client_test.go
```diff
@@ -0,0 +1,248 @@
+package permissions_test
```
tomazpu, line 19: same here

dynatrace/api/v2/settings/objects/permissions/settings_client.go
```diff
@@ -0,0 +1,166 @@
...
+type objectsList struct {
+	Items       []objectStub `json:"items"`
+	NextPageKey string       `json:"nextPageKey"`
+}
+
+type objectStub struct {
+	ObjectID string `json:"objectId"`
+}
+
+type SettingsClient interface {
```
tomazpu: Would `PlatformSettingsClient` be more specific?

dynatrace/api/v2/settings/objects/permissions/settings_client.go
```diff
@@ -0,0 +1,166 @@
...
+
+		for _, objectStub := range ol.Items {
+			objectIDs = append(objectIDs, objectStub.ObjectID)
+		}
+	}
+
+	return objectIDs, nil
+}
+
+const fields = "objectId"
```
tomazpu: I would name it `FieldObjectID`, or something more descriptive.

dynatrace/api/v2/settings/objects/permissions/settings_client.go
```diff
@@ -0,0 +1,166 @@
...
+const pageSize = 100
```
tomazpu: NIT: And this like pageSizeLimit and declare both in a const block on top. Just a style opinion :)

dynatrace/api/v2/settings/objects/permissions/service.go
arthurpitman: Correct. The current implementation doesn't need this.

dynatrace/api/v2/settings/objects/permissions/settings_client.go
arthurpitman: I have tested this and can confirm: `adminAccess` is recorded in the initial query and is not required in further requests. Only `nextPageKey` is required when retrieving subsequent pages.

Kirdock:
Does the export actually work for you? If I try adding it to the resources, and executing the export I get the following panic
`panic: github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/v2/settings/objects/permissions/settings.SettingPermissions does neither have a property 'Name', 'DisplayName', 'Label' or 'Key' nor does it offer a method 'Name()'`

So something like this maybe for the `SettingsPermission` struct?
```go
func (sp *SettingPermissions) Name() string {
	return "permissions"
}
```
 => `permissions_{COUNTER}.settings_permissions.tf` and `resource "dynatrace_settings_permissions" "permissions_1"{...}`

This would be easier to read than the objectID and it's handled via incremental counter (, else we can also return `sp.SettingsObjectID` if you prefer this one.

dynatrace/api/v2/settings/objects/permissions/settings_client.go
```diff
@@ -0,0 +1,182 @@
...
+
+		for _, objectStub := range ol.Items {
+			objectIDs = append(objectIDs, objectStub.ObjectID)
+		}
+	}
+
+	return objectIDs, nil
+}
+
+func (c *settingsClient) getObjectIDsFirstPageWithAdminRetry(ctx context.Context, schemaID string) (api.Response, error) {
```
Kirdock: Depending on what gets merged first, either this one or the admin-access PR should align the behavior (re-use the DoWithAdminAccess function)

arthurpitman:
> Does the export actually work for you? If I try adding it to the resources, and executing the export I get the following panic `panic: github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/v2/settings/objects/permissions/settings.SettingPermissions does neither have a property 'Name', 'DisplayName', 'Label' or 'Key' nor does it offer a method 'Name()'`
> 
> So something like this maybe for the `SettingsPermission` struct?
> 
> ```go
> func (sp *SettingPermissions) Name() string {
> 	return "permissions"
> }
> ```
> 
> => `permissions_{COUNTER}.settings_permissions.tf` and `resource "dynatrace_settings_permissions" "permissions_1"{...}`
> 
> This would be easier to read than the objectID and it's handled via incremental counter (, else we can also return `sp.SettingsObjectID` if you prefer this one.

About `Name()`:  you are correct, I found that too, but wanted to discuss that for "wiring-up" or at least in a follow up PR, because it affects UX. For what it is worth, I was returning the settings object ID in my implementation.

dynatrace/api/v2/settings/objects/permissions/settings_client.go
arthurpitman: Done, now it reuses `DoWithAdminAccess(...)`.

dynatrace/api/v2/settings/objects/permissions/service_test.go
```diff
@@ -132,6 +162,68 @@ func TestService(t *testing.T) {
...
+
+		t.Run("Returns error if GetSchemaIDsWithOwnerBasedAccessControl fails", func(t *testing.T) {
+			settingsClient := &settingsClientStub{
+				getSchemaIDsWithOwnerBasedAccessControl: func(ctx context.Context) ([]string, error) {
+					return nil, assert.AnError
+				},
+			}
+			service := permissionService.ServiceImpl{SettingsClient: settingsClient}
+			stubs, err := service.List(t.Context())
+			assert.Error(t, err)
```
Kirdock: [nit]: should we assert if the error is `assert.AnError`?

dynatrace/api/v2/settings/objects/permissions/service_test.go
```diff
@@ -132,6 +162,68 @@ func TestService(t *testing.T) {
...
+				getSchemaIDsWithOwnerBasedAccessControl: func(ctx context.Context) ([]string, error) {
+					return []string{"schemaID1"}, nil
+				},
+				listObjectsIDsOfSchema: func(ctx context.Context, schemaID string) ([]string, error) {
+					return nil, assert.AnError
+				},
+			}
+			service := permissionService.ServiceImpl{SettingsClient: settingsClient}
+			stubs, err := service.List(t.Context())
+			assert.Error(t, err)
```
Kirdock: [nit]: should we assert if the error is `assert.AnError`?

dynatrace/api/v2/settings/objects/permissions/service_test.go
```diff
@@ -132,6 +162,68 @@ func TestService(t *testing.T) {
...
+			assert.Nil(t, stubs)
+		})
+
+		t.Run("Returns error if getSettingsClient fails", func(t *testing.T) {
+			service := permissionService.ServiceImpl{
+				Credentials: &rest.Credentials{},
+			}
+			stubs, err := service.List(t.Context())
+			assert.Error(t, err)
```
Kirdock: [nit] We could also validate the error here `assert.ErrorIs(t, rest.NoPlatformCredentialsErr, err)`

dynatrace/api/v2/settings/objects/permissions/settings_client.go
Kirdock: I would also vote for that. With this, we have multiple settings clients, so it would be nice to point out the difference

dynatrace/api/v2/settings/objects/permissions/settings_client.go
Kirdock, line 1: [nit] Technically, the existing settings client can also be used, right? We could just use

```go
oauthClient, _ := rest.CreateClassicOAuthBasedClient(ctx, me.credentials)
return settings20.NewClient(nil, oauthClient, schemaIDs)
```

Then we could actually merge both clients and have one instead of two.
We would also not need much changes in the tests, we just move them and adjust the creation of the client

dynatrace/api/v2/settings/objects/permissions/settings_client.go
```diff
@@ -0,0 +1,170 @@
...
+	pageSize       = 100
+)
+
+type schemaList struct {
+	Items []schemaStub `json:"items"`
+}
+
+type schemaStub struct {
+	SchemaID                string `json:"schemaId"`
+	OwnerBasedAccessControl bool   `json:"ownerBasedAccessControl"`
```
Kirdock, line 46: Technically, the `OwnerBasedAccessControl` is optional. At the moment, there is only `true` and `doesn't exist at all`, but this should not matter in our case, as we fall back to false.

dynatrace/api/v2/settings/objects/permissions/settings_client_test.go
```diff
@@ -0,0 +1,327 @@
...
+
+	t.Run("ListObjectsIDsOfSchema", func(t *testing.T) {
+
+		t.Run("retries without admin access if first request fails with forbidden", func(t *testing.T) {
+			server := httptest.NewServer(http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
+				require.Equal(t, request.URL.Query().Get("schemaIds"), "schemaID1")
+				require.Equal(t, request.URL.Query().Get("fields"), "objectId")
+				require.Empty(t, request.URL.Query().Get("nextPageKey"))
+				require.Equal(t, http.MethodGet, request.Method)
+				require.Equal(t, "/platform/classic/environment-api/v2/settings/objects", request.URL.Path)
```
Kirdock, line 116: [nit]: Should we introduce a counter here? At the moment, the test would still succeed if the `adminAccess` is not set at all.
Also, instead of asserting the path and method, we could also just use a server mux (also for other test cases), or is there a limitation with the query params?

dynatrace/api/v2/settings/objects/permissions/settings_client_test.go
```diff
@@ -0,0 +1,327 @@
...
+				require.NoError(t, err)
+			}))
+			defer server.Close()
+
+			url, _ := url.Parse(server.URL)
+			client := permissions.NewSettingsClient(rest.NewClient(url, server.Client()))
+
+			objectIDs, err := client.ListObjectsIDsOfSchema(t.Context(), "schemaID1")
+
+			assert.Error(t, err)
```
Kirdock: FYI, JSON marshal errors can easily be validated via
```go
synErr := &json.SyntaxError{}
assert.ErrorAs(t, err, &synErr)
```

dynatrace/api/v2/settings/objects/permissions/settings_client.go
```diff
@@ -0,0 +1,170 @@
...
+
+	objectIDs := []string{}
+	for _, objectStub := range ol.Items {
+		objectIDs = append(objectIDs, objectStub.ObjectID)
+	}
+
+	for {
+		if ol.NextPageKey == "" {
+			break
+		}
```
Kirdock: [nit] Technically we could merge those
```go
for ol.NextPageKey != "" {
}
```

dynatrace/api/v2/settings/objects/permissions/settings_client.go
```diff
@@ -0,0 +1,170 @@
...
+
+		response, err := c.getObjectsNextPage(ctx, ol.NextPageKey)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get next page of objects: %w", err)
+		}
+
+		ol = objectsList{}
```
Kirdock, line 128: Is this actually needed? Would `json.Unmarshal` not just override everything?

Kirdock (approved):

dynatrace/api/v2/settings/objects/permissions/settings_client.go
Kirdock: clarified: yes, Unmarshal only overrides values that are inside the payload. If the nextPageKey is not inside the response data, it won't override it at all.

arthurpitman: It seems it is. In particular, doing this resets the nextPageKey, ensuring that the loop terminates after the last page is received.

dynatrace/api/v2/settings/objects/permissions/settings_client.go
arthurpitman: I think we could do that in a follow-up if we want. But I decided for yet another client because I wanted to use the functionality of the core library.

tomazpu (approved):

PR #749, opened 2025-08-11
feat: wire up settings permissions

Kirdock:
This adds the settings permission resource to the provider to be used as a resource and for export.
Limitation: Because other resources don't support `adminAccess=true`, only references to objects that the OAuth client and token are owner of are resolved and downloaded.

Because of the generic settings dependency, I also adjusted the generic settings service to support platform and classic credentials

Regarding the export name. Currently, I've set it to `permissions`, which will result in the following
image

 => `permissions_{COUNTER}.settings_permissions.tf` and `resource "dynatrace_settings_permissions" "permissions_1"{...}`

This would be easier to read than the objectID and it's handled via incremental counter, else we can also return `sp.SettingsObjectID` if this one is preferred. 

Example with name `permissions`
```hcl
resource "dynatrace_settings_permissions" "permissions_1" {
  all_users          = "none"
  settings_object_id = "${var.dynatrace_automation_workflow_jira.asdf.id}" // or the plain objectID if the reference is not found
}
```

Example with name `sp.SettingsObjectID`
```hcl
resource "dynatrace_settings_permissions" "my-very-long-object-id-which-can-have-more-than-150-characters-and-basically-tells-nothing" {
  all_users          = "none"
  settings_object_id = "${var.dynatrace_automation_workflow_jira.asdf.id}" // or the plain objectID if the reference is not found
}
```

**Issue:** CA-15665

### DIFF
--- a/dynatrace/api/builtin/generic/service.go
+++ b/dynatrace/api/builtin/generic/service.go
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/url"
 	"strings"
 	"time"
 
@@ -31,9 +32,6 @@ import (
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/settings"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/settings/services/settings20"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/shutdown"
-	crest "github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
-
-	"net/url"
 )
 
 func Service(credentials *rest.Credentials) settings.CRUDService[*generic.Settings] {
@@ -85,17 +83,11 @@ type schemataResponse struct {
 }
 
 func (me *service) List(ctx context.Context) (api.Stubs, error) {
-	client, err := rest.CreateClassicClient(me.credentials.URL, me.credentials.Token)
-	if err != nil {
-		return nil, err
-	}
-	response, err := client.GET(ctx, "api/v2/settings/schemas", crest.RequestOptions{})
-	if err != nil {
-		return nil, err
-	}
+	client := rest.HybridClient(me.credentials)
 	var schemata schemataResponse
-	if response.Body != nil {
-		json.NewDecoder(response.Body).Decode(&schemata)
+	err := client.Get(ctx, "/api/v2/settings/schemas", 200).Finish(&schemata)
+	if err != nil {
+		return nil, err
 	}
 	if len(schemata.Items) == 0 {
 		return api.Stubs{}, nil
@@ -224,19 +216,16 @@ type QueryParams struct {
 }
 
 func (me *service) ListSpecific(ctx context.Context, query QueryParams) (api.Stubs, error) {
-	client, err := rest.CreateClassicClient(me.credentials.URL, me.credentials.Token)
-	if err != nil {
-		return nil, err
-	}
+	client := rest.HybridClient(me.credentials)
 
 	stubs := api.Stubs{}
 	nextPage := true
 	var nextPageKey *string
 	for nextPage {
 		var sol settings20.SettingsObjectList
-		var options crest.RequestOptions
+		var queryValues url.Values
 		if nextPageKey == nil {
-			options.QueryParams = url.Values{
+			queryValues = url.Values{
 				"fields":    []string{"objectId,scope,value,schemaId"},
 				"pageSize":  []string{"100"},
 				"schemaIds": []string{query.Schema},
@@ -244,19 +233,20 @@ func (me *service) ListSpecific(ctx context.Context, query QueryParams) (api.Stu
 				"filter":    []string{query.Filter},
 			}
 		} else {
-			options.QueryParams = url.Values{
+			queryValues = url.Values{
 				"nextPageKey": []string{*nextPageKey},
 			}
 		}
+		u := url.URL{
+			Path:     "/api/v2/settings/objects",
+			RawQuery: queryValues.Encode(),
+		}
 
-		response, err := client.GET(ctx, "api/v2/settings/objects", options)
+		err := client.Get(ctx, u.String()).Finish(&sol)
 		if err != nil {
 			return nil, err
 		}
 
-		if response.Body != nil {
-			json.NewDecoder(response.Body).Decode(&sol)
-		}
 		if len(sol.Items) == 0 {
 			return api.Stubs{}, nil
 		}
--- a/dynatrace/export/resource_descriptor.go
+++ b/dynatrace/export/resource_descriptor.go
@@ -23,7 +23,9 @@ import (
 	"strings"
 
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/grail/segments"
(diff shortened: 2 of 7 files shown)


dynatrace/testing/api/settingstest.go
```diff
@@ -217,10 +217,6 @@ func TestAcc(t *testing.T, opts ...TestAccOptions) {
 		return
 	}
 
-	if v := os.Getenv("DYNATRACE_API_TOKEN"); v == "" {
```
Kirdock, line 220: FYI, I also removed this limitation, as we can also run AC tests with just platform credentials

dynatrace/export/resource_descriptor.go
```diff
@@ -1524,6 +1525,26 @@ var AllResources = map[ResourceType]ResourceDescriptor{
 		manualinsertion.Service,
 		Dependencies.ID(ResourceTypes.WebApplication),
 	),
+	ResourceTypes.SettingsPermissions: NewResourceDescriptor(
+		settingsPermissions.Service,
+
+		// OwnerBasedAccessControl enabled resources
```
arthurpitman, line 1531: Do we need to include `ResourceTypes.AWSAutomationConnections`? I think the schema `builtin:hyperscaler-authentication.aws.connection` does support it.

dynatrace/api/builtin/generic/service.go
```diff
@@ -85,18 +83,12 @@ type schemataResponse struct {
...
 
 func (me *service) List(ctx context.Context) (api.Stubs, error) {
-	client, err := rest.CreateClassicClient(me.credentials.URL, me.credentials.Token)
-	if err != nil {
-		return nil, err
-	}
-	response, err := client.GET(ctx, "api/v2/settings/schemas", crest.RequestOptions{})
+	client := rest.HybridClient(me.credentials)
+	var schemata schemataResponse
+	err := client.Get(ctx, "/api/v2/settings/schemas", 200).Finish(&schemata)
```
tomazpu, line 88: Another thing comes to my mind, and maybe we already discussed it. I don't remember why we are doing two gets with two clients in the list. Maybe we could add a comment or we could break the list down into two methods with meaningful names.

dynatrace/export/resource_descriptor.go
Kirdock: You're right, I forgot about the built-in ones.
Pushed 👍

dynatrace/api/builtin/generic/service.go
```diff
@@ -224,39 +216,33 @@ type QueryParams struct {
...
 			}
 		} else {
-			options.QueryParams = url.Values{
+			queryValues = url.Values{
 				"nextPageKey": []string{*nextPageKey},
 			}
 		}
 
-		response, err := client.GET(ctx, "api/v2/settings/objects", options)
+		err := client.Get(ctx, fmt.Sprintf("/api/v2/settings/objects?%s", queryValues.Encode())).Finish(&sol)
```
tomazpu: I would prefer the usage of `url.URL` over `fmt.Sprintf`. 

Here we use the hybrid client and not the settings20 for a specific one.

dynatrace/api/builtin/generic/service.go
```diff
@@ -85,18 +83,12 @@ type schemataResponse struct {
 }
 
 func (me *service) List(ctx context.Context) (api.Stubs, error) {
-	client, err := rest.CreateClassicClient(me.credentials.URL, me.credentials.Token)
-	if err != nil {
-		return nil, err
-	}
-	response, err := client.GET(ctx, "api/v2/settings/schemas", crest.RequestOptions{})
+	client := rest.HybridClient(me.credentials)
```
tomazpu, line 86: I really don't like this pattern, usually only makes problems down the road. 
I think in the future we should think about the clients being a property on a struct of service.

arthurpitman, line 86: Would it be worth caching this on the `service` like we do in `dynatrace/api/v2/settings/objects/permissions/service.go`?

dynatrace/api/builtin/generic/service.go
Kirdock: We have a settings client here and a generic client. The settings client only handles setting objects and no schemas. Because the settings client contains more functionality, like pagination, we should not use the hybrid client for settings. 
So for now, I kept the implementation as equal as possible.

Everything else, like a refactoring of one of the clients, would be way out of scope. This implementation fixes the problem that the generic settings can't be used without an accessToken, as this would panic when it's inside the settings-permissions dependency.

dynatrace/api/builtin/generic/service.go
Kirdock: => Same as https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/749#discussion_r2273479400

Kirdock: We don't use stubs here like in the permission service tests, so I would say it doesn't matter which way we choose. I don't think that the state of a struct is preserved across Get/Update/etc. calls, so caching would be pointless. Correct me if I'm wrong 🤔

tomazpu (approved):

dynatrace/export/resource_descriptor.go
```diff
@@ -1629,6 +1651,7 @@ var excludeListedResourceGroups = []ResourceExclusionGroup{
 			{ResourceTypes.PlatformBucket, ""},
 			{ResourceTypes.Segments, ""},
 			{ResourceTypes.PlatformSLO, ""},
+			{ResourceTypes.SettingsPermissions, ""},
```
d0weinberger, line 1654: What am I missing here? Why is the newly introduced SettingsPermissions type listed in a list of Officially deprecated resources?

dynatrace/export/resource_descriptor.go
```diff
@@ -1524,6 +1525,27 @@ var AllResources = map[ResourceType]ResourceDescriptor{
...
+		Dependencies.ID(ResourceTypes.ServiceNowConnection),
+		Dependencies.ID(ResourceTypes.AutomationControllerConnections),
+		Dependencies.ID(ResourceTypes.EventDrivenAnsibleConnections),
+		Dependencies.ID(ResourceTypes.K8sAutomationConnections),
+		Dependencies.ID(ResourceTypes.SlackForWorkflows),
+		Dependencies.ID(ResourceTypes.GitHubConnection),
+		Dependencies.ID(ResourceTypes.GitLabConnection),
+		Dependencies.ID(ResourceTypes.MSTeamsConnection),
+		Dependencies.ID(ResourceTypes.AWSAutomationConnections),
+
```
d0weinberger, line 1546: I guess this newline isn't necessary

Kirdock: I made it as a separator to make it more visible

// defined resources
// fallback / schemas not defined as resources

dynatrace/export/resource_descriptor.go
Kirdock: This is not about deprecated resources. This is about limitations, and for this entry, it's about OAuth/Platform exclusive resources.

d0weinberger: Okay, I see it now. Looked at the wrong line

d0weinberger (approved):

PR #750, opened 2025-08-11
test: fix race condition of settings permissions test

Kirdock:
The delete-all test had a race condition. This wasn't a problem in the prod code; it was just the test that updated a counter.


arthurpitman (approved):
Nit: it might be slightly simpler with the `atomic.Int32` type, but it certainly works as it is here.

PR #751, opened 2025-08-11
feat(openpipeline): HCL schema for ingest sources, including some processor types

d0weinberger:
### Why this PR?
This PR is the first in supporting the OpenPipeline settings schemas in the Terraform provider.

### What has changed?

This PR adds the HCL representation of the ingest-sources schema(s).

### How does it do it?

The first commit introduces some example files: A JSON file which represents the payload expected by the settings API and a TF file which shows how the HCL representation of ingest-source resources would look like.

The second commit introduces support for several processors: `drop`, `fieldAdd`, `fieldRename`, `fieldRemove`, and `dql`.
`Schema()`, `MarshalHCL()`, and `UnmarshalHCL()` are implemented.

The third commit introduces support for the ingest-source itself, again with `Schema()`, `MarshalHCL()`, and `UnmarshalHCL()` implemented.

### How is it tested?

There are tests for `MarshalHCL()` and `UnmarshalHCL()` for the ingest source and for each processor type. Also, there is a JSON marshalling test for the ingest source.

### How does it affect users?

It doesn't yet. The feature is under development.

**Issue**: CA-15750

Notes: Unmarshalling of the ingest source is not figured out yet. We would need the `kind` information, which is not part of the JSON payload. We would need to parse that from the settings schema name.

### DIFF
--- a/dynatrace/api/builtin/openpipeline/processors/dql.go
+++ b/dynatrace/api/builtin/openpipeline/processors/dql.go
@@ -1,0 +1,49 @@
+/**
+* @license
+* Copyright 2025 Dynatrace LLC
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+ */
+
+package processors
+
+import (
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/terraform/hcl"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+type DqlAttributes struct {
+	Script string `json:"script"`
+}
+
+const DqlMaxScriptLength = 8192
+
+func (da *DqlAttributes) Schema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"script": {
+			Type:         schema.TypeString,
+			Description:  "The DQL script to apply on the record",
+			Required:     true,
+			ValidateFunc: validation.StringLenBetween(1, DqlMaxScriptLength),
+		},
+	}
+}
+
+func (da *DqlAttributes) MarshalHCL(properties hcl.Properties) error {
+	return properties.Encode("script", da.Script)
+}
+
+func (da *DqlAttributes) UnmarshalHCL(decoder hcl.Decoder) error {
+	return decoder.Decode("script", &da.Script)
+}
--- a/dynatrace/api/builtin/openpipeline/processors/dql_test.go
+++ b/dynatrace/api/builtin/openpipeline/processors/dql_test.go
@@ -1,0 +1,116 @@
+/**
+* @license
+* Copyright 2025 Dynatrace LLC
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+ */
+
+package processors_test
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/builtin/openpipeline/processors"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/terraform/hcl"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDqlAttributes_MarshalHCL(t *testing.T) {
+	var input = processors.DqlAttributes{Script: "fieldsAdd true"}
+	var actual = hcl.Properties{}
+
+	err := input.MarshalHCL(actual)
+
+	assert.Equal(t, hcl.Properties{"script": "fieldsAdd true"}, actual)
+	assert.NoError(t, err)
+}
+
+func TestDqlAttributes_UnmarshalHCL(t *testing.T) {
+	d := schema.TestResourceDataRaw(t,
+		new(processors.DqlAttributes).Schema(),
+		map[string]interface{}{"script": "fieldsAdd true"})
+	assert.NotNil(t, d)
+	var actual processors.DqlAttributes
+	decoder := hcl.DecoderFrom(d)
+
+	err := actual.UnmarshalHCL(decoder)
+
+	assert.Equal(t, processors.DqlAttributes{Script: "fieldsAdd true"}, actual)
+	assert.NoError(t, err)
+}
+
+func TestDqlAttributes_Validate(t *testing.T) {
+	cases := []struct {
+		name    string
+		input   map[string]interface{}
+		wantErr bool
+		errMsg  string
+	}{
+		{
+			name: "valid",
+			input: map[string]interface{}{
+				"script": "fieldsAdd true",
+			},
+			wantErr: false,
+		},
+		{
+			name: "invalid",
+			input: map[string]interface{}{
+				"script":        "fieldsAdd true",
+				"unknown-field": true,
+			},
+			wantErr: true,
+			errMsg:  "Invalid or unknown key",
+		},
+		{
+			name:    "empty",
+			input:   map[string]interface{}{},
+			wantErr: true,
+			errMsg:  "Missing required argument",
+		},
+		{
+			name: "script empty",
+			input: map[string]interface{}{
+				"script": "",
+			},
+			wantErr: true,
+			errMsg:  "expected length of script to be in the range (1 - 8192), got ",
+		},
+		{
+			name: "script exceeds limit",
+			input: map[string]interface{}{
+				"script": strings.Repeat("a", processors.DqlMaxScriptLength+1),
+			},
+			wantErr: true,
+			errMsg:  "expected length of script to be in the range (1 - 8192), got " + strings.Repeat("a", processors.DqlMaxScriptLength+1),
+		},
+	}
+
+	r := &schema.Resource{Schema: new(processors.DqlAttributes).Schema()}
+	for _, tt := range cases {
+		t.Run(tt.name, func(t *testing.T) {
+			c := terraform.NewResourceConfigRaw(tt.input)
+
+			diags := r.Validate(c)
+
+			assert.Equal(t, tt.wantErr, diags.HasError())
+
+			if diags.HasError() {
+				assert.Equal(t, tt.errMsg, diags[0].Summary)
+			}
+		})
+	}
+}
--- a/dynatrace/api/builtin/openpipeline/processors/fields_add.go
+++ b/dynatrace/api/builtin/openpipeline/processors/fields_add.go
@@ -1,0 +1,88 @@
+/**
+* @license
+* Copyright 2025 Dynatrace LLC
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+ */
+
+package processors
+
+import (
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/terraform/hcl"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+type FieldsAddAttributes struct {
+	Fields []*FieldsAddItem `json:"fields"`
+}
+
+func (faa *FieldsAddAttributes) Schema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"field": {
+			Type:        schema.TypeList,
+			Elem:        &schema.Resource{Schema: new(FieldsAddItem).Schema()},
+			Description: "Field to add to the record",
+			Required:    true,
+			MinItems:    1,
+			MaxItems:    50,
+		},
+	}
+}
+
+func (faa *FieldsAddAttributes) MarshalHCL(properties hcl.Properties) error {
+	return properties.EncodeSlice("field", faa.Fields)
+}
+
+func (faa *FieldsAddAttributes) UnmarshalHCL(decoder hcl.Decoder) error {
+	return decoder.DecodeSlice("field", &faa.Fields)
+}
+
+type FieldsAddItem struct {
+	Name  string `json:"name"`
+	Value string `json:"value"`
+}
+
+const FieldsAddMaxNameLength = 256
+const FieldsAddMaxValueLength = 512
+
+func (f *FieldsAddItem) Schema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"name": {
+			Type:         schema.TypeString,
+			Description:  "Name of the field",
+			Required:     true,
+			ValidateFunc: validation.StringLenBetween(0, FieldsAddMaxNameLength),
+		},
+		"value": {
+			Type:         schema.TypeString,
+			Description:  "Value to assign to the field",
+			Required:     true,
+			ValidateFunc: validation.StringLenBetween(0, FieldsAddMaxValueLength),
+		},
+	}
+}
+
+func (f *FieldsAddItem) MarshalHCL(properties hcl.Properties) error {
+	return properties.EncodeAll(map[string]any{
+		"name":  f.Name,
+		"value": f.Value,
+	})
+}
+
+func (f *FieldsAddItem) UnmarshalHCL(decoder hcl.Decoder) error {
+	return decoder.DecodeAll(map[string]any{
+		"name":  &f.Name,
+		"value": &f.Value,
+	})
+}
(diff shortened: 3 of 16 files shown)


dynatrace/api/builtin/openpipeline/ingestsources/settings/ingest_source_test.go
```diff
@@ -0,0 +1,228 @@
...
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+	for _, tc := range cases {
```
Laubi: Are you planning to add more test cases? If not, the table tests are not needed.
Same below

dynatrace/api/builtin/openpipeline/ingestsources/testdata/terraform/example.tf
```diff
@@ -0,0 +1,235 @@
...
+      sample_data = "my sample data"
+      matcher     = "not true"
+      fieldsRemove {
+        fields = ["to-remove-1", "to-remove-2"]
+      }
+    }
+  }
+}
+
+  kind = "events"
```
Laubi: This is not a valid document

dynatrace/api/builtin/openpipeline/processors/dql.go
Laubi, line 1: No tests for any processor?

dynatrace/api/builtin/openpipeline/ingestsources/testdata/terraform/example.tf
```diff
@@ -0,0 +1,235 @@
+resource "dynatrace_openpipeline_ingest_source" "ingestsource" {
```
Laubi: To prevent mixups in the future, should we call it `dynatrace_openpipeline_v2_ingest_source` or `dynatrace_openpipeline_ingest_source_v2`?

dynatrace/api/builtin/openpipeline/ingestsources/settings/ingest_source.go
```diff
@@ -0,0 +1,189 @@
...
+
+func (is *IngestSource) Schema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"kind": {
+			Type:         schema.TypeString,
+			Required:     true,
+			Description:  "Indicates pipeline kind",
+			ValidateFunc: validation.StringInSlice(AllowedKind, true),
+		},
+		"enabled": {
```
Laubi: should we make them enabled by default? Most defined will be so it might make sense. I am also ok with forcing the user to define it

dynatrace/api/builtin/openpipeline/ingestsources/settings/ingest_source.go
```diff
@@ -0,0 +1,189 @@
+package settings
+
+import (
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/builtin/openpipeline/processors"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/terraform/hcl"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+var AllowedKind = []string{
```
Laubi:
```suggestion
var AllowedKinds = []string{
```

I'd go with plural

dynatrace/api/builtin/openpipeline/ingestsources/settings/ingest_source.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+type IngestSource struct {
+	Kind          string             `json:"-"`
+	DefaultBucket *string            `json:"defaultBucket,omitempty"`
+	DisplayName   string             `json:"displayName"`
+	Enabled       bool               `json:"enabled"`
+	PathSegment   string             `json:"pathSegment"`
+	StaticRouting *PipelineReference `json:"staticRouting,omitempty"`
+	Processing    *Processing        `json:"processing"`
```
Laubi: Pointer, but not `omitempty`; is this on purpose?

dynatrace/api/builtin/openpipeline/ingestsources/testdata/terraform/example.tf
d0weinberger: Whoops, forgot to delete all of that.

dynatrace/api/builtin/openpipeline/ingestsources/settings/ingest_source.go
d0weinberger: Actually I think it should not be a pointer, because the API demands `processing` not being null. I'll change that.

dynatrace/api/builtin/openpipeline/ingestsources/testdata/terraform/example.tf
d0weinberger: Hm, I think in the original refinement we concluded that `v2` is not necessary, but we should discuss this again.

dynatrace/api/builtin/openpipeline/ingestsources/settings/ingest_source_test.go
d0weinberger: I will add some more cases, yes. At least one additional case.

dynatrace/api/builtin/openpipeline/processors/dql.go
d0weinberger: There are tests for all processor types in `processor_test.go`. This has the advantage to also test the keys given to the optional properties depending on which processor type is used. I will add the tests here as well, though.

dynatrace/api/builtin/openpipeline/ingestsources/settings/ingest_source.go
d0weinberger: I'll change the default to "true"

dynatrace/api/builtin/openpipeline/ingestsources/settings/ingest_source.go
arthurpitman: A pointer is needed here due to current limitations in `properties.go`.

d0weinberger: Kept the pointer, but added `omitempty`. That doesn't change anything for marshalling, because I added a `MarshalJSON()` function, but for unmarshalling

dynatrace/api/builtin/openpipeline/processors/processor.go
```diff
@@ -0,0 +1,192 @@
...
+			Default:     true,
+		},
+		"id": {
+			Type:        schema.TypeString,
+			Description: "Identifier of the processor. Must be unique within a stage. Must not start with 'dt.' or 'dynatrace.'",
+			Required:    true,
+		},
+		"type": {
+			Type:        schema.TypeString,
+			Description: "Type of the processor.",
```
Kirdock: Should we also add the allowed types here?

dynatrace/api/builtin/openpipeline/processors/processor.go
```diff
@@ -0,0 +1,192 @@
...
+func (p *Processor) Schema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"enabled": {
+			Type:        schema.TypeBool,
+			Description: "Indicates if the object is active",
+			Default:     true,
+		},
+		"id": {
+			Type:        schema.TypeString,
+			Description: "Identifier of the processor. Must be unique within a stage. Must not start with 'dt.' or 'dynatrace.'",
```
Kirdock, line 90: Not sure if we want to tackle this, but we have a lot of other constraints like patterns and min/max length. These are defined in the schema.

dynatrace/api/builtin/openpipeline/processors/processor.go
```diff
@@ -0,0 +1,192 @@
...
+			Required:    true,
+		},
+		"sample_data": {
+			Type:        schema.TypeString,
+			Description: "Sample data related to the processor for documentation or testing",
+			Optional:    true,
+		},
+		"matcher": {
+			Type:        schema.TypeString,
+			Description: "Matching condition to apply on incoming records",
```
Kirdock: Should we also mention that this one doesn't work for the technology type?

dynatrace/api/builtin/openpipeline/processors/fields_add.go
```diff
@@ -0,0 +1,80 @@
...
+func (f *FieldsAddItem) Schema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"name": {
+			Type:        schema.TypeString,
+			Description: "Name of the field",
+			Required:    true,
+		},
+		"value": {
+			Type:        schema.TypeString,
+			Description: " Value to assign to the field",
```
Kirdock:
```suggestion
			Description: "Value to assign to the field",
```

dynatrace/api/builtin/openpipeline/ingestsources/settings/pipeline_reference.go
```diff
@@ -0,0 +1,83 @@
...
+		"pipeline_id": {
+			Type:         schema.TypeString,
+			Required:     false,
+			Description:  "ID of the pipeline. Only used if the pipeline type is \"custom\"",
+			ExactlyOneOf: []string{"pipeline_id", "builtin_pipeline_id"},
+		},
+		"builtin_pipeline_id": {
+			Type:         schema.TypeString,
+			Required:     false,
+			Description:  "ID of the pipeline. Only used if the pipeline type is \"builtin\"",
```
Kirdock: At the moment, this should be sufficient, I guess 🤔. In the documentation, we should then provide a list of allowed values (allowed builtin values per data source)

dynatrace/api/builtin/openpipeline/ingestsources/settings/ingest_source.go
```diff
@@ -0,0 +1,162 @@
...
+	DisplayName   string             `json:"displayName"`
+	Enabled       bool               `json:"enabled"`
+	PathSegment   string             `json:"pathSegment"`
+	StaticRouting *PipelineReference `json:"staticRouting,omitempty"`
+	Processing    *Processing        `json:"processing,omitempty"`
+}
+
+func (is *IngestSource) Schema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"kind": {
```
Kirdock: Looking at the [openpipeline documentation](https://docs.dynatrace.com/docs/discover-dynatrace/platform/openpipeline/use-cases/processing-examples#new-processor), they refer to it as `data type` (`date type` is probably a typo). Should we also call it like this?

dynatrace/api/builtin/openpipeline/ingestsources/settings/ingest_source.go
```diff
@@ -0,0 +1,162 @@
...
+		"path_segment":   &is.PathSegment,
+		"static_routing": &is.StaticRouting,
+		"processing":     &is.Processing,
+	})
+}
+
+func (is *IngestSource) MarshalJSON() ([]byte, error) {
+	var temp = *is
+	if temp.Processing == nil {
+		// The API expected "processing" to be not null.
```
Kirdock: Would it make sense to add a default value then instead?

dynatrace/api/builtin/openpipeline/ingestsources/settings/ingest_source.go
```diff
@@ -0,0 +1,162 @@
...
+		"display_name": is.DisplayName,
+		"path_segment": is.PathSegment,
+	}); err != nil {
+		return err
+	}
+	if is.DefaultBucket != nil {
+		if err := properties.Encode("default_bucket", is.DefaultBucket); err != nil {
+			return err
+		}
+	}
```
Kirdock: Follow-up idea: Should we add a helper function that simply filters/removes the nil values in the map? Then we would not need additional nil handling for each property

dynatrace/api/builtin/openpipeline/ingestsources/settings/ingest_source.go
```diff
@@ -0,0 +1,162 @@
...
+
+type Processing struct {
+	Processors []*processors.Processor `json:"processors,omitempty"`
+}
+
+func (p *Processing) Schema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"processor": {
+			Type:        schema.TypeList,
+			Description: "One processor",
```
Kirdock: Maybe a more descriptive description?

dynatrace/api/builtin/openpipeline/ingestsources/settings/ingest_source.go
```diff
@@ -0,0 +1,162 @@
...
+			Required:    true,
+		},
+		"path_segment": {
+			Type:        schema.TypeString,
+			Description: "The segment of the ingest source, which is applied to the base path",
+			Required:    true,
+		},
+		"static_routing": {
+			Type:        schema.TypeList,
+			Description: "References a pipeline, if present.",
```
Kirdock: We should align the descriptions to either contain or not contain a dot at the end

dynatrace/api/builtin/openpipeline/processors/fields_remove.go
```diff
@@ -0,0 +1,46 @@
...
+type FieldsRemoveAttributes struct {
+	Fields []string `json:"fields"`
+}
+
+func (fra *FieldsRemoveAttributes) Schema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"fields": {
+			Type:        schema.TypeList,
+			Elem:        &schema.Schema{Type: schema.TypeString},
+			Description: "Field to add to the record",
```
Kirdock: Here it's remove, not add

dynatrace/api/builtin/openpipeline/ingestsources/settings/ingest_source_test.go
```diff
@@ -0,0 +1,264 @@
...
+	}{
+		{
+			name:  "minimal-fields-set",
+			input: minimalIngestSource,
+			expected: []byte( // The API demands that "processing" is present, even if it is null, so that seems correct.
+				`{
+					"displayName": "displayName",
+					"enabled": true,
+					"pathSegment": "my.path.segment",
+					"processing": {}
```
Kirdock: Are you sure that this is valid? Looking at the schema, the `processors` array inside of `processing` must be defined

dynatrace/api/builtin/openpipeline/ingestsources/settings/pipeline_reference_test.go
Kirdock, line 1: These tests are actually failing

dynatrace/api/builtin/openpipeline/processors/dql_test.go
```diff
@@ -0,0 +1,98 @@
...
+		},
+		{
+			name: "invalid",
+			input: map[string]interface{}{
+				"script":        "fieldsAdd true",
+				"unknown-field": true,
+			},
+			wantErr: true,
+			errMsg:  "Invalid or unknown key",
+		},
```
Kirdock, line 77: I'm not sure if this is overkill or not, but with this, we're testing the validate and not the marshal/unmarshal, right?

Kirdock (commented):
[nit] should we put the tests into another `*_test` package?

dynatrace/api/builtin/openpipeline/ingestsources/settings/ingest_source_test.go
d0weinberger: I tried it out, it doesn't need to be there.
image

dynatrace/api/builtin/openpipeline/ingestsources/settings/pipeline_reference_test.go
d0weinberger: True, I somehow missed this. Will be fixed.

dynatrace/api/builtin/openpipeline/processors/dql_test.go
d0weinberger: Yes. I wanted to have some tests for the validation properties of the schema

dynatrace/api/builtin/openpipeline/processors/processor.go
d0weinberger: I will add these as well

dynatrace/api/builtin/openpipeline/processors/processor.go
d0weinberger: done

dynatrace/api/builtin/openpipeline/processors/processor.go
d0weinberger: I added as many as I could, but I faced some difficulties. For many constraints we would actually need to use `CustomizeDiff` when creating the resource. Only having the schema, I wasn't able to validate:

- Length of fields given in fieldsRemove, because `ValidateFunc` is ignored for `TypeList`, and therefore not available for `"fields"`, but also the `Elem` used for a `TypeList` must only have the `Type` property set, so also no validation possible here.
- Presence and absence of processor-type specific attributes based on the value of the `type` property. Validation based on the value of another property seems to not be possible.
- This is also the reason why for the `PipelineReference` I could not validate based on the value of `"pipeline_type"`
- I was able to use `ExactlyOneOf` in `PipelineReference` but that only works if you have the whole attribute path from top-down present. And it does absolutely not work if there are collections in-between. The latter point is the reason why I also wasn't able to use `ConflictsWith` for the processor-type-specific attributes, such as `"dql"` or `"fieldsAdd"`, because these processors are in a collection.

dynatrace/api/builtin/openpipeline/processors/processor.go
d0weinberger: Done

dynatrace/api/builtin/openpipeline/ingestsources/settings/ingest_source.go
d0weinberger: I removed the dots

dynatrace/api/builtin/openpipeline/processors/fields_remove.go
d0weinberger: Fixed

Laubi (approved):

dynatrace/api/builtin/openpipeline/processors/processor.go
Kirdock: good enough 🙂👍

dynatrace/api/builtin/openpipeline/ingestsources/settings/ingest_source.go
d0weinberger: I called it kind because that is the way it was called in the old openpipeline implementation, but it makes sense to align with their docs. Will do in a follow-up

dynatrace/api/builtin/openpipeline/ingestsources/settings/ingest_source.go
d0weinberger: I'll experiment with that in a potential follow-up

EDIT: Got an error unfortunately: `processing: Default is not valid for lists or sets`

dynatrace/api/builtin/openpipeline/ingestsources/settings/ingest_source.go
d0weinberger: Good idea! Will experiment with that in follow-up

dynatrace/api/builtin/openpipeline/ingestsources/settings/ingest_source.go
d0weinberger: True, that's not a good description at all. Will take that with me for a follow-up

Kirdock (approved):

PR #759, opened 2025-08-20
fix(export): add validation to exporter to check if credentials are set

tomazpu:
#### **Why** this PR?
Exporting with not set credentials(missconfigured) ended in an unclear log.
```
The environment variable DYNATRACE_TARGET_FOLDER has not been set - using folder 'configuration' as default
Downloading "dynatrace_custom_service_order" Count:  1  
Downloading "dynatrace_direct_share" Count:  0  
Downloading "dynatrace_host_naming_order" Count:  1  
Downloading "dynatrace_oneagent_default_version" Count:  0  
Downloading "dynatrace_openpipeline_business_events" Count:  1  
Downloading "dynatrace_openpipeline_davis_problems" Count:  1  
Downloading "dynatrace_openpipeline_davis_events" Count:  1  
Downloading "dynatrace_openpipeline_metrics" Count:  1  
Downloading "dynatrace_openpipeline_system_events" Count:  1  
Downloading "dynatrace_openpipeline_events" Count:  1  
Downloading "dynatrace_openpipeline_logs" Count:  1  
Downloading "dynatrace_openpipeline_sdlc_events" Count:  1  
Downloading "dynatrace_openpipeline_user_sessions" Count:  1  
Downloading "dynatrace_openpipeline_security_events" Count:  1  
Downloading "dynatrace_openpipeline_spans" Count:  1  
Downloading "dynatrace_openpipeline_user_events" Count:  1  
Downloading "dynatrace_processgroup_naming_order" Count:  1  
Downloading "dynatrace_request_namings" Count:  1  
Downloading "dynatrace_service_naming_order" Count:  1  
Post-Processing Resources ...
Post-Processing Resources - Group child configs with parent configs ...
Finishing touches ...
Writing ___resources___.tf
Writing ___datasources___.tf
Writing main.tf
Writing ___variables___.tf
Writing main ___providers___.tf
Writing modules ___providers___.tf
Remove Non-Referenced Modules ...
Finish Export ...
Terraform executable path:  /opt/homebrew/bin/terraform
Executing 'terraform init'
... finished after 0 seconds
```
From this log, it's unclear whether anything was able to be exported.

#### **What** has changed?
A new credentials type validation, `CredValExport,` is added specifically for export to keep compatibility with other code.

#### **How** does it do it?
Check if the values are set. No specific validation of the validity of strings.

#### How is it **tested**?
No tests where added, manual tested.

Yes, in the future, if the export command is not finding credentials, the user will be presented with a message that points to the problem.
```
The environment variable DYNATRACE_TARGET_FOLDER has not been set - using folder 'configuration' as default
... finished after 0 seconds
 No API Token, Platform Token, or OAuth client has been specified. More detailed information can be found in the documentation at https://registry.terraform.io/providers/dynatrace-oss/dynatrace/latest/docs#configure-the-dynatrace-provider
```

**Issue:**

### DIFF
--- a/dynatrace/export/initialize.go
+++ b/dynatrace/export/initialize.go
@@ -177,7 +177,16 @@ func Initialize(cfgGetter config.Getter) (environment *Environment, err error) {
 
 	var credentials *rest.Credentials
 	configResult, _ := config.ProviderConfigureGeneric(context.Background(), cfgGetter)
-	if credentials, err = config.Credentials(configResult, config.CredValExport); err != nil {
+
+	dwValidationStrategy := config.CredValExport
+	for key, _ := range resArgs {
+		if strings.Contains(key, "iam") {
+			dwValidationStrategy = config.CredValExportIAM
+			break
+		}
+	}
+
+	if credentials, err = config.Credentials(configResult, dwValidationStrategy); err != nil {
 		return nil, err
 	}
 
--- a/dynatrace/export/initialize.go
+++ b/dynatrace/export/initialize.go
@@ -176,9 +176,8 @@ func Initialize(cfgGetter config.Getter) (environment *Environment, err error) {
 	}
 
 	var credentials *rest.Credentials
-
 	configResult, _ := config.ProviderConfigureGeneric(context.Background(), cfgGetter)
-	if credentials, err = config.Credentials(configResult, config.CredValNone); err != nil {
+	if credentials, err = config.Credentials(configResult, config.CredValExport); err != nil {
 		return nil, err
 	}
 
--- a/dynatrace/export/initialize.go
+++ b/dynatrace/export/initialize.go
@@ -180,7 +180,7 @@ func Initialize(cfgGetter config.Getter) (environment *Environment, err error) {
 
 	dwValidationStrategy := config.CredValExport
 	for key, _ := range resArgs {
-		if strings.Contains(key, "iam") {
+		if strings.Contains(key, "_iam_") {
 			dwValidationStrategy = config.CredValExportIAM
 			break
 		}
--- a/provider/config/config.go
+++ b/provider/config/config.go
@@ -46,6 +46,7 @@ const (
 	CredValCluster
 	CredValAutomation
 	CredValNone
+	CredValExport
 )
 
 func validateCredentials(conf *ProviderConfiguration, CredentialValidation int) error {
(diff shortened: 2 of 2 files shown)


provider/config/config.go
```diff
@@ -92,6 +93,15 @@ func validateCredentials(conf *ProviderConfiguration, CredentialValidation int)
 		if len(conf.Automation.EnvironmentURL) == 0 {
 			return fmt.Errorf(" No Environment URL for the Automation API has been specified. Use either the environment variable `DT_AUTOMATION_ENVIRONMENT_URL` or the configuration attribute `automation_env_url` of the provider for that")
 		}
+	case CredValExport:
+		if len(conf.EnvironmentURL) == 0 {
+			return fmt.Errorf(" No Environment URL has been specified. Use either the environment variable `DYNATRACE_ENV_URL` or the configuration attribute `dt_env_url` of the provider for that")
+		}
+
+		if len(conf.APIToken) == 0 && len(conf.Automation.PlatformToken) == 0 && validateCredentials(conf, CredValAutomation) != nil {
```
Kirdock, line 104: This would mean that only downloading IAM resources would not be possible without specifying the Automation env variables, right?

Maybe it would also be good to have a sync here. In the code, the comment before makes me curious
> // We cannot check for the existence of an API Token
// OAuth and Platform Tokens are quite possible

tomazpu: True, I'll check the IAM use case :). Thanks!

dynatrace/export/initialize.go
```diff
@@ -176,9 +176,17 @@ func Initialize(cfgGetter config.Getter) (environment *Environment, err error) {
...
+
+	dwValidationStrategy := config.CredValExport
+	for key, _ := range resArgs {
+		if strings.Contains(key, "iam") {
+			dwValidationStrategy = config.CredValExportIAM
+			break
+		}
+	}
+
+	if credentials, err = config.Credentials(configResult, dwValidationStrategy); err != nil {
```
Kirdock, line 189: What would happen if different types of resources were defined on export? For example: `dynatrace_iam_user`, `dynatrace_segment`, and `dynatrace_alerting_profile`? Would the resulting error message be as good as before or maybe worse?

Also, I think it may be better to assert on `_iam_` instead of just `iam`, just in case there are other resources that contain the name.

tomazpu: I see your point, but after an "IAM" validation, a normal export validation is called, which covers the case of other resources(anything not IAM).

Yes, you can theoretically encounter the case where the first validation returns an error with the account ID, and when you run the command the second time, it prompts you to still check the OAuth or other credentials.

Sure, I can add `_iam_`.

Kirdock: I missed the recursive part; all good 👍

Kirdock (approved):

PR #763, opened 2025-08-21
feat(openpipeline): Improvements for ingest-source HCL handling

d0weinberger:
#### **Why** this PR?
This is a follow-up to #751 introducing some improvements that were discussed on the original PR.

- Nil-values in MarshalHCL: Instead of checking for each field individually, all properties are encoded using `EncodeAll` and nil-values are removed afterwards.
- Descriptions: The description of the `processor` list is improved.

#### How is it **tested**?
Existing tests for marshalling HCL still work. 

It doesn't

**Issue:**
CA-15750

### DIFF
--- a/dynatrace/api/builtin/openpipeline/ingestsource/settings/ingest_source.go
+++ b/dynatrace/api/builtin/openpipeline/ingestsource/settings/ingest_source.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"regexp"
 
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/builtin/openpipeline"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/builtin/openpipeline/processors"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/terraform/hcl"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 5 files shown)


dynatrace/api/builtin/openpipeline/ingestsource/settings/ingest_source.go
```diff
@@ -107,31 +108,18 @@ func (is *IngestSource) Schema() map[string]*schema.Schema {
...
-		if err := properties.Encode("default_bucket", is.DefaultBucket); err != nil {
-			return err
-		}
-	}
-	if is.StaticRouting != nil {
-		if err := properties.Encode("static_routing", is.StaticRouting); err != nil {
-			return err
-		}
-	}
-	if is.Processing != nil && len(is.Processing.Processors) > 0 {
```
Kirdock, line 128: The behavior slightly changed a bit with the removal of `len(is.Processing.Processors)`. Is this a problem?

d0weinberger: It doesn't make a difference, see `properties.go:243`. If `processors` is empty, the encoding result is nil and will be removed with `RemoveNils` which is basically the same outcome as if it wasn't encoded in the first place.

d0weinberger: I added a test case for that

Kirdock (approved):

PR #764, opened 2025-08-21
feat(openpipeline): adds pipeline HCL schema and examples

d0weinberger:
#### **Why** this PR?
This PR adds HCL handling for the OpenPipeline pipeline settings schemas.

#### **What** has changed?

A `Pipeline` schema and HCL marshalling logic, including some validation and tests, have been added. Examples are added as well.

#### **How** does it do it?
The example and the tests could be further improved in follow-ups. Currently, only the `processing` stage is used in tests and examples. The reason for this is that the processors required by the other stages are not yet implemented in the `processors` package.

#### How is it **tested**?
Tests are included.

Currently, it doesn't.

**Issue:**
CA-15946

### DIFF
--- a/dynatrace/api/builtin/openpipeline/pipeline/settings/pipeline.go
+++ b/dynatrace/api/builtin/openpipeline/pipeline/settings/pipeline.go
@@ -1,0 +1,230 @@
+package settings
+
+import (
+	"encoding/json"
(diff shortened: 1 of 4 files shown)


d0weinberger:
Closed, as it became obsolete with the generated code.

PR #767, opened 2025-08-25
feat: openpipeline variant 2 & 3

Kirdock:
#### **Why** this PR?
V2: This is an alternative variant to the current openpipeline pipeline implementation. Instead of having multiple attributes (dql, fieldsAdd, etc.), there are different processors to choose from.
V3: This one is generated

**Issue:**


Kirdock:
Will open another PR with just the generated stuff (without the V2)

PR #768, opened 2025-08-25
chore: deprecate old openpipeline resource

Kirdock:
#### **Why** this PR?
Because of the new dynatrace_openpipeline_v2_* resources, the old one is deprecated and should not be used anymore

#### **What** has changed?
The old resource now prints a warning, when used.
image

#### **How** does it do it?
By adding a deprecation notice to the resource.

They get a deprecation notice when using the old resource.

**Issue:** CA-15951

### DIFF
--- a/dynatrace/api/openpipeline/settings/configuration.go
+++ b/dynatrace/api/openpipeline/settings/configuration.go
@@ -18,6 +18,10 @@ type Configuration struct {
 	Routing        *RoutingTable `json:"routing"`
 }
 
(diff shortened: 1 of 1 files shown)


tomazpu (dismissed):

d0weinberger (approved):

arthurpitman (approved):

d0weinberger:
Let's merge it at the same time as we release our support for OpenPipeline v2

PR #769, opened 2025-08-25
feat(openpipeline): Routing resource

d0weinberger:
#### **Why** this PR?
This PR adds support for the OpenPipeline Routing settings schemas.

#### **What** has changed?
The PR adds the HCL handling and the settings API access for the routing resource.

#### **How** does it do it?
The first commit introduces some example files: A JSON file which represents the payload expected by the settings API and a TF file which shows how the HCL representation of routing resources would look like.

The second commit introduces support for HCL handling of the routing resource.
Schema(), MarshalHCL(), and UnmarshalHCL() are implemented.

The third commit introduces a CRUD service, glueing the routing HCL and the settings API together.

#### How is it **tested**?
Tests for HCL-marshalling and unmarshalling, and HCL-verification are added.
CRUD service tests are still missing.

It doesn't yet. 

**Issue:**
CA-15951

### DIFF
--- a/dynatrace/api/builtin/openpipeline/routing/service.go
+++ b/dynatrace/api/builtin/openpipeline/routing/service.go
@@ -1,0 +1,165 @@
+/**
+* @license
+* Copyright 2025 Dynatrace LLC
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


d0weinberger:
Closed, as it became obsolete with the generated code.

PR #789, opened 2025-09-16
feat(openpipeline): Adds violation path to resource creation and update

d0weinberger:
#### **Why** this PR?
The API for the OpenPipeline settings schemas return error messages, if an unexpected processor type is used in a stage. This message always contains the expected set of processor types. If the whole stage is not supposed to be set for a given data type (such as logs, spans, events, davis.problems ...), i.e. if NO processor type would be valid in the stage, because NO _processor_ would be valid there at all, the error message is something like: 
`Value must be one of []`, meaning: The set of valid processor types is the empty set, meaning there are no valid processors.

This message on its own is already hard to understand, not knowing this context, but it is even harder to understand not knowing the exact stage and the exact processor which violates the API. Finding this out is currently a trial & error - situation. 

This, for example, is an error message that is returned for a call which attempts to create a pipeline, with 20 processors split up across different pipeline stages:

```
module.openpipeline_v2_security_events_pipelines.dynatrace_openpipeline_v2_security_events_pipelines.Test-security_events-yess: Modifying... [id=vu9U3hXa3q0AAAABAC5idWlsdGluOm9wZW5waXBlbGluZS5zZWN1cml0eS5ldmVudHMucGlwZWxpbmVzAAZ0ZW5hbnQABnRlbmFudAAkOGIzM2Q1ZTAtZDE1NC0zNzc2LThhOTUtYzhkNDU5ODRjMzUyvu9U3hXa3q0]
╷
│ Error: Must contain at least 1 non-whitespace character
│ Size must be between 4 and 100
│ 
```

We can deduce, _knowing all of the above_ (which is not self-evident), that we have 4 processors defined which are in a place where they are not valid. Which of the 20 processors are at fault? We don't know.

**Therefore** I want to add - at least for the OpenPipeline V2 resources - always the violation path as well, which states the path of the violating property in the JSON payload, therefore helping us pinpoint the processors which need to be removed.

For a different example, this would look as follows:

```
module.openpipeline_v2_security_events_pipelines.dynatrace_openpipeline_v2_security_events_pipelines.Test-security_events-yess: Modifying... [id=vu9U3hXa3q0AAAABAC5idWlsdGluOm9wZW5waXBlbGluZS5zZWN1cml0eS5ldmVudHMucGlwZWxpbmVzAAZ0ZW5hbnQABnRlbmFudAAkOGIzM2Q1ZTAtZDE1NC0zNzc2LThhOTUtYzhkNDU5ODRjMzUyvu9U3hXa3q0]
╷
│ Error: builtin:openpipeline.security.events.pipelines/0/processing/processors/0/id: Must contain at least 1 non-whitespace character
│ builtin:openpipeline.security.events.pipelines/0/processing/processors/0/id: Size must be between 4 and 100
│ 
```

#### **What** has changed and **how**?
The `Generic` struct gets an additional boolean flag `AlwaysPrintViolationPath` and a new constructing func `NewGenericWithViolationPathPrintOption` which allows setting this flag. For the OpenPipeline resources, this new func is used, setting the flag to `true`. In all other cases, the flag will default to false.

The flag is then used when determining whether the violation path should be written. Previously, this was only the case for very specific types of errors.

I chose the flag over writing it always to keep the impact radius small. There are many resources and use cases depending on `Generic` and I don't want to potentially harm them when trying to solve an OpenPipeline V2-specific issue.

#### How is it **tested**?
It isn't

The feature is not released yet, so it isn't, but as soon as the OpenPipeline V2 feature is released, the error messages returned when using the wrong type of processor in the wrong place will be much more helpful, reducing work for support.

**Issue:**
CA-14694


resources/generic.go
```diff
@@ -47,6 +47,15 @@ func NewGeneric(resourceType export.ResourceType, credVal ...int) *Generic {
 	return &Generic{Type: resourceType, Descriptor: descriptor, CredentialValidation: cv}
 }
 
+func NewGenericWithViolationPathPrintOption(resourceType export.ResourceType, alwaysPrintViolationPath bool, credVal ...int) *Generic {
```
arthurpitman: Could calling the function itself set `alwaysPrintViolationPath` to `true`, rather than via an explicit parameter?

d0weinberger: Sure. You're right, I think that would look a bit nicer, actually. I'll adapt that.

arthurpitman:
Suggestion: let's strip the schema ID from the path, and the first index. This should make the message more readable.

Kirdock (dismissed):

Kirdock (approved):

dynatrace/rest/error.go
```diff
@@ -252,15 +252,23 @@ func (me Error) ViolationMessage() string {
...
 	}
 	return result
 }
 
+func removeSchemaAndObjectIndexFromConstraintViolationPath(path string) string {
+	pathSubstrings := strings.SplitN(path, "/", 3)
+	if len(pathSubstrings) < 3 {
+		return path
+	}
+	return strings.Join(pathSubstrings, "/")
```
Kirdock:
```suggestion
	return strings.Join(pathSubstrings[2:], "/")
```

Kirdock (changes requested):

Kirdock (approved):

PR #794, opened 2025-09-23
Fixed panic and added asynchronous configuration when exporting dynat…

Reinhard-Pilz-Dynatrace:
#### **Why** this PR?
Listing all the available settings for the resource `dynatrace_custom_tags` contained a `panic(err)` when one of the requests is failing.

It shows itself only when a lot of custom tags are about to get exported. Since this code is highly asynchronous - all entity types, all tags and all entities for these tags are getting pulled in parallel - a `Too Many Requests` response is bound to happen sooner or later.

#### **What** has changed?
* The export functionality sends at most 4 concurrent requests for this specific resource
* The number of these concurrent requests is configurable using the environment variable `DYNATRACE_MAX_CONCURRENT_CUSTOM_TAG_LIST_REQUESTS`. Default: 4. Max: 20. Min: 1
* This allows users to speed up the export if necessary
* In case the REST API / API Gateway nevertheless responds with "too many requests", the export doesn't break. The export functionality notes that down in the warnings log

#### **How** does it do it?

* A channel based rate limiter ensures that not more than 4 requests are getting executed at the same time.
* Changes are local to the custom tags service. No other services are affected

#### How is it **tested**?
* Introduced `list_test.go` that validates the rate limitation

The export functionality shouldn't panic anymore when `dynatrace_custom_tags` is getting exported.
We've excluded it from the default set of resources anyways, because it is pretty exotic (not a simple CRUD resource). But customer apparently love to export them nevertheless.

### DIFF
--- a/dynatrace/api/v1/config/customtags/list/list_test.go
+++ b/dynatrace/api/v1/config/customtags/list/list_test.go
@@ -1,0 +1,158 @@
+package list
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/rest"
+)
+
+const (
+	NUM_ENTITY_TYPES = 50
+	NUM_CUSTOM_TAGS  = 50
+	NUM_ENTITIES     = 50
+)
+
+type TestClient struct {
+	latency     time.Duration // artificial per-request delay inside Finish
+	inFlight    int64         // current in-flight
+	maxInFlight int64         // max observed in-flight
+	total       int64         // total Finish calls
+}
+
+func NewTestClient(latency time.Duration) *TestClient {
+	return &TestClient{latency: latency}
+}
+
+func (client *TestClient) Get(ctx context.Context, url string, expectedStatusCodes ...int) rest.Request {
+	return &Request{client: client, ctx: ctx}
+}
+func (client *TestClient) Post(ctx context.Context, url string, payload any, expectedStatusCodes ...int) rest.Request {
+	panic("unsupported operation")
+}
+func (client *TestClient) Put(ctx context.Context, url string, payload any, expectedStatusCodes ...int) rest.Request {
+	panic("unsupported operation")
+}
+func (client *TestClient) Delete(ctx context.Context, url string, expectedStatusCodes ...int) rest.Request {
+	panic("unsupported operation")
+}
+func (client *TestClient) Credentials() *rest.Credentials {
+	panic("unsupported operation")
+}
+
+// Accessors for assertions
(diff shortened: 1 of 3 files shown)


dynatrace/api/v1/config/customtags/list/list_test.go
```diff
@@ -0,0 +1,158 @@
...
+		if cur <= max {
+			break
+		}
+		if atomic.CompareAndSwapInt64(&c.maxInFlight, max, cur) {
+			break
+		}
+	}
+
+	// count total requests
+	atomic.AddInt64(&c.total, 1)
```
Kirdock, line 75: We aren't asserting this one. So, it's just an incremented value that is never validated/used. Either we should validate it or we should get rid of it

dynatrace/api/v1/config/customtags/list/list_test.go
```diff
@@ -0,0 +1,158 @@
...
+func TestCustomTagExport(t *testing.T) {
+	client := NewTestClient(time.Millisecond * 2)
+	_, err := List(t.Context(), client)
+	if err != nil {
+		t.Error(err)
+		return
+	}
+	if got := client.MaxConcurrent(); got > DefaultMaxConcurrent {
+		t.Fatalf("saw %d concurrent requests; want <= 8", got)
+	}
```
Kirdock, line 157: [minor] testify lib usage
```suggestion
	assert.LessOrEqual(t, int64(DefaultMaxConcurrent), client.MaxConcurrent())
```

dynatrace/api/v1/config/customtags/list/list_test.go
```diff
@@ -0,0 +1,158 @@
...
+	"testing"
+	"time"
+
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/rest"
+)
+
+const (
+	NUM_ENTITY_TYPES = 50
+	NUM_CUSTOM_TAGS  = 50
+	NUM_ENTITIES     = 50
```
Kirdock, line 17: Is 50 really needed? `DefaultMaxConcurrent + 1` (5) should be sufficient here, right? It would decrease the execution time a lot

dynatrace/api/v1/config/customtags/list/list_test.go
```diff
@@ -0,0 +1,158 @@
...
+	// update maxInFlight if needed (lock-free CAS loop)
+	for {
+		max := atomic.LoadInt64(&c.maxInFlight)
+		if cur <= max {
+			break
+		}
+		if atomic.CompareAndSwapInt64(&c.maxInFlight, max, cur) {
+			break
+		}
+	}
```
Kirdock, line 72: Is the loop really needed or would this be sufficient?
```suggestion
	max := atomic.LoadInt64(&c.maxInFlight)
	if cur > max {
		atomic.CompareAndSwapInt64(&c.maxInFlight, max, cur)
	}
```

Kirdock (approved):

PR #796, opened 2025-09-23
feat(openpipeline-v1): support "multi value constant" value assignment type

Kirdock:
#### **Why** this PR?
The API also returns and supports a `multiValueConstant` `ValueAssignment` type. Therefore, adjustments are needed so that the export and deploy is working correctly.
Current behavior: Downloading a `multiValueConstant` doesn't set the multi-value-constant value (string array). Deploying it won't work, as the `multiValueConstant` field is required if the type is set.
Now: Download sets the `multi_value_constant` field, and it also deploys it.

#### **What** has changed?
The `multiValueConstant` value type is now supported.

#### **How** does it do it?
By adding a new `multi_value_constant` field

#### How is it **tested**?
Manual tests (download & deploy)

They are now able to download and deploy multi-value-constant values assignments.

**Issue:** CA-16528

### DIFF
--- a/dynatrace/api/openpipeline/settings/processor.go
+++ b/dynatrace/api/openpipeline/settings/processor.go
@@ -1,6 +1,8 @@
 package openpipeline
 
 import (
+	"fmt"
+
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/terraform/hcl"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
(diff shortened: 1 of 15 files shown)


dynatrace/api/openpipeline/settings/processor.go
```diff
@@ -519,24 +520,32 @@ func (ep *ValueAssignment) Schema() map[string]*schema.Schema {
 			Description: "Strategy to assign a value",
 			Optional:    true,
 		},
+		"multi_value_constant": {
+			Type:        schema.TypeList,
+			Elem:        &schema.Schema{Type: schema.TypeString},
+			Description: "Strategy to assign a value",
```
arthurpitman:
```suggestion
			Description: "The constant multi value to assign",
```

dynatrace/api/openpipeline/settings/processor.go
```diff
@@ -497,9 +497,10 @@ func (ep *DavisEventProperty) UnmarshalHCL(decoder hcl.Decoder) error {
 
 type ValueAssignment struct {
 	// Type Defines the actual set of fields depending on the value. See one of the following objects:
-	Type     string  `json:"type"`
-	Field    *string `json:"field"`
-	Constant *string `json:"constant,omitempty"`
+	Type               string   `json:"type"`
```
arthurpitman, line 502: Now that we have three options, `constant`, `field` and `multiValueConstant`, should we validate this, or do you think the API errors are enough?

Kirdock: We do it for some resources, we could also add it for this one via the HandlePreconditions

dynatrace/api/openpipeline/settings/processor.go
Kirdock: Maybe it makes sense to mention that the type has to be set accordingly, so that this field can be used

dynatrace/api/openpipeline/settings/processor.go
Kirdock: done 👍

dynatrace/api/openpipeline/settings/processor.go
Kirdock: done 👍

arthurpitman (approved):

PR #797, opened 2025-09-24
fix(openpipeline-v1): Align with spec

### LINKED ISSUE #782: `matcher` attribute for `technology_processor` in pipeline processing for `dynatrace_openpipeline_logs` doesn't do anything
**Describe the bug**
`matcher` attribute for `technology_processor` in pipeline processing for `dynatrace_openpipeline_logs` doesn't seem to map correctly in the api. From what I can tell, `matcher` should map to `customMatcher` for the technology processor, but it seems to map to `matcher` which doesn't do anything. 

**To Reproduce**
```tf
resource "dynatrace_openpipeline_logs" "this" {
  pipelines {
    pipeline {
      enabled = true
      id = "dotnet"
      display_name = "dotnet"
      processing {
        processor {
          technology_processor {
            enabled = true
            description = "dotnet"
            id = "dotnet"
            matcher = "true"
            technology_id = "applications_dotnet_001"
          }
        }
      }
    }
  }
}
```

**Expected behavior**
`customMatcher` is set to `matcher` for `technology_processor`

**Additional context**
https://github.com/dynatrace-oss/terraform-provider-dynatrace/blob/6cec849f1c68d8b66cb2c86b4849d87835aa01aa/dynatrace/api/openpipeline/settings/processor.go#L543

Kirdock:
#### **Why** this PR?
Several fields were missing, and some were incorrect.
For the Technology processor, `description` and `matcher` can't be set, but `custom_matcher` can.

Fixes #782 

#### **What** has changed?
- Technology processor added (added to span example)
- Cost allocation added
- Product allocation added
- Azure log forwarding processor added
- Security event extraction processor added

#### **How** does it do it?
By adding all missing fields and fixing existing ones.

#### How is it **tested**?
Tests are enabled again, and examples are extended to use all the fields if possible.
Verified locally that the deployed events, spans, and logs example equals the exported config.

Users can now enter everything that is supported by the API spec (API spec and actual allowed values differ; e.g., technology processor is supported by logs but not for events)

**Issue:** CA-16528

### Open question
Do we want to remove the bucket dependency for openpipeline?
The thing is that this one can easily replace unintended fields. If a bucket has the name `custom` and I'm using this in descriptions, the string `custom` is replaced with a bucket reference. For strings/content, this isn't a problem, but for fields like `custom_matcher`.


dynatrace/api/openpipeline/testdata/terraform/example-security-events.tf.disabled
Kirdock, line 1: I renamed and disabled this one for tests as this runs into `Configuration disabled`

dynatrace/testing/api/settingstest.go
```diff
@@ -230,7 +230,7 @@ func TestAcc(t *testing.T, opts ...TestAccOptions) {
 				return
 			}
 			for _, entry := range entries {
-				if !strings.HasSuffix(entry.Name(), ".tf") {
+				if !strings.HasSuffix(entry.Name(), ".tf") || strings.HasSuffix(entry.Name(), "__providers__.tf") {
```
arthurpitman, line 233: Should this be in its own commit?

dynatrace/api/openpipeline/settings/processor_endpoint.go
```diff
@@ -69,6 +69,7 @@ type EndpointProcessor struct {
 	fieldsAddProcessor    *FieldsAddProcessor
 	fieldsRemoveProcessor *FieldsRemoveProcessor
 	fieldsRenameProcessor *FieldsRenameProcessor
+	technologyProcessor   *TechnologyProcessor
```
arthurpitman, line 72: nit: should we make this line last in the struct?

Kirdock: I added it here so it's aligned with the spec
image

arthurpitman (approved):

PR #799, opened 2025-09-25
fix(openpipeline-v2): delete routing configuration

Kirdock:
#### **Why** this PR?
OpenPipeline routing configurations could not be deleted.
Because routing is a single object, and deleting is not supported, we have to update it to the default state

#### **What** has changed?
Routing configurations are now set to default (no routes) on destroy

#### **How** does it do it?
By adding an option to the ServiceOptions to do an update with a certain value instead of a delete.

#### How is it **tested**?
Manual tests

Terraform destroy is now working for routing configurations.

### DIFF
--- a/dynatrace/settings/services/settings20/service_options.go
+++ b/dynatrace/settings/services/settings20/service_options.go
@@ -28,6 +28,8 @@ type ServiceOptions[T settings.Settings] struct {
 	LegacyID       func(id string) string
 	Name           func(ctx context.Context, id string, v T) (string, error)
 	HijackOnCreate func(err error, service settings.RService[T], v T) (*api.Stub, error)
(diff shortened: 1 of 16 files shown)


dynatrace/settings/services/settings20/service_options.go
```diff
@@ -28,6 +28,8 @@ type ServiceOptions[T settings.Settings] struct {
 	LegacyID       func(id string) string
 	Name           func(ctx context.Context, id string, v T) (string, error)
 	HijackOnCreate func(err error, service settings.RService[T], v T) (*api.Stub, error)
+	// instead of calling the delete, it's calling update with the given value. This is useful for single configuration objects (multiObject: false)
```
arthurpitman, line 31:
```suggestion
	// UpdateOnDelete is a value that, if specified, is used to update an object rather than deleting it. This is useful for single configuration objects (multiObject: false).
```

PR #807, opened 2025-09-29
fix(document): Ensure `owner` can't be set and remove non-existent `actor`

arthurpitman:
#### **Why** this PR?
This PR tidies up the `dynatrace_document` resource, removing several inconsistencies with the API

 - As the `owner` field can't be used to update the owner of the document, this is marked as solely computed
 - The `actor` field is removed as the API doesn't actually support this.
 - The `schemaVersion` is removed as it is not used by the API or exposed via HCL.
 
#### How is it **tested**?
Existing acceptance tests.

Very little, except that users will no longer be able to set the `owner` field, which didn't work anyway. Furthermore, an error is now emitted, if the `actor` field is used.

**Issue:** CA-16604

### DIFF
--- a/dynatrace/api/documents/document/settings/document.go
+++ b/dynatrace/api/documents/document/settings/document.go
@@ -57,11 +57,9 @@ func (me *Document) Schema() map[string]*schema.Schema {
 			ValidateDiagFunc: ValidateTypePossibleValues([]string{"dashboard", "notebook", "launchpad"}),
 		},
 		"owner": {
-			Type:             schema.TypeString,
-			Description:      "The ID of the owner of this document",
-			Optional:         true,
-			Computed:         true,
-			ValidateDiagFunc: ValidateUUIDOrEmpty,
+			Type:        schema.TypeString,
+			Description: "The ID of the owner of this document",
+			Computed:    true,
 		},
 		"content": {
 			Type:        schema.TypeString,
--- a/dynatrace/api/documents/document/settings/document.go
+++ b/dynatrace/api/documents/document/settings/document.go
@@ -24,8 +24,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
-const SchemaVersion = 3
-
 type Document struct {
 	Name          string `json:"name" maxlength:"200"`
 	Content       string `json:"content,omitempty"`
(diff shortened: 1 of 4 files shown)


Reinhard-Pilz-Dynatrace (dismissed):

dynatrace/api/documents/document/settings/document.go
```diff
@@ -62,14 +59,13 @@ func (me *Document) Schema() map[string]*schema.Schema {
 			Description:      "The user context the executions of the document will happen with",
 			Optional:         true,
 			Computed:         true,
+			Deprecated:       "Remove this attribute's configuration as it not supported by the API and the attribute will be removed in a future version of the provider.",
```
Kirdock: On the one hand, we're removing the owner field (removed optional and therefore can't be set in the resource), but on the other hand, we're deprecating the actors field 🤔. We should go one way or the other; removing both or deprecating both, I would say

arthurpitman: The `owner` field can still be used as a computed field, so I don't think it is appropriate to deprecate it, or is there another way to do this?

On the other hand, I'd like to eventually remove the `actor` field, but as it appears in exports as an empty string, I don't want to break these and other existing configurations by removing it straight away.

Kirdock: Yes, the owner field can still be used as a computed field, but not inside the resource.
This is currently working and won't produce any error on terraform plan
```hcl
resource "dynatrace_document" "doc" {
  name = "My doc"
  owner = "my-uuid"
  type = "launchpad"
  content = jsonencode({})
  private = false
}
```
If the owner field is not optional anymore, this will produce an error and will therefore be a breaking change. The outcome (breaking change) is like the actor adjustment. So, even if the owner field can still be used via reference `dynatrace_document.doc.owner`, it can't be used inside the resource anymore.

arthurpitman: Let's remove `actor`.

dynatrace/api/documents/document/settings/document.go
```diff
@@ -24,17 +24,14 @@ import (
...
-	Type          string `json:"type"`
-	Actor         string `json:"actor,omitempty" maxlength:"36" format:"uuid"`
-	Owner         string `json:"owner,omitempty" format:"uuid"`
-	Version       int    `json:"version,omitempty"`
-	SchemaVersion int    `json:"schemaVersion,omitempty"`
+	Name      string `json:"name" maxlength:"200"`
+	Content   string `json:"content,omitempty"`
+	IsPrivate bool   `json:"isPrivate,omitempty"`
+	Type      string `json:"type"`
+	Actor     string `json:"actor,omitempty" maxlength:"36" format:"uuid"`
```
Kirdock: The actor can be removed then, right?

tomazpu (approved):

Kirdock (approved):

PR #808, opened 2025-09-30
feat(document): support custom IDs 

Kirdock:
#### **Why** this PR?
The documents API supports setting a custom ID. Therefore, the provider needs to be adapted to support this case.

#### **What** has changed?
An optional `custom_id` can now be set for the document.

#### **How** does it do it?
By adding it to the HCL schema and forwarding it to the dashboard API

#### How is it **tested**?
New tests were added, and the testing utilities were also extended and refactored.

Users can now define a custom ID for their documents.

**Issue:** CA-16591

### Notes to the reviewer
`id` can't be used for the HCL schema, as this seems to be a reserved (import) property

### DIFF
--- a/dynatrace/api/documents/document/service.go
+++ b/dynatrace/api/documents/document/service.go
@@ -26,6 +26,7 @@ import (
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/rest"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/settings"
+
 	docapi "github.com/dynatrace/dynatrace-configuration-as-code-core/api"
 	docclient "github.com/dynatrace/dynatrace-configuration-as-code-core/clients/documents"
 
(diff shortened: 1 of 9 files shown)


arthurpitman (commented):
One thing I notice is that `custom_id` is included for all documents when they are exported. Is this intended / desirable? 

The problem I see is that if the ID was a UUID, that is not really a custom ID, these resources have to be modified before they can be applied, otherwise the API complains: `The value must not be a UUID`.

Kirdock:
> One thing I notice is that `custom_id` is included for all documents when they are exported. Is this intended / desirable?
> 
> The problem I see is that if the ID was a UUID, that is not really a custom ID, these resources have to be modified before they can be applied, otherwise the API complains: `The value must not be a UUID`.

Now `custom_id` isn't set for export if it's a UUID.
https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/808/files#diff-0beaf8bb2cf9ee805b85cb9a72a356fa0af57e1ecc0af67d24f27e1df42e387dR86-R91

I also changed the validation function from `ValidateFunc` to `ValidateDiagFunc` because I saw the other one is deprecated

tomazpu (approved):

arthurpitman (approved):

PR #817, opened 2025-10-06
fix(#786): custom extension monitoring configuration

kishikawa12:


Reinhard-Pilz-Dynatrace (dismissed):

dynatrace/api/v2/hub/extension/config/service.go
```diff
@@ -196,6 +196,9 @@ func (me *service) Create(ctx context.Context, v *extension_config.Settings) (*a
 }
 
 func (me *service) ensureInstalled(ctx context.Context, name string, version string) error {
+	if strings.HasPrefix(name, "custom:") {
```
Kirdock, line 200: Would it make sense to have a check if it's installed before? Right now, we're assuming that every custom extension exists, right?
With the endpoint `GET /extensions/{extensionName}/{extensionVersion}`, we could check if a specific version of an extension is installed and we would only need to call the POST endpoint if it isn't installed and not a custom extension

kishikawa12: @Kirdock That's a good point, but I don't believe we can install it if it's missing?

If you are referring to simply checking for if it's installed or not for returning an error, the proposed `GET /extensions/{extensionName}/{extensionVersion}` and existing `POST /extensions/{extensionName}/monitoringConfigurations` returns the same error: `Extension {name}, version {version} not found or operation forbidden`.

Kirdock: Yes, the next API call would return the same error. I just thought that the `ensureInstalled` should correctly return the info if the extension is installed or not, else the "custom extension installation is ignored" behavior is a bit hidden if you don't know/see the implementation of the `ensureInstalled` function at first glance.

kishikawa12: Makes sense, I'll introduce the GET. Thanks!

Kirdock (approved):

dynatrace/api/v2/hub/extension/config/service.go
```diff
@@ -197,6 +197,13 @@ func (me *service) Create(ctx context.Context, v *extension_config.Settings) (*a
 
 func (me *service) ensureInstalled(ctx context.Context, name string, version string) error {
 	client := rest.HybridClient(me.credentials)
+	if strings.HasPrefix(name, "custom:") {
+		request := client.Get(ctx, fmt.Sprintf("/api/v2/extensions/%s/%s", url.PathEscape(name), url.QueryEscape(version)), 200)
+		request.SetHeader("Accept", "application/json; charset=utf-8")
```
Kirdock, line 202: What's the reason here? Why must this be set for this but not for the other calls?

Reinhard-Pilz-Dynatrace: It is apparently the only REST endpoint so far, that insists on that "Accept" header. No doubt a surprise for us as well.
As for the "why" - we don't have an answer. But we made sure to not introduce any default behavior. Like before, no other REST calls will be sending that header.

PR #823, opened 2025-10-09
feat: Update core library and use `slog` rather than `logr`

arthurpitman:
#### **Why** this PR?
The core library no longer uses `logr`, we'd like to move away from that here too. With this PR, it is no longer a direct dependency.

- Core library dependency is updated to the latest commit
- As a result of updating the core library, some indirect dependencies are updated.
- The settings client is updated to use `slog`

#### How is it **tested**?
Current tests

No change aside from slightly nicer log messages.

**Issue:**  CA-15472


go.sum
```diff
@@ -98,6 +98,8 @@ github.com/dlclark/regexp2 v1.11.4 h1:rPYF9/LECdNymJufQKmri9gV604RvvABwgOA8un7yA
 github.com/dlclark/regexp2 v1.11.4/go.mod h1:DHkYz0B9wPfa6wondMfaivmHpzrQ3v9q8cnmRbL6yW8=
 github.com/dynatrace/dynatrace-configuration-as-code-core v0.9.1-0.20250929124837-64123fbe037b h1:9XjZQIaH6zvytrO4EaNql8B3EvHwO0qus360wcmwK0g=
 github.com/dynatrace/dynatrace-configuration-as-code-core v0.9.1-0.20250929124837-64123fbe037b/go.mod h1:WS25NEP6eSgnnj3Uul4aJKE3ozR6SLeCLcs2f01v1vU=
+github.com/dynatrace/dynatrace-configuration-as-code-core v0.9.1-0.20251009053059-e28f35682982 h1:gHTvrIXV+Nj0Du9Siq9X/s54hovAjJN/MqFmQZP+PBo=
```
Kirdock: Could you do a `go mod tidy`? I assume this will then also remove the dependency in the `go.mod` file

arthurpitman: Done, now updated to the _actual_ commit of the core library. It also updated some other dependencies.

dynatrace/settings/services/settings20/client.go
```diff
@@ -204,14 +203,12 @@ func (c client) Update(ctx context.Context, id string, data []byte) (Response, e
...
 
-	logger := logr.FromContextOrDiscard(ctx)
-
 	var resp *http.Response
 	var err error
 	var responseBody []byte
 
 	for i := 0; i < c.retrySettings.maxRetries; i++ {
-		logger.V(1).Info(fmt.Sprintf("Trying to update setting with id %q (%d/%d retries)", id, i+1, c.retrySettings.maxRetries))
+		slog.DebugContext(ctx, "Trying to update setting", slog.String("objectId", id), slog.Int("retryCount", i+1), slog.Int("maxRetryCount", c.retrySettings.maxRetries))
```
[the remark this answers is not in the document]
arthurpitman: The suggested names are about as good as what is already there, so I'll leave it as is this time.

go.mod
```diff
@@ -37,6 +36,7 @@ require (
 	github.com/cloudflare/circl v1.6.1 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/fatih/color v1.18.0 // indirect
+	github.com/go-logr/logr v1.4.3 // indirect
```
d0weinberger, line 39: 🤔  shouldn't that be gone completely?

arthurpitman: No, I think actually not, as it is used by `github.com/hashicorp/go-hclog`.

d0weinberger (approved):

Kirdock (approved):

PR #826, opened 2025-10-14
fix(openpipeline-v2): set correct enum values

Kirdock:
#### **Why** this PR?
The enum values weren't correctly right. There were some casing issues, e.g., "Bizevent" => "bizevent" or "FieldsAdd" => "fieldsAdd".

#### **What** has changed?
- Enums in descriptions updated
- Also added the comment of the enum property in the schema if there is one.

#### **How** does it do it?
By updating every enum description property in the schema

#### How is it **tested**?
No tests needed, this is just visual.

They see the correct values in the documentation

### DIFF
--- a/dynatrace/api/builtin/openpipeline/bizevents/ingestsources/settings/generic_value_assignment.go
+++ b/dynatrace/api/builtin/openpipeline/bizevents/ingestsources/settings/generic_value_assignment.go
@@ -29,7 +29,7 @@ type GenericValueAssignment struct {
 	Constant           *string                        `json:"constant,omitempty"`           // Constant value
 	Field              *ValueAssignmentFromFieldEntry `json:"field,omitempty"`              // Value from field
 	MultiValueConstant []string                       `json:"multiValueConstant,omitempty"` // Constant multi value
-	Type               AssignmentType                 `json:"type"`                         // Possible Values: `Constant`, `Field`, `MultiValueConstant`
+	Type               AssignmentType                 `json:"type"`                         // Type of value assignment. Possible Values: `constant`, `field`, `multiValueConstant`.
 }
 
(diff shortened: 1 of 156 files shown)


dynatrace/api/builtin/openpipeline/bizevents/ingestsources/settings/generic_value_assignment.go
```diff
@@ -29,7 +29,7 @@ type GenericValueAssignment struct {
 	Constant           *string                        `json:"constant,omitempty"`           // Constant value
 	Field              *ValueAssignmentFromFieldEntry `json:"field,omitempty"`              // Value from field
 	MultiValueConstant []string                       `json:"multiValueConstant,omitempty"` // Constant multi value
-	Type               AssignmentType                 `json:"type"`                         // Possible Values: `Constant`, `Field`, `MultiValueConstant`
+	Type               AssignmentType                 `json:"type"`                         // Type of value assignment. Possible Values: `constant`, `field`, `multiValueConstant`.
```
tomazpu, line 32: A consistency question: I noticed that sometimes the comment is prefixed (`Type of value assignment. Possible Values:....`), and other times it just starts with (`Possible Values: ....`). 

I know it's not changing any behavior, but do we want to align this?

Kirdock: The difference is that those that start with something have a description in the schema. Should I get rid of the schema description for enums?

Kirdock: Team decision: We decided to keep it like that

d0weinberger (approved):

arthurpitman (approved):

PR #828, opened 2025-10-14
feat(openpipeline-v2): add pipeline group resources

Kirdock:
#### **Why** this PR?
To support the upcoming openpipeline pipeline-groups schema

#### **What** has changed?
New pipeline-groups resources added for each kind

#### **How** does it do it?
Generating via schema

#### How is it **tested**?
Manual tests with one example locally. Tests will follow with E2E after this one is merged

It doesn't affect users at the moment, as it's not wired up yet. Still, once it is available, they will be able to use the pipeline-groups feature.

**Issue:** CA-16606


d0weinberger (dismissed):

dynatrace/api/builtin/openpipeline/bizevents/pipelinegroups/schema.json
```diff
@@ -0,0 +1,185 @@
...
+				}
+			},
+			"summaryPattern": "{isTargetPipelinePlaceholder}-{pipelineId}",
+			"type": "object",
+			"version": "0",
+			"versionInfo": ""
+		}
+	},
+	"version": "1.21"
+}
```
tomazpu, line 185: new line missing at the end of file

dynatrace/api/builtin/openpipeline/bizevents/pipelinegroups/service.go
```diff
@@ -0,0 +1,32 @@
+/**
+* @license
+* Copyright 2020 Dynatrace LLC
```
tomazpu, line 3: Do we need to update the date?

dynatrace/api/builtin/openpipeline/bizevents/pipelinegroups/schema.json
Kirdock: I would say this is not really needed for JSON files (there won't be any additions below the object, as this would also lead to an invalid syntax), and because they're generated, I would prefer to either adjust it in the generator or leave it as it currently is :)

dynatrace/api/builtin/openpipeline/bizevents/pipelinegroups/service.go
Kirdock: Because this affects all generated files, I'll create a follow-up PR where I update all license headers of OpenPipeline-V2 👍

Kirdock: Follow up:
https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/848

tomazpu (dismissed):

d0weinberger (dismissed):

arthurpitman (dismissed):

Kirdock:
Too many updates/breaking changes. I'll close this and open a new one

PR #831, opened 2025-10-15
feat(aws-connection): Enables code for the feature

d0weinberger:
#### **Why** this PR?
This PR enables the creation of AWS connections via Terraform which are, for example, used in Workflows.

#### **What** has changed?
The code was already there, but disabled. I enabled it again, did some manual testing with it, and fixed/changed parts of it.

- In `role_arn/service.go:71`, I let `Update` return nil. Only `Create` does anything, which is updating the `role_arn` property of the corresponding connection. 
- In `aws/service.go`, the commented logic for the `Update` did not work after enabling it, because even if an ARN is set in the Terraform resource, it would be overwritten by the ARN set on the server (which is empty). So the created ARN resource did not lead to an update of the connection settings on the server. I removed all of that and just called `me.service.Update`, but need to talk to @Reinhard-Pilz-Dynatrace whether this might be problematic, or to understand the intention of the code that was there before. 
- The call for adding the role ARN needed some retry logic. The reason for this is that when creating the IAM role on AWS, it is not immediately in effect. It can take a few seconds before this is the case. If the AWS connection settings object is updated immediately after IAM role creation, the call will fail, as HAS will fail to assume the role. I implemented a retry loop using `retry.RetryContext`. We retry on server-errors, non-HTTP-errors and errors of code 400 and 404 to handle the case with eventual consistency. The default timeout is 20 minutes. The timeout for the retry loop can be set with a timeouts block, as seen in the example below, where it is set to 30 minutes. One minute (if available) is deducted from the deadline for the retry context to have enough time for finalizing tasks. Example: 

```
resource "dynatrace_aws_connection_role_arn" "test-aws-connection-rolebased-arn" {
  aws_connection_id = dynatrace_aws_connection.test-aws-connection-rolebased.id
  role_arn          = aws_iam_role.example_role_rolebased.arn

  timeouts {
    create = "30m"
  }
}
```

#### **How** does it do it?
The first step is to create a `dynatrace_aws_connection` resource. This also creates the settings object on the server. 
Then, after having created the IAM role in AWS, the `dynatrace_aws_connection_role_arn` resource is used. This resource updates the `role_arn` of the AWS connection settings object on the server.

Since the AWS connection settings object is practically immutable after creation, all attributes and sub-resources of the resource `dynatrace_aws_connection` are flagged as `forceNew`. The `Update` should only ever be called via `role_arn/service` to set the `role_arn`.

So for `dynatrace_aws_connection`, if a terraform resource is
 - created --> `Create` is called, new settings object is created (without role ARN)
 - changed --> because of `forceNew`,  the old resource is deleted and a new one is created
 - removed --> `Delete` is called and the connection is deleted.

For `dynatrace_aws_connection_role_arn`, if a terraform resource is
 - created --> `Create` is called, which calls `Update` for `dynatrace_aws_connection`, so the AWS connection settings object is updated - this should only set the ARN. Other changes, or resetting the ARN, would lead to API errors
 - changed --> returns `nil`.
 - removed --> `Delete` is called but nothing happens - `nil` is returned. Since the role ARN is also immutable once it is set, we cannot update the AWS connection with an empty ARN.

#### How is it **tested**?
Just manual tests right now. No E2E test setup.
An example using web identity is added.
An acceptance test is added, but skipped for now.

Users can now use the TF resources `dynatrace_aws_connection` and `dynatrace_aws_connection_role_arn` to configure AWS connections, which in turn can be used in, e.g. workflows.

**Issue:**
CA-16697


dynatrace/api/builtin/hyperscalerauthentication/connections/aws/testdata/terraform/example_webidentity.tf
```diff
@@ -0,0 +1,43 @@
...
+        {
+            "Effect": "Allow",
+            "Principal": {
+                "Federated": "${aws_iam_openid_connect_provider.dynatrace-oidc-provider.arn}"
+            },
+            "Action": "sts:AssumeRoleWithWebIdentity",
+            "Condition": {
+                "StringEquals": {
+                    "${aws_iam_openid_connect_provider.dynatrace-oidc-provider.url}:sub": "dt:connection-id/${dynatrace_aws_connection.test-aws-connection.id}",
+                    "${aws_iam_openid_connect_provider.dynatrace-oidc-provider.url}:aud": <TENANT_URL>/app-id/dynatrace.aws.connector"
```
Kirdock:
```suggestion
                    "${aws_iam_openid_connect_provider.dynatrace-oidc-provider.url}:aud": "<TENANT_URL>/app-id/dynatrace.aws.connector"
```

dynatrace/api/builtin/hyperscalerauthentication/connections/aws/testdata/terraform/example_webidentity.tf
```diff
@@ -0,0 +1,43 @@
...
+  url = "https://token.dynatrace.com"
+
+  client_id_list = [
+    "<TENANT_URL>/app-id/dynatrace.aws.connector",
+  ]
+}
+
+resource "aws_iam_role" "example_role" {
+  name = "#name#"
+  assume_role_policy = <<EOF
```
Kirdock: Would someting speak against using `jsonencode`? With this we would have a better markdown and it would improve readability.

dynatrace/api/builtin/hyperscalerauthentication/connections/aws/service.go
Kirdock, line 1: Isn't the implementation now as the same as the default?
So, would't it be enough to just have this inside?
```go
func Service(credentials *rest.Credentials) settings.CRUDService[*awsconnection.Settings] {
	return settings20.Service[*awsconnection.Settings](credentials, SchemaID, SchemaVersion)
}

```

dynatrace/api/builtin/hyperscalerauthentication/connections/aws/role_arn/service.go
Kirdock, line 1: The point of this resource is updating the ARN, right?
Does the `role_arn` property then really need to have `ForceNew: true`?

dynatrace/api/builtin/hyperscalerauthentication/connections/aws/role_arn/service.go
```diff
@@ -68,30 +70,36 @@ func (me *service) SchemaID() string {
...
 		connValue.AWSRoleBasedAuthentication.RoleARN = v.RoleARN
 	} else if connValue.AWSWebIdentity != nil {
 		connValue.AWSWebIdentity.RoleARN = v.RoleARN
 	}
-	return me.connService.Update(ctx, id, &connValue)
+
+	var err error
+	var retry = 10
+	for i := 0; i < retry; i++ {
+		if err = me.connService.Update(ctx, v.AWSConnectionID, &connValue); err == nil {
```
Kirdock: [nit] Should we limit it to only retry on 400 status codes?

dynatrace/api/builtin/hyperscalerauthentication/connections/aws/role_arn/service.go
```diff
@@ -68,30 +70,36 @@ func (me *service) SchemaID() string {
...
-	return &api.Stub{ID: v.AWSConnectionID, Name: v.AWSConnectionID}, nil
-}
-
-func (me *service) Update(ctx context.Context, id string, v *role_arn.Settings) error {
 	// we're not allowed to update with an empty roleARN
 	// But we want to avoid errors to get thrown back
 	if v.RoleARN == "" {
-		return nil
+		return nil, errors.New("role_arn must not be empty")
 	}
```
Kirdock: Should we move the validation to the schema?
```go
import (
	"github.com/dynatrace-oss/terraform-provider-dynatrace/terraform/hcl"
	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
)


"role_arn": {
			Type:             schema.TypeString,
			Description:      "The ARN of the AWS role that should be assumed.",
			ForceNew:         true,
			Required:         true,
			ValidateDiagFunc: validation.ToDiagFunc(validation.StringIsNotEmpty),
		},
```

dynatrace/api/builtin/hyperscalerauthentication/connections/aws/service.go
d0weinberger: I think you're right, yes.

dynatrace/api/builtin/hyperscalerauthentication/connections/aws/role_arn/service.go
d0weinberger: Conceptually, the ARN - once set - cannot be changed. It's immutable, so the `Update` operation should actually not be supported. Thinking about this, I'd argue it's better to return an error here. It _should_ not be called, given `ForceNew` being `true`, but just to be safe - let's return an error.

dynatrace/api/builtin/hyperscalerauthentication/connections/aws/role_arn/service.go
d0weinberger: Yes, that's better I think. Will do that 👍

dynatrace/api/builtin/hyperscalerauthentication/connections/aws/role_arn/service.go
d0weinberger: We could, but I don't see any immediaty harm in retrying on every error. Do you?
What I'd like to discuss, though, is: number of retries, duration of timeout and whether these 2 values should be customizable, i.e. should we pass this from the outside via env vars instead of magic numbers? What do you think?

dynatrace/api/builtin/hyperscalerauthentication/connections/aws/testdata/terraform/example_webidentity.tf
d0weinberger: Fixed

dynatrace/api/builtin/hyperscalerauthentication/connections/aws/service.go
d0weinberger: Done

dynatrace/api/builtin/hyperscalerauthentication/connections/aws/testdata/terraform/example_webidentity.tf
d0weinberger: Done

dynatrace/api/builtin/hyperscalerauthentication/connections/aws/role_arn/service.go
d0weinberger: Done

dynatrace/api/builtin/hyperscalerauthentication/connections/aws/role_arn/service.go
```diff
@@ -68,33 +75,88 @@ func (me *service) SchemaID() string {
...
+		remaining := time.Until(dl)
+		if remaining <= 0 {
+			return nil, nil, 0, context.DeadlineExceeded
+		}
+		var retryTimeout time.Duration
+		if remaining > timeoutDeadlineBuffer {
+			retryTimeout = remaining - timeoutDeadlineBuffer
+		} else {
+			retryTimeout = remaining
+		}
```
Kirdock: I'm not quite sure what that does.
`timeoutDeadlineBuffer` is like an expected amount of time of API call (e.g., min time)?
If the remaining time is less than `timeoutDeadlineBuffer` I would expect that the call is not executed. With this `else` here, I don't quite get the point of the `timeoutDeadlineBuffer`

d0weinberger: If the timeout for the whole create-process triggered via Terraform is, say, 10 minutes, then I don't want to have the whole 10 minutes allocated for the (internal) update call to the settings API. I want to give it maybe 9 minutes, so there is still enough time for all the other operations after the update call to finish, before the whole create times out.

Whether to do or not to do the update call if the remaining time is less than what we would like to have for buffer ... yes, that's a fair point. To play it save, we could just not do the call at all instead of using the whole remaining time for the update call.

Kirdock: @arthurpitman could you also have a look?
I would prefer either returning in the `else` branch with `DeadlineExceeded` or not considering a `timeoutDeadlineBuffer` at all

Kirdock (dismissed):
The current implementation solves the issue. The other comment open is minor and doesn't have a big impact on the implementation. So lgtm

Kirdock (dismissed):

Kirdock (approved):

PR #833, opened 2025-10-15
test: add E2E tests and add pipeline-groups to provider

Kirdock:
#### **Why** this PR?
The pipeline groups need to be tested and there need to be some examples

#### **What** has changed?
This adds the resource to the provider; making `dynatrace_openpipeline_v2_<source>_pipelinegroups` available for each source.

This adds the E2E tests for the pipeline groups. The examples are mostly identical.
It's two pipelines, an empty one and one with davis/metricExtractions/securityContext defined, depending on whether it's enabled on that resource.
One pipeline group that creates a group with the stages of a pipeline (the second-mentioned one) and assigns it to another pipeline (the first-mentioned one)

#### **How** does it do it?
By adding examples to pipeline groups and tests that execute them.

#### How is it **tested**?
Every example is tested except the one that are not enabled.

They can now use the new pipeline-groups feature.

**Issue:** CA-16607


Kirdock:
Too many updates/breaking changes. I'll close this and open a new one

PR #836, opened 2025-10-16
feat: allow the use of feature flags

Kirdock:
#### **Why** this PR?
Some features may not be available yet. Therefore, we need to have the ability to enable and disable resources in the provider.

#### **What** has changed?
This allows us to make use of feature flags. With this, features/resources can be enabled/disabled. Some refactoring had to be done because even if `AllResources` is set to the correct values according to the envs, this can't be tested in tests, because at the time we are setting the env in the test, the value is already set

#### **How** does it do it?
- Conditionally adds some resources to the provider and AllResources slice
- The `AllResources` variable changed to a function

#### How is it **tested**?
New tests added.
New ones added here https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/828
Manually tested on https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/833

If the FF is disabled
- export is still working and doesn't include it. If specified it will state unknown resource.
- terraform reports that it's an unknown resource

If the FF is enabled
- export downloads pipeline groups and pipeline groups can by specified via argument
- terraform applies pipeline group resources

So far, no changes for users. Currently, with this PR, there aren't any changes in behavior.

**Issue:** CA-16606

### DIFF
--- a/provider/featureflag/featureflag.go
+++ b/provider/featureflag/featureflag.go
@@ -1,0 +1,65 @@
+package featureflags
+
+import (
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+)
+
+const (
+	OpenPipelinePipelineGroups FeatureFlag = "FEAT_OPENPIPELINE_PIPELINE_GROUP"
+)
+
+var defaultValues = map[FeatureFlag]defaultValue{
+	OpenPipelinePipelineGroups: false,
+}
+
+type (
+	// FeatureFlag represents a command line switch to turn certain features
+	// ON or OFF. Values are read from environment variables defined by
+	// the feature flag. The feature flag can have default value which is used
+	// when the resp. environment variable does not exist
+	FeatureFlag string
+
+	defaultValue = bool
+)
+
+func (ff FeatureFlag) String() string {
+	return ff.EnvName()
+}
+
+// EnvName gives back the environment variable name for the feature flag
+func (ff FeatureFlag) EnvName() string {
+	return string(ff)
+}
+
+// Enabled look up between known temporary and permanent flags and evaluates it.
+// Feature flags are considered to be "enabled" if their resp. environment variable
+// is set to 1, t, T, TRUE, true or True.
+// Feature flags are considered to be "disabled" if their resp. environment variable
+// is set to 0, f, F, FALSE, false or False.
+func (ff FeatureFlag) Enabled() bool {
+	v, exists := defaultValues[ff]
+	if exists {
+		return enabled(ff, v)
+	}
+
+	panic(fmt.Sprintf("unknown feature flag %s", ff))
+}
+
+// enabled evaluates the feature flag.
+// Feature flags are considered to be "enabled" if their resp. environment variable
+// is set to 1, t, T, TRUE, true or True.
+// Feature flags are considered to be "disabled" if their resp. environment variable
+// is set to 0, f, F, FALSE, false or False.
+func enabled(ff FeatureFlag, d defaultValue) bool {
+	if val, ok := os.LookupEnv(ff.EnvName()); ok {
+		value, err := strconv.ParseBool(strings.ToLower(val))
+		if err != nil {
+			return d
+		}
+		return value
+	}
+	return d
+}
--- a/provider/featureflag/featureflag.go
+++ b/provider/featureflag/featureflag.go
@@ -1,37 +1,43 @@
+/*
+ * @license
+ * Copyright 2025 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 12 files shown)


provider/provider_test.go
Kirdock, line 1: I refactored this test file a bit to make use of testify. Tests will be added here, once we have resources behind a FF

provider/featureflag/featureflag.go
```diff
@@ -0,0 +1,65 @@
...
+	"strconv"
+	"strings"
+)
+
+const (
+	OpenPipelinePipelineGroups FeatureFlag = "FEAT_OPENPIPELINE_PIPELINE_GROUP"
+)
+
+var defaultValues = map[FeatureFlag]defaultValue{
+	OpenPipelinePipelineGroups: false,
```
Laubi: QQ: Why don't we define FF with their default values at once?

E.g.

```
var OpenPipeline ... = FeatureFlag{ name: "FEAT_OPENP...GROUP", default: false }
```
imo this would make the code a bit more maintainable, as the default value and FF is in one place, instead of two

provider/featureflag/featureflag.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+type (
+	// FeatureFlag represents a command line switch to turn certain features
+	// ON or OFF. Values are read from environment variables defined by
+	// the feature flag. The feature flag can have default value which is used
+	// when the resp. environment variable does not exist
+	FeatureFlag string
+
+	defaultValue = bool
```
Laubi: Imo this is redundant

provider/featureflag/featureflag.go
```diff
@@ -0,0 +1,65 @@
...
+// is set to 1, t, T, TRUE, true or True.
+// Feature flags are considered to be "disabled" if their resp. environment variable
+// is set to 0, f, F, FALSE, false or False.
+func (ff FeatureFlag) Enabled() bool {
+	v, exists := defaultValues[ff]
+	if exists {
+		return enabled(ff, v)
+	}
+
+	panic(fmt.Sprintf("unknown feature flag %s", ff))
```
Laubi: This panic would not be necessary with having the default value inside the FF definition, and then we can also easily combine `enabled` and `Enabled`

provider/featureflag/featureflag_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"testing"
+
+	featureflags "github.com/dynatrace-oss/terraform-provider-dynatrace/provider/featureflag"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFeatureFlag(t *testing.T) {
+	ff := featureflags.OpenPipelinePipelineGroups // here can be any FF with false as default
+
+	for _, fv := range []string{"0", "f", "F", "FALSE", "false", "False", "fAlSe", "", "othervalue"} {
```
Laubi: Don't use shotgun testing, use table tests instead (separate runs at least)

provider/featureflag/featureflag_test.go
```diff
@@ -0,0 +1,63 @@
...
+		assert.False(t, ff.Enabled())
+	}
+
+	for _, tv := range []string{"1", "t", "T", "TRUE", "true", "tRuE", "True"} {
+		t.Setenv(ff.EnvName(), tv)
+		assert.True(t, ff.Enabled())
+	}
+}
+
+func TestFeatureFlagIDEnabled(t *testing.T) {
```
Laubi: I'd split this into two distinct test functions

Laubi (commented):
I am concerned that we refactor a lot of calls (`AllResources`) without having the current need to do so - the feature flag in the code is not used. What is the impact? How often is it called? Is is neglectable?

Kirdock:
@Laubi I just took the implementation (and tests) of Monaco, so that the FF management is aligned. I would have valued alignment more than a slightly better one. I can adapt it of course.
FF in use here: https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/828/commits/3188a7943e364e65e1e253332619a533bd4302a1.

The reason for making it a function is already written down in the PR description
> Some refactoring had to be done because even if `AllResources` is set to the correct values according to the envs, this can't be tested in tests, because at the time we are setting the env in the test, the value is already set.

The test here, would panic because `enums.go` would have a different value than the provider, because the env is already set when loading the file; in the test it's already too late to set it.
https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/828/files#diff-978ff4fdc176f5a02cba8681ff7f0c943bf0740a2ca7a68ff4e9f08d1efd3744R157

Kirdock:
I refactored the logic a bit now. Now, `AllResources` doesn't have to be a function and for the provider, where the panic was, the resources are lazy-evaluated (new created).

PR #839, opened 2025-10-20
fix: Active Gate URL parsing

### LINKED ISSUE #832: Not possible to use Dynatrace ActiveGate REST API module with Terraform
**Describe the bug**

It is not possible to use the Terraform provider with the Dynatrace ActiveGate [REST API module](https://docs.dynatrace.com/docs/ingest-from/dynatrace-activegate/configuration/configure-activegate#api_mod).

In restricted enterprise networks with very limited access to the internet, it is common to use the Dynatrace ActiveGate with the [REST API module](https://docs.dynatrace.com/docs/ingest-from/dynatrace-activegate/configuration/configure-activegate#api_mod) enabled to comunicate with the public Dynatrace API.

**To Reproduce**

The following does not work:

```hcl
provider "dynatrace" {
  account_id = "..."

  dt_env_url   = "https://load-balancer.******.intranet:9999/e/my-environment-id"
  dt_api_token = "*****"
}
```

It seems that the Terraform provider is doing some magic URL parsing which is totally unnecessary in our case:

```
.../e/my-environment-id/e/my-environment-id/api/v2/settings/objects...
```

Image

**Expected behavior**

It should be possible to use Dynatrace ActiveGate in the intranet with the Terraform provider.

Kirdock:
#### **Why** this PR?
Active Gate URLs pointing to non UUID environment IDs weren't working. For `myhost:9999/e/my-env-id` the classic requests always attached another `/e/my-env-id`.

#### **What** has changed?
The URL is now correctly parsed. Active Gate URLs should work again.
The error handling for requests also slightly changed a bit. Errors of `POST/PUT/PATCH` request were swallowed. Because of this the error message pointed to the correct URL but the real request was pointing to a different one (the one with the additional `/e/env_id`).

#### **How** does it do it?
Instead of conditionally removing a URL path prefix, the initial path of the provided dt_env_url is not added again in the first place. The API client that's in use already contains this information (which host and which initial path like active gate). That's why the path was attached a second time.

#### How is it **tested**?
Added another test that checks for the correct URL

They should now be able to use Terraform when using Active Gate.

**Issue:** CA-16828, #832

### DIFF
--- a/dynatrace/rest/hybrid_client_test.go
+++ b/dynatrace/rest/hybrid_client_test.go
@@ -3,8 +3,15 @@ package rest
 import (
 	"context"
 	"errors"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
 	"strings"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 const mockToken = "########"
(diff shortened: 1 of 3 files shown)


dynatrace/rest/hybrid_client_test.go
```diff
@@ -141,6 +147,24 @@ func TestHybridClient(t *testing.T) {
 	}
 }
 
+func TestHybridClientActiveGateRequest(t *testing.T) {
```
tomazpu: Would be nice to also add some test that fail (classic url == 0, and url.Parse fail). 

I would also add some headers to the test.

Kirdock: done 👍

tomazpu (approved):

realtimetodie:
When will this bugfix be released?

PR #857, opened 2025-11-03
docs: update resources doc for release

Kirdock:
#### **Why** this PR?
The docs (md files) need to be updated (generated based on the tmpl files).

#### **What** has changed?
Docs are updated

#### **How** does it do it?
- By adding the template files for the resources `dynatrace_aws_connection` and `dynatrace_aws_role_arn`
- By running the generator

They will see the updated doc once published.

**Issue:**


tomazpu (dismissed):

docs/resources/aws_connection_role_arn.md
```diff
@@ -0,0 +1,115 @@
+---
+layout: ""
+page_title: "dynatrace_aws_role_arn Resource - terraform-provider-dynatrace"
+subcategory: "Connections"
```
d0weinberger, line 4: should be `dynatrace_aws_connection_role_arn` here an in the rest of the doc

docs/resources/aws_connection_role_arn.md
```diff
@@ -0,0 +1,115 @@
...
+### Read-Only
+
+- `id` (String) The ID of this resource.
+
+<a id="nestedblock--timeouts"></a>
+### Nested Schema for `timeouts`
+
+Optional:
+
+- `create` (String)
```
d0weinberger, line 115: We could link to the hashicorp docs here https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/retries-and-customizable-timeouts

templates/resources/aws_connection_role_arn.md.tmpl
```diff
@@ -0,0 +1,51 @@
+---
+layout: ""
+page_title: "dynatrace_aws_role_arn Resource - terraform-provider-dynatrace"
```
d0weinberger: `dynatrace_aws_connection_role_arn`

documentation/supported-resources.md
```diff
@@ -34,7 +34,9 @@ Dynatrace Configuration as Code via Terraform supports the following resources,
...
+| dynatrace_aws_connection                                | /api/v2/settings/objects (schema: builtin:hyperscaler-authentication.connections.aws)                                                                                                          | settings.read, settings.write                                                                     |
 | dynatrace_aws_credentials                               | /api/config/v1/aws/credentials                                                                                                                                                                 | ReadConfig, WriteConfig                                                                           |
+| dynatrace_aws_role_arn                                  | /api/v2/settings/objects (schema: builtin:hyperscaler-authentication.connections.aws)                                                                                                          | settings.read, settings.write                                                                     |
```
d0weinberger: `dynatrace_aws_connection_role_arn`

docs/resources/aws_connection.md
```diff
@@ -0,0 +1,123 @@
...
+
+# dynatrace_aws_connection (Resource)
+
+-> This resource requires the API token scopes **Read settings** (`settings.read`) and **Write settings** (`settings.write`)
+
+-> This resource requires the OAuth scopes **Read settings** (`settings:objects:read`) and **Write settings** (`settings:objects:write`)
+
+## Requirements
+This resource must be used in combination with the `dynatrace_aws_role_arn` resource.
+Ensure you configure both resources together for a valid AWS connection.
```
d0weinberger, line 17: Should we go into more detail here? I.e. "first create dynatrace_aws_connection, then create the aws_iam_role using the settings object id, aftewards create the dynatrace_aws_connection_role_arn" or is this anyways apparent fromt the example?

docs/resources/aws_connection_role_arn.md
Kirdock: This block is generated by the doc-plugin. Not sure if it is even possible to put a description here. We can only set a duration for the timeout but no description. I would not put custom text in here, as this leads to inconsistencies and will easily be overwritten and overseen when we generate the docs

docs/resources/aws_connection.md
Kirdock: I would say the example should be enough, as we mention below, `An example of how to set up both resources can be found in the Resource Example Usage`

docs/resources/aws_connection_role_arn.md
d0weinberger: That's unfortunate, but since it's a built-in Terraform concept, we can maybe assume that users know it.

d0weinberger (approved):

PR #863, opened 2025-11-06
fix(export): list-exclusions cannot be set with other flags anymore

d0weinberger:
#### **Why** this PR?
Previously it was allowed to set other flags together with `-list-exclusions`, e.g.,
`terraform-provider-dynatrace -export -list-exclusions -import-state`
This was confusing because actually after the exclusions were listed the command would just return and no actual export was done, making the `-import-state` flag useless. This behavior was not properly documented in our [documentation](https://docs.dynatrace.com/docs/deliver/configuration-as-code/terraform/terraform-cli-commands#options-and-flags). 

#### **What** has changed?
The first necessary change is an improvement of the documentation, which is in the works.

To make it less confusing, I changed the behavior such that you get an error if another flag is listed together with `list-exclusions`, explaining its usage. 

##### Examples

**Command: `-export -list-exclusions`**
-> Exclusion list is printed

**Command: `-export -list-exclusions -import-state`**
-> `-list-exclusions cannot be combined with other flags
Usage: terraform-provider-dynatrace -export -list-exclusions`

**Command: `-export -import-state -list-exclusions`**
-> `-list-exclusions cannot be combined with other flags
Usage: terraform-provider-dynatrace -export -list-exclusions`

Using `-export -list-exclusions <any-other-flag>` now gives an error message which also explains the proper usage.

**Issue:**
CA-17016

### DIFF
--- a/dynatrace/export.go
+++ b/dynatrace/export.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/export"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider/config"
+	"golang.org/x/exp/slices"
 )
 
 func Export(args []string, cfgGetter config.Getter) bool {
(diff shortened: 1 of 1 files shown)


arthurpitman (dismissed):
I'm fine with this, as long as we're happy with the change in behavior 👍

dynatrace/export.go
```diff
@@ -38,6 +38,11 @@ func Export(args []string, cfgGetter config.Getter) bool {
 
 	if len(args) > 2 {
 		if strings.TrimSpace(args[2]) == "-list-exclusions" {
+			if len(args) > 3 {
```
Kirdock: This won't trigger if the list-exclusions is the third argument, right? Should we move the list-exclusions flag to the `createFlags` function, add some mutual exclusions and early exit for this one?

d0weinberger: True, if it's anywhere else you would get `flag provided but not defined: -list-exclusions`
To conveniently handle mutual exclusions, I think we would need to switch to cobra, right?

If needed, I could make the handling of flags more sophisticated, yes. The suggest change in this PR is a minimal solution. The more important change is the docs change in my opinion.

Kirdock: > To conveniently handle mutual exclusions, I think we would need to switch to cobra, right?

You're right, the flags library doesn't seem to support mutual exclusions. Then let's leave it here.

Should we maybe add a simple change to handle the check for the `list-exclusions` flag better?
```go
if slices.ContainsFunc(args, func(arg string) bool { return strings.TrimSpace(arg) == "-list-exclusions" }) {
```
We could then also remove the first if with `len(args) > 2`

Kirdock (approved):

PR #872, opened 2025-11-11
fix: Classic OAuth client can be `nil` if not available

arthurpitman:
#### **Why** this PR?
The implementations of `dynatrace_davis_anomaly_detectors`, `dynatrace_generic_setting`, `dynatrace_site_reliability_guardian`, and `dynatrace_automation_workflow_slack` resources create a settings client that always has an OAuth client regardless of whether credentials are available or not.

This OAuth client doesn't work. For settings resources that must use OAuth, in this case `dynatrace_davis_anomaly_detectors`, this implementation leads to the original error message being discarded and more confusing error mentioning OAuth:

```
module.davis_anomaly_detectors.dynatrace_davis_anomaly_detectors.Potential_log_data_anomaly: Creating...
╷
│ Error: failed to create object: Post "https://....dynatrace.com/api/v2/settings/objects?repairInput=true": oauth2: cannot fetch token: 400 Bad Request
│ Response: {"errorCode":400,"message":"Bad Request","issueId":"..."}
│ 
``` 

Note: the exact error message depends on which operation is being performed (in this instance create).

With the fix in this PR, no OAuth client is created if no credentials are specified, and so the following is returned:

```
module.davis_anomaly_detectors.dynatrace_davis_anomaly_detectors.Potential_log_data_anomaly: Creating...
╷
│ Error: Could not do validation as request was not done using oAuth.
```

- `CreateClassicOAuthBasedClient()` now returns `nil` if no OAuth client ID or secret is available. By returning `nil` in this situation, the function no longer returns a client that has an API token set. This is OK because in all instances where this client is used, an explicit client with API token set is created and used first. The function also no longer returns an error as this was always dropped.
-  Improves error handling in `CreateClassicClient()`. The `classicClientCache` is not read if a URL is not specified. The cache is also not written to if the client could not be created.

#### How is it **tested**?
NA

Users will see the correct error

**Issue:** CA-17061

### DIFF
--- a/dynatrace/rest/api_token_client.go
+++ b/dynatrace/rest/api_token_client.go
@@ -139,6 +139,11 @@ func CreateClassicOAuthBasedClient(ctx context.Context, credentials *Credentials
 }
 
 func CreateClassicClient(classicURL string, apiToken string) (*rest.Client, error) {
+	if classicURL == "" {
+		// sanity check - this should not be empty anymore at this point
+		return nil, NoClassicURLDefinedErr
+	}
+
 	classicClientCacheMutex.Lock()
 	defer classicClientCacheMutex.Unlock()
 
@@ -153,10 +158,13 @@ func CreateClassicClient(classicURL string, apiToken string) (*rest.Client, erro
 	factory = factory.WithHTTPListener(logging.HTTPListener("classic "))
 
 	client, err := factory.CreateClassicClient()
+	if err != nil {
+		return nil, err
+	}
 
 	classicClientCache[classicURL] = client
 
-	return client, err
+	return client, nil
 }
 
 func (me *classic_request) Finish(optionalTarget ...any) error {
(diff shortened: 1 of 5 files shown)


arthurpitman:
With the exception of the MonitorOutageHandling tests, the tests are green

dynatrace/rest/api_token_client.go
```diff
@@ -131,11 +135,15 @@ func CreateClassicOAuthBasedClient(ctx context.Context, credentials *Credentials
 	)
 
 	oauthClient.SetHeader("User-Agent", version.UserAgent())
-	oauthClient.SetHeader("Authorization", "Api-Token "+credentials.Token)
```
Kirdock, line 134: Should we then also change the logging to `platform`?
```go
		crest.WithHTTPListener(logging.HTTPListener("platform")),
```

Kirdock:
So, when looking at the underlying open issue, this means that invalid OAuth credentials will result in the same behavior as before when this PR is merged, but if there isn't any OAuth defined, it would lead to a different message (the request to classic requires OAuth), right?

arthurpitman:
> So, when looking at the underlying open issue, this means that invalid OAuth credentials will result in the same behavior as before when this PR is merged, but if there isn't any OAuth defined, it would lead to a different message (the request to classic requires OAuth), right?

Correct. It's all about that case: if no OAuth credentials are defined, no OAuth client is created, meaning that no OAuth requests are made in these resources. These operations stop after trying and failing with an access token, and that error is returned.

Kirdock (approved):

PR #874, opened 2025-11-12
doc: Update documentation ready for release

arthurpitman:
#### **Why** this PR?
The documentation `.md` files need to be updated based on the `.tmpl` files.

#### **What** has changed?
Documentation is updated.

#### **How** does it do it?
This PR is the changes made by running:
```
go run github.com/hashicorp/terraform-plugin-docs/cmd/tfplugindocs
```

#### How is it **tested**?
NA

Users will see the updated documentation once it is published.

**Issue:** NA

### DIFF
--- a/docs/resources/autotag_rules.md
+++ b/docs/resources/autotag_rules.md
@@ -83,7 +83,7 @@ Required:
 Optional:
 
 - `attribute_rule` (Block List, Max: 1) no documentation available (see [below for nested schema](#nestedblock--rules--rule--attribute_rule))
-- `entity_selector` (String) The documentation of the entity selector can be found [here](https://dt-url.net/apientityselector).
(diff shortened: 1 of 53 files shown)


provider/version/version.go
```diff
@@ -24,7 +24,7 @@ import "fmt"
 var (
 	// Version is a string representing the current version of the Dynatrace Terraform provider.
 	// This should be in `major.minor.patch` format, with prereleases having an optional suffix introduced by a dash, for example, `1.2.0-beta`.
-	Version = "(unknown-version)"
+	Version = "1.87.1"
```
Kirdock: I don't think this change was intended, right? 😅

arthurpitman: 😅 no, not intended

Kirdock (approved):

Issue #868, opened 2025-11-07
Provider produces inconsistent result when applying dynatrace_oneagent_features


mattBaumBeneva:
**Describe the bug**
When applying the following code, apply fails:

```
resource "dynatrace_oneagent_features" "JAVA_UEM_INSTRUMENTATION" {
  enabled = false
  key     = "JAVA_UEM_INSTRUMENTATION"
  scope   = "environment"
}
```

The error message is:
```
08:29:30  │ Error: Provider produced inconsistent result after apply
08:29:30  │ 
08:29:30  │ When applying changes to
08:29:30  │ module.main.module.oneagent_features[0].dynatrace_oneagent_features.JAVA_UEM_INSTRUMENTATION,
08:29:30  │ provider "provider[\"registry.terraform.io/dynatrace-oss/dynatrace\"]"
08:29:30  │ produced an unexpected new value: Root object was present, but now absent.
08:29:30  │ 
08:29:30  │ This is a bug in the provider, which should be reported in the provider's
08:29:30  │ own issue tracker.
```

The plan contains:
```
13:10:07    # module.main.module.oneagent_features[0].dynatrace_oneagent_features.JAVA_UEM_INSTRUMENTATION will be created
13:10:07    + resource "dynatrace_oneagent_features" "JAVA_UEM_INSTRUMENTATION" {
13:10:07        + _restore_ = (known after apply)
13:10:07        + enabled   = false
13:10:07        + id        = (known after apply)
13:10:07        + key       = "JAVA_UEM_INSTRUMENTATION"
13:10:07        + scope     = "environment"
13:10:07      }
```

**To Reproduce**
1. Apply configuration
2. See error

**Expected behavior**
The configuration should be applied without error.

**Additional context**
We did not have any errors when applying with provider version 1.86.0 on november 3rd 2025. Today, we cannot complete an apply with versions 1.86.0, 1.87.0 or 1.87.1.
Has there been a Dynatrace API change causing this error?

tomazpu:
Hey @mattBaumBeneva, 

After investigating your issue, I found out that "JAVA_UEM_INSTRUMENTATION" was downgraded to experimental.

The expected behavior is that the feature is not customer-facing anymore and can only be accessed by devops.

From the provider's perspective, we will improve the error message provided.

Hope this helps! Tomaz

PR #878, opened 2025-11-13
fix: Improve error message on Read

tomazpu:
#### **Why** this PR?
This PR is connected to the GitHub issue #868, where an API, after a successful POST request, returns a 404 on the consequent read operation with the objectId.

 As there is a state mismatch, Terraform will fail without an error message.

#### **What** has changed?

1. Removed unused code block
2. Introduced APIError 
3. Using the APIError and removing the silent failing on a 404 error

#### How is it **tested**?
Manually tested and covered by tests

The user should be provided with a better error message. Example below.
`│ Error: API Error: <GET> Settings modification only visible for devops`
**Issue:**

### DIFF
--- a/resources/generic.go
+++ b/resources/generic.go
@@ -468,15 +468,7 @@ func (me *Generic) Read(ctx context.Context, d *schema.ResourceData, m any) diag
 			return diag.FromErr(err)
 		}
 	}
-	if err != nil {
-		if restError, ok := err.(rest.Error); ok {
-			if restError.Code == 404 {
-				d.SetId("")
-				return diag.Diagnostics{}
-			}
-		}
-		return diag.FromErr(err)
-	}
+
 	return me.ReadForSettings(ctx, d, m, sttngs)
 }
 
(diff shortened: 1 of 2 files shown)


resources/generic.go
```diff
@@ -458,25 +458,12 @@ func (me *Generic) Read(ctx context.Context, d *schema.ResourceData, m any) diag
...
-	if err != nil {
-		if restError, ok := err.(rest.Error); ok {
-			if restError.Code == 404 {
-				d.SetId("")
-				return diag.Diagnostics{}
-			}
 		}
-		return diag.FromErr(err)
+
+		d.SetId("")
```
Kirdock, line 463: We have several behavior changes in here. It's not just the `404 shows an error` fix.

1. `d.SetId("")` and `diag.FromErr` is now always called, even if the `service.Get(ctx, d.Id(), sttngs)` before succeeds (the one inside the `re-run with confighcl` check)
2. `setId` is also called if it's a non-404 status code (before it was only for 404). Not sure if this is good or bad. If there is a server outage (e.g., 503), would this mean that the resource is removed from the state file?
3. If you create a resource via Terraform and then delete it via the API (outside of Terraform), will the next `terraform apply` produce an error instead of suggesting to create it?

PR #879, opened 2025-11-14
test: add build tags and minimize pipeline

Kirdock:
#### **Why** this PR?
The test pipeline has numerous redundant steps. Some (unit) tests aren't even executed because they would need to be manually added to the workflows file. Some tests even are a mix of unit and integration.

#### **What** has changed?
- Mixes of unit and integration tests are now split
- Go build tags are introduced (unit and integration)
- Removed some dead code
- Replaced the custom assert function for `json_dashboard` with the built-in one.
- Simplified the test pipeline. One step for unit tests and one for integration tests.

#### **How** does it do it?
- By splitting unit and integration tests into `service.go` (integration) and `service_unit_test.go`
- Adding the Go build tag `unit` to unit tests
- Adding the Go build tag `integration` to integration tests
- By removing all steps from the testing pipeline and replacing it with one step that executes `unit` tests and one that executes `integration` tests

#### How is it **tested**?
Pipeline still succeeds and runs all tests

Doesn't affect users.

**Issue:** CA-16363


.github/workflows/tests.yml
```diff
@@ -4,4616 +4,50 @@ on:
...
-  build:
-    name: Build
+  unit-tests:
+    name: Unit Tests
     runs-on: ubuntu-latest
     steps:
-      - name: Set up Go 1.19
-        uses: actions/setup-go@v3
+      - name: Set up Go
+        uses: actions/setup-go@v4
```
tomazpu: We should pin the actions that we're using to a commit.

dynatrace/api/builtin/hyperscalerauthentication/connections/aws/role_arn/service_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build unit
+
```
tomazpu, line 2: file is missing license header

dynatrace/api/builtin/hyperscalerauthentication/connections/aws/service_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build integration
+
```
tomazpu, line 2: file is missing license header

dynatrace/api/v1/config/customtags/list/list_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build unit
+
```
tomazpu, line 2: file is missing license header

dynatrace/api/v2/settings/objects/permissions/convert/convert_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build unit
+
```
tomazpu, line 2: file is missing license header

dynatrace/api/v2/settings/objects/permissions/reconcile/reconcile_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build unit
+
```
tomazpu, line 2: file is missing license header

dynatrace/settings/services/cache/cache_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build unit
+
```
tomazpu, line 2: file is missing license header

dynatrace/rest/hybrid_client_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build unit
+
```
tomazpu, line 2: file is missing license header

dynatrace/rest/admin_retry_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build unit
+
```
tomazpu, line 2: file is missing license header

dynatrace/api/v2/settings/objects/permissions/settings/user_accessor_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build unit
+
```
tomazpu, line 2: file is missing license header

dynatrace/api/v2/settings/objects/permissions/settings/setting_permissions_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build unit
+
```
tomazpu, line 2: file is missing license header

dynatrace/api/v2/settings/objects/permissions/settings/group_accessor_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build unit
+
```
tomazpu, line 2: file is missing license header

dynatrace/api/builtin/hyperscalerauthentication/connections/aws/role_arn/service_test.go
Kirdock: It's out of scope for this ticket, as all newly introduced files have a license header.
I could do a follow-up after this one is merged. This affects all files, not just the test files.
Follow up: CA-17117

dynatrace/api/v1/config/customtags/list/list_test.go
Kirdock: It's out of scope for this ticket, as all newly introduced files have a license header.
I could do a follow-up after this one is merged. This affects all files, not just the test files.
Follow up: CA-17117

dynatrace/api/v2/settings/objects/permissions/convert/convert_test.go
Kirdock: It's out of scope for this ticket, as all newly introduced files have a license header.
I could do a follow-up after this one is merged. This affects all files, not just the test files.
Follow up: CA-17117

dynatrace/api/v2/settings/objects/permissions/reconcile/reconcile_test.go
Kirdock: It's out of scope for this ticket, as all newly introduced files have a license header.
I could do a follow-up after this one is merged. This affects all files, not just the test files.
Follow up: CA-17117

dynatrace/settings/services/cache/cache_test.go
Kirdock: It's out of scope for this ticket, as all newly introduced files have a license header.
I could do a follow-up after this one is merged. This affects all files, not just the test files.
Follow up: CA-17117

dynatrace/rest/hybrid_client_test.go
Kirdock: It's out of scope for this ticket, as all newly introduced files have a license header.
I could do a follow-up after this one is merged. This affects all files, not just the test files.
Follow up: CA-17117

dynatrace/rest/admin_retry_test.go
Kirdock: It's out of scope for this ticket, as all newly introduced files have a license header.
I could do a follow-up after this one is merged. This affects all files, not just the test files.
Follow up: CA-17117

.github/workflows/tests.yml
Kirdock: done

dynatrace/api/builtin/hyperscalerauthentication/connections/aws/service_test.go
Kirdock: It's out of scope for this ticket, as all newly introduced files have a license header.
I could do a follow-up after this one is merged. This affects all files, not just the test files.
Follow up: CA-17117

tomazpu (dismissed):

d0weinberger:
You added `-p 1` for all integration tests now whereas previously it was only used for the OpenPipelineV2 tests. This leads to a significant increase in test duration, right? The integration test stage on this PR is taking 51 minutes, whereas for other PRs with the old test setup I see lower numbers, e.g. 38 minutes here: https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/878

Would it make sense to go with two jobs and two sorts of tags? `integration-test` and `integration-test-nonparallel` or something like that? Or would we, for the sake of simplicity, rather take the increase in test duration?

d0weinberger:
Other changes I noticed:

For `TestLogCustomAttributes` and `TestLogCustomSource` we did not set `DT_NO_REPAIR_INPUT: "false"`, but I imagine this does not matter too much.

For `TestPlatformBuckets` we now miss the following env vars that previously were set:
```
DYNATRACE_LOG_HTTP: "terraform-provider-dynatrace.log"
DYNATRACE_HTTP_RESPONSE: true 
```

d0weinberger (dismissed):

Kirdock:
> You added `-p 1` for all integration tests now whereas previously it was only used for the OpenPipelineV2 tests. This leads to a significant increase in test duration, right? The integration test stage on this PR is taking 51 minutes, whereas for other PRs with the old test setup I see lower numbers, e.g. 38 minutes here: #878
> 
> Would it make sense to go with two jobs and two sorts of tags? `integration-test` and `integration-test-nonparallel` or something like that? Or would we, for the sake of simplicity, rather take the increase in test duration?

Even with `-p 2` EOF errors occurred. Before it probably wasn't an error because they were separate executions (dedicated cache, client, etc.). I could try splitting it and having some `go test` with `go list` and `grep` so that it can be split from a-h, i-o, etc.

Kirdock:
> Other changes I noticed:
> 
> For `TestLogCustomAttributes` and `TestLogCustomSource` we did not set `DT_NO_REPAIR_INPUT: "false"`, but I imagine this does not matter too much.
> 
> For `TestPlatformBuckets` we now miss the following env vars that previously were set:
> 
> ```
> DYNATRACE_LOG_HTTP: "terraform-provider-dynatrace.log"
> DYNATRACE_HTTP_RESPONSE: true 
> ```

I would assume this doesn't matter. As the tests succeed, I don't see a big problem
For the log stuff, I would expect that this was temporary in there to debug this test. If needed we can simply set the env inside the test

Kirdock:
> You added `-p 1` for all integration tests now whereas previously it was only used for the OpenPipelineV2 tests. This leads to a significant increase in test duration, right? The integration test stage on this PR is taking 51 minutes, whereas for other PRs with the old test setup I see lower numbers, e.g. 38 minutes here: #878
> 
> Would it make sense to go with two jobs and two sorts of tags? `integration-test` and `integration-test-nonparallel` or something like that? Or would we, for the sake of simplicity, rather take the increase in test duration?

I've tried it with several approaches now, but the matrix one seems the most stable one.
I tried
- removing all custom implementation inside the core-client
- not using the oauth2 lib (by simply using a platform token, which doesn't rely on the oauth2 lib currently)
- by adapting some options for the http-client

Still, even with everything tried out, the tests run into `unexpected EOF` errors if they run in parallel via `go test -tags=integration -v ./...`

tomazpu (approved):

PR #881, opened 2025-11-18
feat(extensions): Adds scope parameter

d0weinberger:
#### **Why** this PR?
Some monitoring configurations for extensions expect specific scopes to be set. Before this PR, users were not able to set the scope themselves. We tried to deduce it from the other attributes of the `dynatrace_hub_extension_config` resource, but this approach is not sufficient for all potential scopes. To enable Terraform management for additional extensions, this resource is extended.

#### **What** has changed?
This enables to set the `scope` parameter manually in HCL for `dynatrace_hub_extension_config` resources.
If it is set, it is used for the value of `scope`. If it is not set, the previous approach, where we try to deduce the scope from the rest of the attributes of the resource, is taken.

#### **How** does it do it?
The `scope` parameter is added to the HCL schema and it is used in the CRUD service

#### How is it **tested**?
An additional testfile is added where the scope parameter is set.

Users can set the `scope` parameter for `dynatrace_hub_extension_config` resources which require this.

**Issue:**
CA-17125

### DIFF
--- a/dynatrace/api/v2/hub/extension/config/testdata/terraform/withscope.tf
+++ b/dynatrace/api/v2/hub/extension/config/testdata/terraform/withscope.tf
@@ -1,5 +1,6 @@
 resource "dynatrace_hub_extension_config" "com_dynatrace_extension_jmx-weblogic-cp" {
   name = "com.dynatrace.extension.jmx-weblogic-cp"
+  scope = "environment"
     value = jsonencode(
     {
       "activationContext": "LOCAL",
--- a/dynatrace/api/v2/hub/extension/config/service.go
+++ b/dynatrace/api/v2/hub/extension/config/service.go
@@ -93,6 +93,7 @@ func (me *service) Get(ctx context.Context, id string, v *extension_config.Setti
 
 	injectScope(response.Scope, v)
 	v.Name = name
+	v.Scope = response.Scope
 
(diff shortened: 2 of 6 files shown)


dynatrace/api/v2/hub/extension/config/testdata/terraform/withscope.tf
Kirdock, line 1: With the file name change, does the `tmpl` file also need to be updated?

dynatrace/api/v2/hub/extension/config/testdata/terraform/withscope.tf
```diff
@@ -1,5 +1,6 @@
-resource "dynatrace_hub_extension_config" "com_dynatrace_extension_jmx-weblogic-cp" {
+resource "dynatrace_hub_extension_config" "com_dynatrace_extension_jmx-weblogic-cp2-scope" {
```
Kirdock: The different resource names (this one and the `noscope.tf` one) are most probably the cause of the failing test

dynatrace/api/v2/hub/extension/config/testdata/terraform/withscope.tf
d0weinberger: Good point!

Kirdock (approved):

PR #903, opened 2025-11-25
feat: Add `dynatrace_azure_connection` and `dynatrace_azure_connection_authentication` resources for managing Azure connections

arthurpitman:
#### **Why** this PR?
This PR adds two new resources, `dynatrace_azure_connection` and `dynatrace_azure_connection_authentication`, for managing Azure connections, both with client secrets and federated identity credentials.

Using a client secret is very straightforward: 
```
resource "dynatrace_azure_connection" "example" {
  name = "Example Azure connection"
  type = "clientSecret"
  client_secret {
    client_secret  = azuread_application_password.example.value
    application_id = azuread_application_registration.example.client_id
    directory_id   = var.azure_tenant_id
    consumers = [ ... ]
  }
}
```

Using federated identity credentials is a three step process of creating the connection, then the Azure resource to accept that connection and finally updating the connection such that the validation succeeds:
```
resource "dynatrace_azure_connection" "example" {
  name = "Example Azure connection"
  type = "federatedIdentityCredential"
  federated_identity_credential {
    consumers = [ ... ]
  }
}

resource "azuread_application_federated_identity_credential" "example" {
  application_id = azuread_application_registration.example.id
  display_name   = "ExampleFederatedCredential"
  audiences      = ["${var.dynatrace_environment_url}/app-id/dynatrace.microsoft.azure.connector"]
  issuer         = var.dynatrace_token_issuer
  subject        = "dt:connection-id/${dynatrace_azure_connection.example.id}"
}

resource "dynatrace_azure_connection_authentication" "example" {
  azure_connection_id = dynatrace_azure_connection.example.id
  application_id      = azuread_application_registration.example.client_id
  directory_id        = var.azure_tenant_id
}
```

#### **What** has changed?
- Firstly, the retry logic is moved from the AWS connection and adapted slightly to be reused.
- The following 4 commits are the result of running the generator for the settings schema and tweaking the output.
- Finally, manual adaptations are made to create the `dynatrace_azure_connection_authentication` resource.

#### **How** does it do it?
The use of `dynatrace_azure_connection` with client secrets is straightforward and basically is fully generated, perhaps with only minor tweaks for when the resource must be forcefully recreated.

For federated client credentials, the `dynatrace_azure_connection` does not include the application and directory ID used as these trigger validation. Instead these are exposed by the `dynatrace_azure_connection_authentication` resource that then updates the underlying connection.

Creation and updates to for `dynatrace_azure_connection` perform validation which, based on testing, may not succeed first time as changes on the Azure side might require 10 seconds or so to go though. As a result, these operations are retried for a period determined by `retry.DurationUntilDeadlineOrDefault(...)` if `retry.ClassifyRetryError(...)` indicates they should be. 

One further complexity is that the application and directory IDs used for federated identity credentials cannot be changed once set. This would normally lead to slow errors, as these requests would be retried for the entire retry period before eventually failing. This is because both connection not ready style errors and changes to these fields trigger constraint violation errors (HTTP 400).

Using `ForceNew` here is not enough as it still results in an attempted change to the underlying connection object. Instead we look for a change using the `CustomizeDiff(...)` method on the authentication settings object which detects changes to application and directory ID during plan. The services `Create(...)` method also checks this to catch the case where the target connection is changed but the not the application or directory ID.

#### How is it **tested**?
Currently, extensive manual testing. Two example/test files are included in the PR; however, running these as automated acceptance tests would require a valid Azure subscription.

Unit tests have been added for the Create method of the dynatrace_azure_connection_authentication resource service.

Users will be able to manage Azure connections by creating valid `builtin:hyperscaler-authentication.connections.azure` Settings objects, both for client secrets and federated identity credentials.

**Other notes to reviewers:**
- Before merging, I'd squash the commits somewhat; currently, there are separate commits for the generated content, making it easier to understand what has really been written for these resources.
- Should we consider renaming the `dynatrace_azure_connection_authentication` resource? It can only be used with federated identity credentials, so maybe the name should reflect this.
- In cases where just the connection must be recreated, using a targeted destroy is a good option: `terraform destroy -target dynatrace_azure_connection.example`. Running `terraform apply` after will recreate things.

**Issue:** CA-16696


d0weinberger:
So, for the client_secret case we can already set `application_id` and `directory_id` when creating `dynatrace_azure_connection` but for the federated identity case we set it later with `dynatrace_azure_connection_authentication`? Does the API require it this way?

I find it a bit confusing that you can set `application_id` and `directory_id` in both resources. Is there a way to set them only in `dynatrace_azure_connection` or `dynatrace_azure_connection_authentication`? Potentially, you only need the retry logic in `dynatrace_azure_connection_authentication` then.

Of course, this makes the client_secret case less convenient, because you suddenly need two resources for it when one would have been enough. This is a downside

EDIT: Okay, it seems like we have no other choice here - this is how the API works, unfortunately.

dynatrace/api/builtin/hyperscalerauthentication/connections/azure/settings/settings.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+type Settings struct {
+	ClientSecret                *ClientSecretConfig          `json:"clientSecret,omitempty"`
+	FederatedIdentityCredential *FederatedIdentityCredential `json:"federatedIdentityCredential,omitempty"`
+	Name                        string                       `json:"name"` // The name of the connection
+	Type                        Type                         `json:"type"` // Azure Authentication mechanism to be used by the connection. Possible Values: `clientSecret`, `federatedIdentityCredential`
+}
+
+const DefaultTimeout = 2 * time.Minute
```
d0weinberger, line 35: Did you double-check the chosen default timeout?

dynatrace/api/builtin/hyperscalerauthentication/connections/azure/service.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api"
+	retrycommon "github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/builtin/hyperscalerauthentication/connections/retry"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/rest"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/settings"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/settings/services/settings20"
+)
+
+const SchemaVersion = "0.0.10"
+const SchemaID = "builtin:hyperscaler-authentication.connections.azure"
+const DefaultTimeout = 2 * time.Minute
```
d0weinberger, line 36: Same question about the default timeout here

dynatrace/api/builtin/hyperscalerauthentication/connections/azure/settings/settings.go
arthurpitman: Honestly for creation I have never had it need more than 10 seconds, the exact value is "to be determined".

dynatrace/api/builtin/hyperscalerauthentication/connections/azure/service.go
arthurpitman: Both timeouts need to be the same, as they will be applied to the context that eventually is used in the retry logic.

dynatrace/api/builtin/hyperscalerauthentication/connections/azure/authentication/settings/settings.go
```diff
@@ -0,0 +1,114 @@
...
+)
+
+type Settings struct {
+	Name              string
+	AzureConnectionID string
+	ApplicationID     *string // Application (client) ID of your app registered in Microsoft Azure App registrations
+	DirectoryID       *string // Directory (tenant) ID of Microsoft Entra ID
+}
+
+const DefaultTimeout = 2 * time.Minute
```
d0weinberger: Same here. Actually, since it's always the same number, I think it would make sense to have this as a common constant somewhere and reference to it.

dynatrace/api/builtin/hyperscalerauthentication/connections/azure/authentication/settings/settings.go
```diff
@@ -0,0 +1,114 @@
...
+func (me *Settings) UnmarshalHCL(decoder hcl.Decoder) error {
+	return decoder.DecodeAll(map[string]any{
+		"azure_connection_id": &me.AzureConnectionID,
+		"application_id":      &me.ApplicationID,
+		"directory_id":        &me.DirectoryID,
+	})
+}
+
+func (me *Settings) CustomizeDiff(ctx context.Context, rd *schema.ResourceDiff, i any) error {
+	// try to check if an attempt is being made to change application and directory IDs for a connection where they were already set.
```
d0weinberger, line 85: Ah, interesting! I think that's missing in the AWS case, right? 🤔  we might need to add it there as well

arthurpitman:
> So, for the client_secret case we can already set `application_id` and `directory_id` when creating `dynatrace_azure_connection` but for the federated identity case we set it later with `dynatrace_azure_connection_authentication`? Does the API require it this way?
> 
> I find it a bit confusing that you can set `application_id` and `directory_id` in both resources. Is there a way to set them only in `dynatrace_azure_connection` or `dynatrace_azure_connection_authentication`? Potentially, you only need the retry logic in `dynatrace_azure_connection_authentication` then.
> 
> Of course, this makes the client_secret case less convenient, because you suddenly need two resources for it when one would have been enough. This is a downside
> 
> EDIT: Okay, it seems like we have no other choice here - this is how the API works, unfortunately.

Basically the reason we need two resources or retries is because the API is performing validation.

Federated identity credentials require two because we need the API to "try" the connection *after* the federated identity credentials have been set up on Azure. And we need the retry here because it may take a while for it to really become ready.

Using a client secret only requires one resource, but does need a retry because again the Dynatrace API may be too quick and try the secret before it is truly ready on the Azure end.

dynatrace/api/builtin/hyperscalerauthentication/connections/azure/authentication/settings/settings.go
arthurpitman: @Kirdock helped me here :) It's pretty cool that we can catch this during the plan, a win for usability!

d0weinberger (dismissed):
The only comments I still had were some nit-picks about the timeout value. I guess there might be a follow-up depending on what value we decide on, but 2 minutes seem fine for now.
I share your sentiment on the resource naming `dynatrace_azure_connection_authentication`, but I also don't have a better idea right now.

dynatrace/api/builtin/hyperscalerauthentication/connections/azure/authentication/service.go
```diff
@@ -0,0 +1,96 @@
...
+	if err := me.connService.Get(ctx, v.AzureConnectionID, &connValue); err != nil {
+		return nil, err
+	}
+
+	if (connValue.Type != azureconnection_settings.Types.Federatedidentitycredential) || (connValue.FederatedIdentityCredential == nil) {
+		return nil, errors.New("the associated Azure Connection is not configured for Federated Identity Credential authentication")
+	}
+
+	// it is not possible to change directory or application ID after they have been set
+	if connValue.FederatedIdentityCredential.DirectoryID != nil || connValue.FederatedIdentityCredential.ApplicationID != nil {
```
d0weinberger, line 81: What you _could_ do is also check for equality. If `DirectoryID` and `ApplicationID` are not nil but equal to the values that are already set, it`s not really an update.

Kirdock: Then I would say that the user should use an import block instead

Kirdock (dismissed):

d0weinberger (dismissed):

d0weinberger (approved):

PR #909, opened 2025-11-26
feat(aws-connection): Handle role ARN updates

d0weinberger:
#### **Why** this PR?
The state before this PR: If a `dynatrace_aws_connection` already has a role ARN set, but you create a `dynatrace_aws_connection_role_arn` resource regardless, referencing the AWS connection and ultimately trying to update the role ARN, this will fail with validation errors, as the connection, once all fields are set, becomes immutable.

Since the update fails, we retry unnecessarily for (by default) 2 minutes, before failing with an error message saying that the role ARN cannot be updated.

#### **What** has changed?
The purpose of this change is to fail fast if someone tries to update the role ARN for a connection that already has the role ARN set. If the role ARN of the connection is already set and not equal to the value we try to set, we return an error. Otherwise, it's fine.

#### **How** does it do it?
In `Create` of the CRUD service for the `dynatrace_aws_connection_role_arn` resource `X` we fetch the accompanying AWS connection and check its fields in comparison with the role ARN value we get from `X`. 
If the role ARN of the connection is already set and not equal to the value we try to set, we return
```
AWS Role ARN is already set for the specified AWS Connection
```

#### How is it **tested**?
Unit tests have been added for `Create`. 

Trying to update an already immutable connection fails earlier and with a more specific error message.

**Issue:**
CA-17192

### DIFF
--- a/dynatrace/api/builtin/hyperscalerauthentication/connections/aws/role_arn/service.go
+++ b/dynatrace/api/builtin/hyperscalerauthentication/connections/aws/role_arn/service.go
@@ -75,9 +75,25 @@ func (me *service) Create(ctx context.Context, v *role_arn.Settings) (*api.Stub,
 	if err := me.connService.Get(ctx, v.AWSConnectionID, &connValue); err != nil {
 		return nil, err
 	}
-	if connValue.AWSRoleBasedAuthentication != nil {
+
+	if connValue.Type == awsconnection_settings.Types.AWSRoleBasedAuthentication && connValue.AWSRoleBasedAuthentication != nil {
+		// no update needed if there are no changes
+		if connValue.AWSRoleBasedAuthentication.RoleARN == v.RoleARN {
+			return &api.Stub{ID: v.AWSConnectionID, Name: v.AWSConnectionID}, nil
+		}
+		// it is not possible to change the role ARN after it has been set
+		if connValue.AWSRoleBasedAuthentication.RoleARN != "" {
+			return nil, errors.New("AWS Role ARN is already set for the specified AWS Connection")
+		}
 		connValue.AWSRoleBasedAuthentication.RoleARN = v.RoleARN
-	} else if connValue.AWSWebIdentity != nil {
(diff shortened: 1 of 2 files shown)


dynatrace/api/builtin/hyperscalerauthentication/connections/aws/role_arn/service.go
```diff
@@ -75,9 +75,15 @@ func (me *service) Create(ctx context.Context, v *role_arn.Settings) (*api.Stub,
...
+			return nil, errors.New("AWS Role ARN is already set for the specified AWS Connection")
+		}
 		connValue.AWSRoleBasedAuthentication.RoleARN = v.RoleARN
-	} else if connValue.AWSWebIdentity != nil {
+	} else if connValue.Type == awsconnection_settings.Types.AWSWebIdentity && connValue.AWSWebIdentity != nil {
+		if connValue.AWSWebIdentity.RoleARN != "" && connValue.AWSWebIdentity.RoleARN != v.RoleARN {
+			return nil, errors.New("AWS Role ARN is already set for the specified AWS Connection")
+		}
 		connValue.AWSWebIdentity.RoleARN = v.RoleARN
 	}
```
Kirdock, line 98: Am I right, that this will do a request if there already is a set value and it equals the set value in the HCL?

d0weinberger: Basically, yes

arthurpitman (dismissed):

dynatrace/api/builtin/hyperscalerauthentication/connections/aws/role_arn/service.go
d0weinberger: I changed that now. If there is no change, we don't attempt an update.

Kirdock: With this, we basically manually added an import block, right?
I'm just not sure, if we should have such logic inside or not.

Kirdock (approved):

PR #916, opened 2025-12-05
doc(azure-connection): Add resource documentation

arthurpitman:
#### **Why** this PR?
The PR adds the documentation templates for the `dynatrace_azure_connection` and `dynatrace_azure_connection_authentication` resources.

Documentation templates that include references to the updated example Terraform files.

#### How is it **tested**?
N/A

Documentation is now available for these resources

**Issue:** CA-17189

### DIFF
--- a/dynatrace/api/builtin/hyperscalerauthentication/connections/azure/testdata/azure_connector_client_secret.tf
+++ b/dynatrace/api/builtin/hyperscalerauthentication/connections/azure/testdata/azure_connector_client_secret.tf
@@ -29,7 +29,7 @@ resource "azuread_application_password" "example" {
 
 # Create Azure connection
 resource "dynatrace_azure_connection" "example" {
-  name = "Example Azure Connector - Secret"
+  name = "#name#"
   type = "clientSecret"
   client_secret {
     client_secret  = azuread_application_password.example.value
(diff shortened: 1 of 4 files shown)


arthurpitman:
Two points to clarify:
- How should `dynatrace_azure_connection_authentication` resources be exported?
- Is there a link we can add to Dynatrace documentation?

Kirdock:
> How should `dynatrace_azure_connection_authentication` resources be exported?

At the moment, it should be working, right? For the dependencies, we can only resolve the  `azure_connection_id`, as this is the only other resource managed by us.

For the `dynatrace_azure_connection` resource, this one will be in the needs-work folder because of secrets, right?

Kirdock (approved):

d0weinberger (approved):

arthurpitman:
> > How should `dynatrace_azure_connection_authentication` resources be exported?
> 
> At the moment, it should be working, right? For the dependencies, we can only resolve the `azure_connection_id`, as this is the only other resource managed by us.
> 
> For the `dynatrace_azure_connection` resource, this one will be in the needs-work folder because of secrets, right?

- I've created a follow-up PR to restrict `dynatrace_azure_connection_authentication` to federated identity credentials (#918)
- About `.requires_attention`, this is something for an additional follow-up.

PR #924, opened 2025-12-09
feat(export): add management-zone scoped export

mahdibouaziz:
#### **Why** this PR?
We need to export only configs tied to a specific management zone for backup/restore and scoped moves; exporting the whole tenant is noisy and manual pruning is error-prone.

#### **What** has changed?
- Added `-management-zone "<name>"` flag to resolve a zone by name (including legacy ID) and store it on the export environment.
- Filtered discovery/download to that zone and resources referencing it (by name/ID/legacy ID/scope), excluding out-of-scope items.

#### **How** does it do it?
- Resolves the management zone via settings v2 service; fails fast if not found.
- Applies zone-aware filtering during discovery and post-GET validation; marks non-matching resources as excluded.
- Keeps compatibility with existing flags (`-ref`,  `-exclude`, `-skip-terraform-init`, parallel control).

#### How is it **tested**?
- Manual runs with `-management-zone "<name>"`, with/without `-ref`, and with targeted resource lists; verified only zone-scoped resources are written.

- Optional enhancement: default behavior unchanged. When used, outputs are limited to the specified management zone, reducing noise and manual cleanup.

**Issue:** #923

### DIFF
--- a/dynatrace/export/environment.go
+++ b/dynatrace/export/environment.go
@@ -38,6 +38,7 @@ import (
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/v2/entity"
 	entitysettings "github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/v2/entity/settings"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/rest"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/settings"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/settings/services/cache"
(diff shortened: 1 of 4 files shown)


mahdibouaziz:
Hey @Kirdock @tomazpu @arthurpitman,
Kindly check this when you get a chance. Thanks!

Laubi:
Hi @mahdibouaziz,
 
Thank you for your contribution.

We have discussed your change internally and have decided not to merge it. While we understand the use case you have in mind and the benefits it brings in that scenario, we believe that merging this change would require too much ongoing support. As you're currently the only one requesting it, we don't feel comfortable taking this over at this time.

Since this is an extension to the download functionality of the provider, we recommend maintaining this implementation in a custom fork. 
I also encourage you to suggest this feature in the [Dynatrace Community](https://community.dynatrace.com/). 

Thank you for your understanding.

PR #935, opened 2025-12-15
refactor: support platform tokens for all settings based resources

Kirdock:
#### **Why** this PR?
There are several resources (`dynatrace_davis_anomaly_detectors`, `dynatrace_site_reliability_guardian`, `dynatrace_automation_workflow_slack`, and `dynatrace_generic_setting`) that don't support platform tokens.
Platform tokens should be supported for each of the mentioned resources.

#### **What** has changed?
The mentioned resources now support platform tokens

#### **How** does it do it?
By making use of either the generic settings service or the hybrid client.
The old special settings client has been removed now.
The fallback for token => OAuth is still implemented for the anomaly detector and the generic settings, to prevent breaking existing configurations. The logic is slightly different, as the error parsing is now already done by the generic settings service, so we just need to `error.As` the returned error to get the constraint violations.

#### How is it **tested**?
- Existing tests still succeed
- New test for the anomaly detector added (fallback from token => OAuth)
- Local export of generic settings worked

They're now able to use platform tokens for the mentioned resources.

**Issue:** CA-17297

### Additional notes:
**Why not adapt the existing client instead of refactoring and removing it?**
This would add additional complexity and also redundancy. It would not have been easy to modify the existing client because this one didn't care about platform URLs. Every CRUD operation would have needed to be modified, and additional logic would have been required to differentiate between an OAuth client and a classic client (if OAuth, then replace `/api/...` with `platform/classic-environment/...`). In the end, it would have copied more from the hybrid client, and therefore, I used the hybrid client instead to remove complexity


dynatrace/api/builtin/generic/service.go
```diff
@@ -41,28 +44,26 @@ type service struct {
...
-	oauthClient := rest.CreateClassicOAuthBasedClient(ctx, me.credentials)
-	return settings20.NewClient(tokenClient, oauthClient, schemaIDs)
+func (me *service) Client() rest.Client {
+	return rest.HybridClient(me.credentials)
+}
+
+func (me *service) getOAuthClient() rest.Client {
+	cred := *me.credentials
+	cred.Token = ""
+	return rest.HybridClient(&cred)
```
d0weinberger: Forcing the hybrid client to use OAuth by removing the token from the credentials feels a bit hacky, IMHO.
Using `rest.CreateClassicOAuthBasedClient(ctx, me.credentials)` would be clearer

dynatrace/api/builtin/generic/service.go
```diff
@@ -126,70 +160,73 @@ func (me *service) Validate(v *generic.Settings) error {
...
+		SchemaID: v.SchemaID,
+		Scope:    scope,
+		Value:    json.RawMessage(v.Value),
+	}}
+
+	var response []settings20.SettingsObjectCreateResponse
+	err := client.Post(ctx, settingsObjectEndpoint, obj).Finish(&response)
 	if err != nil {
+		if rest.IsRequiresOAuthError(err) && me.credentials.ContainsOAuthOrPlatformToken() {
+			return me.create(ctx, v, me.getOAuthClient())
```
d0weinberger: I know it _shouldn't_ cause a problem, but is there any risk of this leading to an infinite recursion?

dynatrace/api/builtin/generic/service.go
```diff
@@ -126,70 +160,73 @@ func (me *service) Validate(v *generic.Settings) error {
...
-		if response.StatusCode == 0 {
-			return errors.New(errMsgOAuthRequired)
+	u, err := url.JoinPath(settingsObjectEndpoint, id)
+	if err != nil {
+		return newSettingsURLJoinError(err)
+	}
+	err = client.Put(ctx, u, obj).Finish(nil)
+	if err != nil {
+		if rest.IsRequiresOAuthError(err) && me.credentials.ContainsOAuthOrPlatformToken() {
+			return me.update(ctx, id, v, me.getOAuthClient())
```
d0weinberger: Same here. I'm wondering whether an incorrectly configured OAuth client due to some error in the credentials could lead to `IsRequiresOAuthError` return true, leading to an infinite recursion

dynatrace/api/builtin/davis/anomalydetectors/service.go
```diff
@@ -33,132 +31,64 @@ const SchemaVersion = "1.0.10"
...
 }
 
-func (me *service) Client(ctx context.Context, schemaIDs string) *settings20.Client {
-	tokenClient, _ := rest.CreateClassicClient(me.credentials.URL, me.credentials.Token)
-	oauthClient := rest.CreateClassicOAuthBasedClient(ctx, me.credentials)
-	return settings20.NewClient(tokenClient, oauthClient, schemaIDs)
+func (me *service) getOAuthService() settings.CRUDService[*anomalydetectors.Settings] {
+	cred := *me.credentials
+	cred.Token = ""
+	return settings20.Service[*anomalydetectors.Settings](&cred, SchemaID, SchemaVersion)
```
d0weinberger: What do you think about creating a `settings20.OAuthService()` func to specifically request an OAuth service instead of doing it indirectly via deleting the token?

dynatrace/settings/services/settings20/service.go
```diff
@@ -359,14 +359,7 @@ type Matcher interface {
...
 func (me *service[T]) skipRepairInput() bool {
-	if IsSkipRepairSchemaID(me.schemaID) {
-		return true
-	}
-	if NO_REPAIR_INPUT {
-		return true
-	}
-
-	return false
+	return NO_REPAIR_INPUT
```
d0weinberger, line 362: So, we don't have any schema IDs anymore for which we don't want to skip the `repairInput`?

dynatrace/api/builtin/generic/service.go
Kirdock: The `CreateClassicOAuthBasedClient` doesn't exist anymore and was lacking features. It was also only targeting classic and not platform, so it was missing the platform and platform token support.

dynatrace/api/builtin/generic/service.go
Kirdock: This would lead to a different error message though, as the OAuth request would fail and the response would then be from the SSO and not settings

dynatrace/settings/services/settings20/service.go
Kirdock: This was a remaining thing from the special client (`dynatrace/settings/services/settings20/client.go`) that always returned false. As the client is removed, it's function can also be removed

dynatrace/api/builtin/generic/service.go
Kirdock: Shouldn't be possible, as the error is explicitly evaluated against the settings response. If OAuth is invalid, it wouldn't even hit the settings endpoint and fail during the SSO. As we explicitly check for the `OAuth required` error message, missing permission lead to a different one.

dynatrace/api/builtin/davis/anomalydetectors/service.go
Kirdock: I would fear that this would lead to a lot of code duplication and will be very similar to the hybrid client.
The reason for not using the platformClient is that the hybrid client contains all the magic to support classic AND OAuth. If we are using the platform client we would have to adjust every CRUD call to target the right environment. From `api/v2/...` to `platform/classic-environment/...`.  This would end in a hybrid_client that is just slightly different but most of the functionality would remain the same

d0weinberger (dismissed):

dynatrace/api/builtin/generic/service.go
Kirdock: Even though, I'll adjust it a bit

dynatrace/api/builtin/generic/service.go
Kirdock: Even though, I'll adjust it a bit

dynatrace/api/builtin/davis/anomalydetectors/service.go
Kirdock: I found another solution. Will use that. Will commit it soon

Kirdock: Solution committed. I made use of the context key that is already used in the hybrid_client. Therfore, it's not needed to adjust the credentials and I can simply try it again with a different context.

d0weinberger (approved):

arthurpitman (commented):
The PR looks good. My only concern is how equivalent the underlying clients really are. The change of client may not be significant, it may also be an improvement. For example, I believe the retry settings have changed for these resources.

Kirdock:
> The PR looks good. My only concern is how equivalent the underlying clients really are. The change of client may not be significant, it may also be an improvement. For example, I believe the retry settings have changed for these resources.

I would still say it's an improvement. Before, there were actually two retry implementations. The client had its own implementation and because it was using the core client, there was additional retry logic

arthurpitman (approved):

PR #944, opened 2025-12-18
feat: environments data source `dynatrace_iam_environments`

### LINKED ISSUE #869: Add a new data source provider to get a list of all dynatrace environments
When using the [Dynatrace Account Management API](https://api.dynatrace.com/spec#/Environment%20management) you can use the [GET
/env/v1/accounts/{accountUuid}/environments](https://api.dynatrace.com/spec#/Environment%20management/EnvironmentResourcesController_getEnvironmentResources) endpoint ot list all environments and management zones of an account. 

Example response

```json
{
  "tenantResources": [
    {
      "name": "string",
      "id": "string"
    }
  ],
  "managementZoneResources": [
    {
      "parent": "string",
      "name": "string",
      "id": "string"
    }
  ]
}
```

The response contains a list of the account's environments `tenantResources` and  a list of the management zones `managementZoneResources`. It would be great to add a new data source `dynatrace_tenants` (similar to the existing [`dynatrace_tenant`](https://registry.terraform.io/providers/dynatrace-oss/dynatrace/latest/docs/data-sources/tenant)) to get a list of all available tenant IDs `tenantResources` as the output.

This would be useful for permission handling in multi-environment accounts so we do not need to hardcode the environment IDs in our Terraform configurations.

Kirdock:
#### **Why** this PR?
To support the case where someone needs all available environments of an account.

#### **What** has changed?
A new data source `dynatrace_iam_environments` is added.

#### **How** does it do it?
By querying `/env/v2/accounts/{account_id}/environments` of an account and filtering based on the set properties of the data source.

#### How is it **tested**?
Manual tests.

They can now query their available environments.

**Issue:** CA-17072
Fixes #869

### Additional notes
Because `id` is reserved by the plugin SDK, I used `uuid` to have a separation.

A better approach would have been to have a parent `filter` object that contains the fields that can be filtered on. I decided against that because we aren't using this structure for other data sources.

### DIFF
--- a/datasources/iam/environments/data_source.go
+++ b/datasources/iam/environments/data_source.go
@@ -1,0 +1,143 @@
+/**
+* @license
+* Copyright 2020 Dynatrace LLC
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+ */
+
+package environments
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider/config"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider/logging"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// Attribute keys
+const (
+	// this one is different from attrID inside environments list because "id" is a reserved keyword on root level.
+	attrEnvID        = "env_id"
+	attrID           = "id"
+	attrName         = "name"
+	attrActive       = "active"
+	attrEnvironments = "environments"
+	attrURL          = "url"
+)
+
+func DataSource() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: logging.EnableDSCtx(DataSourceRead),
+		Description: "Returns the environments matching the provided filter criteria",
+		Schema: map[string]*schema.Schema{
+			attrEnvID: {
+				Type:        schema.TypeString,
+				Description: "Filter by ID of the environment",
+				Optional:    true,
+			},
+			attrName: {
+				Type:        schema.TypeString,
+				Description: "Filter by friendly name of the environment",
+				Optional:    true,
+			},
+			attrActive: {
+				Type:        schema.TypeBool,
+				Description: "Filter by active status of the environment",
+				Optional:    true,
+			},
+			attrEnvironments: {
+				Type:        schema.TypeList,
+				Description: "List of environments matching the filter criteria",
+				Computed:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						attrID: {
+							Type:        schema.TypeString,
+							Description: "The ID of the environment",
+							Computed:    true,
+						},
+						attrName: {
+							Type:        schema.TypeString,
+							Description: "Friendly name of the environment",
+							Computed:    true,
+						},
+						attrActive: {
+							Type:        schema.TypeBool,
+							Description: "Property to determine if environment is active",
+							Computed:    true,
+						},
+						attrURL: {
+							Type:        schema.TypeString,
+							Description: "The URL of the environment",
+							Computed:    true,
+						},
+					},
+				},
+			},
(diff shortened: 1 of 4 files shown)


datasources/iam/environments/data_source.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+const baseURL = "/env/v2/accounts"
+
+func DataSource() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: logging.EnableDSCtx(DataSourceRead),
+		Description: "Returns the environments matching the provided filter criteria",
+		Schema: map[string]*schema.Schema{
+			"uuid": {
```
d0weinberger: I want to really veto `uuid` here, because it just isn't a uuid.
What about `environment_id`?

datasources/iam/environments/data_source.go
```diff
@@ -0,0 +1,134 @@
...
+	"context"
+	"fmt"
+
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider/config"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider/logging"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+const baseURL = "/env/v2/accounts"
```
d0weinberger: [nit] This one is only used in `service.go`, so I'd move it there.

datasources/iam/environments/data_source.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	service := newEnvironmentService(creds)
+	envs, err := service.Get(ctx)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	// Get filter values
+	filterID, hasFilterID := d.GetOk("uuid")
```
d0weinberger: I know we don't do that anywhere else, but maybe we should: Wdyt about having string constants for "uuid", "name", and so on an reuse them instead of always using the hardcoded strings?

datasources/iam/environments/data_source.go
```diff
@@ -0,0 +1,134 @@
...
+				Description: "Filter by active status of the environment",
+				Optional:    true,
+			},
+			"environments": {
+				Type:        schema.TypeList,
+				Description: "List of environments matching the filter criteria",
+				Computed:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"id": {
```
d0weinberger: wait, is `id` okay here?

datasources/iam/environments/data_source.go
Kirdock: Missed this one, thx

datasources/iam/environments/data_source.go
Kirdock: Decided with @arthurpitman to go with `env_id` on root level and `id` inside of environments. 👍 
The reason for uuid was simply because IAM describes it in their open spec as UUID

datasources/iam/environments/data_source.go
Kirdock: The `id` is only reserved on root level, so we decided to go with just `id` because we are already in an `environments` array

datasources/iam/environments/data_source.go
Kirdock: Hmmm... @arthurpitman, your opinion on that?

Kirdock: Pushed it for now. If not wanted, I can revert it 👍

templates/data-sources/iam_environments.md.tmpl
```diff
@@ -0,0 +1,41 @@
...
+## Example Usage
+
+### Query active environments
+```terraform
+data "dynatrace_iam_environments" "active_environments" {
+  active = true
+}
+
+output "active_environments" {
+  value = data.dynatrace_iam_environments.active_environments.environments[*].uuid
```
d0weinberger: Should also be `id` here then instead of `uuid`

templates/data-sources/iam_environments.md.tmpl
```diff
@@ -0,0 +1,41 @@
...
+```
+
+### Query environment by name
+```terraform
+data "dynatrace_iam_environments" "by_name" {
+  name = "production"
+}
+
+output "environment_uuid" {
+  value = data.dynatrace_iam_environments.by_name.environments[0].uuid
```
d0weinberger: same here

templates/data-sources/iam_environments.md.tmpl
Kirdock: Ah, forgot that these are not generated.

Kirdock: Updated 👍

templates/data-sources/iam_environments.md.tmpl
Kirdock: Updated 👍

d0weinberger (approved):

arthurpitman (approved):

PR #947, opened 2025-12-24
feat: Add `dynatrace_iam_service_user` resource

arthurpitman:
#### **Why** this PR?
This PR adds support for service users by introducing the `dynatrace_iam_service_user` resource and data source.

The `dynatrace_iam_service_user` resource works similarly to `dynatrace_iam_user` resource, allowing the `name`, `description` and `groups` of the service user to be specified.

As groups are assigned using a second API call, a freshly created service user must be explicitly deleted if this second API call fails. This adds complexity to service, and might be addressed by adding a dedicated resource to assign groups in the future.

The `dynatrace_iam_service_user` data source allows a service user can be looked up directly by `id` or by the first matching `name` or `email`. Currently, only a single service user is returned.

#### How is it **tested**?
Currently, IAM resources or data sources in the provider are not covered by E2E tests.
This PR includes tests for the CRUD operations of the service user service as well as the data source read operation.

Users will be able to manage service users using the provider.

**Issue:** CA-16921

### DIFF
--- a/dynatrace/api/iam/serviceusers/service.go
+++ b/dynatrace/api/iam/serviceusers/service.go
@@ -1,0 +1,254 @@
+/**
+* @license
+* Copyright 2025 Dynatrace LLC
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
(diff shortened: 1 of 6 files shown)


arthurpitman:
Closed in favor of #950

PR #948, opened 2025-12-30
fix(request-naming): Correctly decode `unknowns` from HCL for `BaseComparisonInfo`

d0weinberger:
#### **Why** this PR?
Currently, users can not use the `generic` comparison type, as `unknowns` cannot be used.

#### **What** has changed?
The way `unknowns` was decoded was wrong and always yielded an empty result. This PR fixes this.

#### **How** does it do it?
If `unknowns` is present in the HCL, we unmarshal it via `json.Unmarshal`

#### How is it **tested**?
No test. Manual check that for `dynatrace_request_naming` with `unknowns` set in `comparison`, the correct payload is sent.

It fixes a bug and enables users to make use of `unknowns` in `comparison` in `dynatrace_request_naming`.

**Issue:**
CA-17511


Kirdock:
Would something speak against adding the `unknowns` field to one of the examples?

Laubi (approved):

Kirdock (approved):

PR #950, opened 2026-01-07
feat: Add `dynatrace_iam_service_user` resource

arthurpitman:
#### **Why** this PR?
This PR adds support for service users by introducing the `dynatrace_iam_service_user` resource.

The `dynatrace_iam_service_user` resource works similarly to `dynatrace_iam_user` resource, allowing the `name`, `description` and `groups` of the service user to be specified. Unlike the `dynatrace_user` resource, the actual user ID is used as the `id` of the resource, and `email`, which also includes the ID, is a computed value. For this reason, the resource currently doesn't provide an additional `uid` value.

As groups are assigned using a second API call, a freshly created service user is explicitly deleted if this second API call fails. This does add complexity to the service, and might be addressed by adding a dedicated resource to assign groups in the future.

The implementation is kept as simple as possible; we may want to change this. Compared to the `dynatrace_iam_user` resource, it doesn't [translate errors to `rest.Error`](https://github.com/dynatrace-oss/terraform-provider-dynatrace/blob/b8eaf64efec4d6c944817a367f64f6afcf7ec45a/dynatrace/api/iam/users/service.go#L117), or [ignore service users that have already been deleted](https://github.com/dynatrace-oss/terraform-provider-dynatrace/blob/b8eaf64efec4d6c944817a367f64f6afcf7ec45a/dynatrace/api/iam/users/service.go#L181-L183). 

#### How is it **tested**?
Currently, IAM resources in the provider are not covered by E2E tests.
This PR includes tests for the CRUD operations of the service user service. As the IAM client cannot be cached in the service (it doesn't refresh the bearer token), the `iamClientGetter` interface is used to allow the service to be tested.

Users will be able to manage service users using the provider.

**Issue:** CA-16921

### DIFF
--- a/dynatrace/api/iam/serviceusers/service.go
+++ b/dynatrace/api/iam/serviceusers/service.go
@@ -1,0 +1,249 @@
+/**
+* @license
+* Copyright 2025 Dynatrace LLC
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+ */
+
+package serviceusers
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/iam"
+	serviceusers "github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/iam/serviceusers/settings"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/rest"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/settings"
+)
+
+type iamClientGetter interface {
+	New() iam.IAMClient
+}
+
+type iamClientGetterImp struct {
+	clientID     string
+	accountID    string
+	clientSecret string
+	tokenURL     string
+	endpointURL  string
+}
+
+func (me *iamClientGetterImp) ClientID() string {
+	return me.clientID
+}
+
+func (me *iamClientGetterImp) AccountID() string {
+	return me.accountID
+}
+
+func (me *iamClientGetterImp) ClientSecret() string {
+	return me.clientSecret
+}
+
+func (me *iamClientGetterImp) TokenURL() string {
+	return me.tokenURL
+}
+
+func (me *iamClientGetterImp) EndpointURL() string {
+	return me.endpointURL
+}
+
+func (me *iamClientGetterImp) New() iam.IAMClient {
+	return iam.NewIAMClient(me)
+}
+
+type serviceUserServiceClient struct {
+	iamClientGetter iamClientGetter
+	accountID       string
+	endpointURL     string
+}
+
+func Service(credentials *rest.Credentials) settings.CRUDService[*serviceusers.ServiceUser] {
+	return &serviceUserServiceClient{
+		iamClientGetter: &iamClientGetterImp{
+			clientID:     credentials.IAM.ClientID,
+			accountID:    credentials.IAM.AccountID,
+			clientSecret: credentials.IAM.ClientSecret,
+			tokenURL:     credentials.IAM.TokenURL,
+			endpointURL:  credentials.IAM.EndpointURL,
+		},
+		accountID:   credentials.IAM.AccountID,
+		endpointURL: credentials.IAM.EndpointURL,
+	}
+}
+
+func (me *serviceUserServiceClient) SchemaID() string {
+	return "accounts:iam:serviceusers"
+}
+
+// createResponse represents the response from creating a service user
+type createResponse struct {
+	UID   string `json:"uid"`
+	Email string `json:"email"`
+	Name  string `json:"name"`
+}
+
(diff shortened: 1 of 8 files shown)


dynatrace/api/iam/serviceusers/settings/service_user.go
```diff
@@ -0,0 +1,80 @@
...
+type ServiceUser struct {
+	Email       string   `json:"email,omitempty"`
+	Name        string   `json:"name"`
+	Description string   `json:"description,omitempty"`
+	Groups      []string `json:"-"`
+}
+
+func (me *ServiceUser) GetName() string {
+	return me.Name
+}
```
Kirdock: Is this used somewhere?

dynatrace/api/iam/serviceusers/service.go
```diff
@@ -0,0 +1,249 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var response getUserPartialResponse
+	if err = json.Unmarshal(responseBytes, &response); err != nil {
+		return nil, err
+	}
+
+	groups := []string{}
```
Kirdock: nit
```suggestion
	groups := make([]string, 0, len(response.Groups))
```

dynatrace/api/iam/serviceusers/settings/service_user.go
```diff
@@ -0,0 +1,80 @@
...
+		},
+		"description": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Description: "The description of the service user",
+		},
+		"groups": {
+			Type:        schema.TypeSet,
+			Optional:    true,
+			MinItems:    1,
```
Kirdock: Do we even need `MinItems`? On the one hand, not given is the same as an empty array, and on the other hand, why should we limit it?

Kirdock (dismissed):

dynatrace/api/iam/serviceusers/settings/service_user.go
arthurpitman: No need to limit it, I'll update it

dynatrace/api/iam/serviceusers/settings/service_user.go
arthurpitman: No, it is not needed, as we have a `Name` field here, so I will remove it.

Kirdock (approved):

dynatrace/api/iam/serviceusers/service.go
```diff
@@ -0,0 +1,249 @@
...
+
+// createResponse represents the response from creating a service user
+type createResponse struct {
+	UID   string `json:"uid"`
+	Email string `json:"email"`
+	Name  string `json:"name"`
+}
+
+func (me *serviceUserServiceClient) Create(ctx context.Context, serviceUser *serviceusers.ServiceUser) (*api.Stub, error) {
+	responseBytes, err := me.iamClientGetter.New().POST(ctx, fmt.Sprintf("%s/iam/v1/accounts/%s/service-users", me.endpointURL, me.accountID), serviceUser, 201, false)
```
d0weinberger, line 100: [minor] the URL pattern `%s/iam/v1/accounts/%s/service-users` is repeated several times. You could extract it to a constant.

dynatrace/api/iam/serviceusers/service.go
```diff
@@ -0,0 +1,249 @@
...
+	if err := json.Unmarshal(responseBytes, &response); err != nil {
+		return nil, err
+	}
+
+	if err := me.updateGroupAssignments(ctx, response.Email, serviceUser.Groups); err != nil {
+		deleteErr := me.Delete(ctx, response.UID)
+		if deleteErr != nil {
+			return nil, fmt.Errorf("failed to create service user: %v; additionally failed to clean up service user: %v", err, deleteErr)
+		}
+		return nil, err
```
d0weinberger, line 115:
```suggestion
		return nil, fmt.Errorf("failed to create service user: %v", err)
```

dynatrace/api/iam/serviceusers/service.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+func (me *serviceUserServiceClient) Update(ctx context.Context, id string, serviceUser *serviceusers.ServiceUser) error {
+	// Update the service user details
+	if _, err := me.iamClientGetter.New().PUT(ctx, fmt.Sprintf("%s/iam/v1/accounts/%s/service-users/%s", me.endpointURL, me.accountID, id), serviceUser, 200, false); err != nil {
+		return err
+	}
+
+	// Update group assignments
+	return me.updateGroupAssignments(ctx, serviceUser.Email, serviceUser.Groups)
```
d0weinberger, line 189: If in `Create` there is an error with updating the group assignment, we delete the service user again, so we basically rollback the whole `Create`. Should we roll back the `Update` here as well, i.e. update to the previous state, if `updateGroupAssignments` fails?

Or, would it make sense, maybe, to have the service-user-to-group-assignment handled via another dedicated resource?

dynatrace/api/iam/serviceusers/service.go
arthurpitman: I'd prefer to leave these as they are, as this is the flavor of all these IAM resources, and at least makes it very clear how the URL is formed.

dynatrace/api/iam/serviceusers/service.go
arthurpitman: We roll back the create because otherwise a service user not assigned to any groups will be left over. Re-applying will create another, rather than updating this one, as the ID is not stored in the state.

On the other hand, if the update fails, reapplying it will lead to a consistent state if it succeeds.

While I'd prefer to have a separate resource for group assignments, this would be less consistent with the user resource.

dynatrace/api/iam/serviceusers/service.go
arthurpitman: For me, line 113 above is the exception, in general errors in these resources are not wrapped.

d0weinberger (approved):

tomazpu (approved):

PR #951, opened 2026-01-09
feat: Add `dynatrace_iam_service_user` data source

arthurpitman:
#### **Why** this PR?
This PR adds the `dynatrace_iam_service_user` data source which allows a service user to be retrieved by ID:

```
data "dynatrace_iam_service_user" "sus" {
    id = "01234567-0123-4567-89ab-cdef01234567"
} 
```

or email:

```
data "dynatrace_iam_service_user" "sus" {
    email = "01234567-0123-4567-89ab-cdef01234567@service.sso.dynatrace.com"
}
```

or name:

```
data "dynatrace_iam_service_user" "su" {
  name = "test"
} 
```

The data block will then have a value like:
```
{
  "description" = "Test service user"
  "email" = "01234567-0123-4567-89ab-cdef01234567@service.sso.dynatrace.com"
  "groups" = tolist([
    "01234567-0123-4567-89ab-cdef01234567",
  ])
  "id" = "01234567-0123-4567-89ab-cdef01234567"
  "name" = "test"
}
```

#### **What** has changed?
Adds `dynatrace_iam_service_user` data source.

#### **How** does it do it?
Adds `DataSource()` that returns a `schema.Resource` with an associated `ReadContext`.

#### How is it **tested**?
Explicit tests for the data source are added, including tests for retrieval by ID, name, and email.

`dynatrace_iam_service_user` data source is available.

**Issue:** CA-17089

### DIFF
--- a/datasources/iam/serviceusers/data_source.go
+++ b/datasources/iam/serviceusers/data_source.go
@@ -1,0 +1,181 @@
+/**
+* @license
+* Copyright 2026 Dynatrace LLC
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+ */
+
+package serviceusers
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/iam/serviceusers"
+	su "github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/iam/serviceusers/settings"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider/config"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider/logging"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func DataSource() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: logging.EnableDSCtx(DataSourceRead),
+		Description: "Fetches IAM service user details by name, ID, or email",
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Description:  "The name of the service user",
+				ExactlyOneOf: []string{"name", "id", "email"},
+			},
+			"id": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Description:  "The UUID of the service user",
+				ExactlyOneOf: []string{"name", "id", "email"},
+			},
+			"email": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				Description:  "The email of the service user",
+				ExactlyOneOf: []string{"name", "id", "email"},
+			},
+			"description": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The description of the service user",
+			},
+			"groups": {
+				Type:        schema.TypeSet,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Computed:    true,
+				Description: "The UUIDs of the groups the service user belongs to",
+			},
+		},
+	}
+}
+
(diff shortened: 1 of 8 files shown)


datasources/iam/serviceusers/data_source.go
```diff
@@ -0,0 +1,188 @@
...
+				Description:  "The email of the service user",
+				ExactlyOneOf: []string{"name", "id", "email"},
+			},
+			"description": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The description of the service user",
+			},
+			"groups": {
+				Type:        schema.TypeList,
```
Kirdock: Shouldn't that be a Set instead?
On the one hand, it's TypeSet in the service user resource, and on the other hand, it's TypeList in the users data source.
But it should not matter much because we are sorting it.

d0weinberger (dismissed):

templates/data-sources/iam_service_user.md.tmpl
```diff
@@ -0,0 +1,64 @@
...
+# dynatrace_iam_service_user (Data Source)
+
+-> **Dynatrace SaaS only**
+
+-> To utilize this resource, please define the environment variables `DT_CLIENT_ID`, `DT_CLIENT_SECRET`, `DT_ACCOUNT_ID` with an OAuth client including the following permission: **Allow read access for identity resources (users and groups)** (`account-idm-read`).
+
+This data source allows you to specify the email address of the user and produces an ordered list of group IDs this user is a member of
+
+## Example Usage
+Query a service user by name:
```
d0weinberger, line 18: It might make sense to add to the docs that an error will be thrown if there is more than one service user found for the same name.

templates/data-sources/iam_service_user.md.tmpl
```diff
@@ -0,0 +1,64 @@
...
+  The data source `dynatrace_iam_service_user` covers queries for a service user.
+---
+
+# dynatrace_iam_service_user (Data Source)
+
+-> **Dynatrace SaaS only**
+
+-> To utilize this resource, please define the environment variables `DT_CLIENT_ID`, `DT_CLIENT_SECRET`, `DT_ACCOUNT_ID` with an OAuth client including the following permission: **Allow read access for identity resources (users and groups)** (`account-idm-read`).
+
+This data source allows you to specify the email address of the user and produces an ordered list of group IDs this user is a member of
```
Kirdock:
```suggestion
This data source allows you to specify the email address of the service user and produces an ordered list of group IDs this user is a member of
```

datasources/iam/serviceusers/data_source_test.go
```diff
@@ -0,0 +1,309 @@
...
+				assert.Equal(t, "test-uid", id)
+				v.Name = "Test User"
+				v.Email = "test@example.com"
+				v.Description = "Test description"
+				v.Groups = []string{"group-1"}
+				return nil
+			},
+		}
+
+		diags := dataSourceReadWithService(context.Background(), d, mockService)
```
Kirdock: nit (and for other occurrences)
```suggestion
		diags := dataSourceReadWithService(t.Context(), d, mockService)
```

datasources/iam/serviceusers/data_source.go
```diff
@@ -0,0 +1,188 @@
...
+		return "", nil, err
+	}
+
+	var foundID string
+	var foundServiceUser su.ServiceUser
+	for _, stub := range stubs {
+		var serviceUser su.ServiceUser
+		if err := service.Get(ctx, stub.ID, &serviceUser); err != nil {
+			return "", nil, err
+		}
```
Kirdock: Should we maybe change the implementation here and for GET via email? Here, it's one request per service user, which is not optimal given IAM's strict request limits. The `/iam/v1/accounts/{accountUuid}/service-users` returns the email, name and ID, therefore the filter can be applied here and only if we found the desired service user, we can fetch its groups

templates/data-sources/iam_service_user.md.tmpl
d0weinberger: Good point, @Kirdock , but then I'd say the second occurrence of `user` should also be changed to `service user`, i.e.

```suggestion
This data source allows you to specify the email address of the service user and produces an ordered list of group IDs this service user is a member of
```

arthurpitman: Somehow this sentence got lost, I have completely rewritten it now, so please take another look.

datasources/iam/serviceusers/data_source.go
arthurpitman: This makes sense, so I've updated the underlying service to offer `GetAll(...)` with this extended information.

datasources/iam/serviceusers/data_source.go
arthurpitman: I made it a set now and removed the sorting 👍

d0weinberger (approved):

Kirdock (approved):

PR #954, opened 2026-01-13
feat: resource updates for v329

Kirdock:
#### **Why** this PR?
Some APIs have changed with Dynatrace version 1.329. Therefore, the Terraform resources also need updating.

#### **What** has changed?
For most of the resources there are only documentation updates.
`dynatrace_disk_options`: new field `monitorTmpfs`. Default to `false`

#### How is it **tested**?
It's mostly about documentation.
Added the new property of `dynatrace_disk_options` to the example, which is tested.

They will see the updated documentation and can make use of the new property of `dynatrace_disk_options`.

**Issue:** CA-17601

### DIFF
--- a/dynatrace/api/builtin/disk/options/settings/settings.go
+++ b/dynatrace/api/builtin/disk/options/settings/settings.go
@@ -1,6 +1,6 @@
 /**
 * @license
-* Copyright 2020 Dynatrace LLC
+* Copyright 2025 Dynatrace LLC
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -18,6 +18,7 @@
 package options
 
 import (
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/opt"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/terraform/hcl"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
@@ -25,7 +26,8 @@ import (
 type Settings struct {
 	DisableNfsDiskMonitoring bool          `json:"disableNfsDiskMonitoring"` // Deactivate NFS monitoring on all supported systems
 	Exclusions               DiskComplexes `json:"exclusions,omitempty"`     // OneAgent automatically detects and monitors all your mount points, however you can create exception rules to remove disks from the monitoring list.
-	NfsShowAll               bool          `json:"nfsShowAll"`               // When disabled OneAgent will try to deduplicate some of nfs disks. Disabled by default, applies only to Linux hosts. Requires OneAgent 1.209 or later
+	MonitorTmpfs             bool          `json:"monitorTmpfs"`             // Activate tmpfs monitoring on Linux systems
+	NfsShowAll               *bool         `json:"nfsShowAll,omitempty"`     // When disabled OneAgent will try to deduplicate some of nfs mount points. Disabled by default, applies only to Linux hosts.
 	Scope                    *string       `json:"-" scope:"scope"`          // The scope of this setting (HOST, HOST_GROUP). Omit this property if you want to cover the whole environment.
 }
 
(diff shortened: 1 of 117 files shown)


dynatrace/api/builtin/openpipeline/bizevents/ingestsources/schema.json
```diff
@@ -2470,6 +2475,46 @@
 						{
 							"maxLength": 32,
 							"type": "LENGTH"
+						},
+						{
```
d0weinberger, line 2479: There are a lot of new preconditions added. Should these preconditions also be reflected in the preconditions for the HCL schema?

Kirdock: This is actually not about preconditions but constraints. I think we haven't done any constraint validation for resources yet. So, I would vote for either doing it for all or not doing it at all. Might be something for the future :)

dynatrace/api/builtin/problem/notifications/email/settings/email.go
```diff
@@ -55,40 +55,37 @@ func (me *Email) Schema() map[string]*schema.Schema {
 		},
 		"bcc": {
 			Type:        schema.TypeSet,
-			Optional:    true,
-			MinItems:    1,
```
d0weinberger, line 59: Interesting. Optional with MinItems set to 1. MinItems is gone now. This should not break anything though, right?

dynatrace/api/builtin/problem/notifications/http/header.go
```diff
@@ -143,21 +143,21 @@ func (me *Header) Schema() map[string]*schema.Schema {
...
 		"name": {
 			Type:        schema.TypeString,
-			Description: "The name of the HTTP header",
+			Description: "The name of the HTTP header.",
 			Required:    true,
 			ForceNew:    ForceNewOnHeaders,
 		},
 		"secret_value": {
 			Type:        schema.TypeString,
-			Description: "The value of the HTTP header as a sensitive property. May contain an empty value. `secret_value` and `value` are mutually exclusive. Only one of those two is allowed to be specified.",
```
d0weinberger, line 152: I suspect in this case, and maybe others, the more elaborate description was specifically written to make the Terraform docs more informative. Removing these descriptions make the docs worse, IMHO.

dynatrace/api/builtin/problem/notifications/email/settings/email.go
Kirdock: Looking at the schema, this is the correct definition (without minItems). The schema has `"minObjects": 0,`
So, it should not break anything

dynatrace/api/builtin/problem/notifications/http/header.go
Kirdock: On the one hand, yes, but on the other hand, the schema should be more descriptive in the first place.

dynatrace/api/builtin/problem/notifications/service.go
```diff
@@ -1,6 +1,6 @@
 /**
 * @license
-* Copyright 2020 Dynatrace LLC
+* Copyright 2025 Dynatrace LLC
```
d0weinberger, line 3: Technically, they should all be 2026 now

Kirdock: On the other hand, the year set, should be from the creation date not update date, right? 🤔

dynatrace/api/builtin/rum/web/capturecustomproperties/settings/custom_property.go
```diff
@@ -45,9 +45,9 @@ func (me *CustomProperties) UnmarshalHCL(decoder hcl.Decoder) error {
 }
 
 type CustomProperty struct {
-	CaseInsensitiveNamingEnabled bool          `json:"caseInsensitiveNamingEnabled,omitempty"` // Field name validation should be case-insensitive
```
d0weinberger, line 48: subtle change with `omitempty` removed. On purpose?

Kirdock: Yes, as it is `nullable: false`

dynatrace/api/builtin/disk/options/settings/settings.go
```diff
@@ -48,9 +50,14 @@ func (me *Settings) Schema() map[string]*schema.Schema {
 			MinItems:    1,
 			MaxItems:    1,
 		},
+		"monitor_tmpfs": {
+			Type:        schema.TypeBool,
+			Description: "Activate tmpfs monitoring on Linux systems",
+			Optional:    true, // new required property. Default to false (default bool value)
```
d0weinberger, line 56: :+1:

d0weinberger (dismissed):

dynatrace/api/builtin/problem/notifications/http/header.go
Kirdock: I just updated the preconditions, as this was missing

d0weinberger (approved):

tomazpu (approved):

PR #956, opened 2026-01-16
feat(request-naming): Adds comparison type `string_one_agent_attribute`

d0weinberger:
#### **Why** this PR?
Instead of fixing the HCL-decoding for `unknowns` in `BaseComparisonInfo` we decided to go with just adding the types of comparison that the API supports but were not yet present in the Terraform provider. Fortunately, it was only one type.

#### **What** has changed?
The comparison type `STRING_ONE_AGENT_ATTRIBUTE` is added.

#### **How** does it do it?
It follows the same pattern as the other comparison types. I actually copied the `STRING_REQUEST_ATTRIBUTE` type and adapted it. The only difference is that we completely ignore `unknowns` for this new type.

#### How is it **tested**?
I added a Terraform example file which is first used for manual testing, but which also will be tested in E2E tests.

Users will be able to configure OneAgent attribute comparisons for APIs such as calculated service metrics or request naming.

**Issue:**
CA-17511


dynatrace/api/v1/config/metrics/calculated/service/settings/comparisoninfo/string_one_agent_attribute.go
```diff
@@ -0,0 +1,154 @@
...
+			Required:    true,
+			Description: "The OneAgent attribute to extract from",
+		},
+	}
+}
+
+func (me *StringOneAgentAttribute) MarshalHCL(properties hcl.Properties) error {
+	if err := properties.Unknowns(me.Unknowns); err != nil {
+		return err
+	}
```
Kirdock, line 79: If we ignore the `Unknowns` for this type, do we still need this?
Also, do we need the `MarshalJSON` and `UnmarshalJSON` (not speaking about the HCL marshal) if we don't do anything with `Unknowns`?

dynatrace/api/v1/config/metrics/calculated/service/settings/comparisoninfo/string_one_agent_attribute.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+// StringOneAgentAttribute Comparison for `STRING_ONE_AGENT_ATTRIBUTE` attributes, specifically of the **String** type.
+type StringOneAgentAttribute struct {
+	BaseComparisonInfo
+	Value                *string                          `json:"value,omitempty"`         // The value to compare to.
+	Values               []string                         `json:"values,omitempty"`        // The values to compare to.
+	CaseSensitive        *bool                            `json:"caseSensitive,omitempty"` // The comparison is case-sensitive (`true`) or not case-sensitive (`false`).
+	Comparison           StringRequestAttributeComparison `json:"comparison"`              // Operator of the comparision. You can reverse it by setting **negate** to `true`.
```
Kirdock: This should be `StringOneAgentAttributeComparison`, right?

d0weinberger: Yes, true!

dynatrace/api/v1/config/metrics/calculated/service/settings/comparisoninfo/string_one_agent_attribute.go
d0weinberger: No, we don't, but I'd like to deal with this in the other ticket where we get rid of `Unknowns` everywhere where it is not used.

Kirdock (approved):

PR #962, opened 2026-01-22
fix: `dynatrace_iam_policy_bindings_v2` incorrect policy binding update

Kirdock:
#### **Why** this PR?
If a policy is removed in the `dynatrace_iam_policy_bindings_v2` resource, an update won’t have any effect. It will show that the policy will be removed in the plan, but it won’t actually be removed.

#### **What** has changed?
Policy bindings can now be removed.

#### **How** does it do it?
By not deleting/updating only existing bindings defined in the resource but also comparing them against the deployed ones, so that non-existent ones in the resource are actually deleted.

#### How is it **tested**?
A new test was added that covers updating a binding with a new policy, a removed policy, and a modified policy.

Updating existing bindings via removing policies is now working.

**Issue:** CA-17779

### DIFF
--- a/dynatrace/api/iam/v2bindings/service.go
+++ b/dynatrace/api/iam/v2bindings/service.go
@@ -120,7 +120,7 @@ type BindingsResponse struct {
 	} `json:"policyBindings"`
 }
 
-func (me *BindingServiceClient) getGroupPolicyBindingIDs(ctx context.Context, id string) ([]string, error) {
+func (me *BindingServiceClient) getGroupPolicyBindingUUIDs(ctx context.Context, id string) ([]string, error) {
 	groupID, levelType, levelID, err := splitID(id)
 	if err != nil {
 		return nil, err
@@ -145,7 +145,7 @@ func (me *BindingServiceClient) Get(ctx context.Context, id string, v *bindings.
 	}
 	client := iam.NewIAMClient(me)
 
-	policyIDs, err := me.getGroupPolicyBindingIDs(ctx, id)
+	policyIDs, err := me.getGroupPolicyBindingUUIDs(ctx, id)
 	if err != nil {
 		return err
 	}
@@ -222,7 +222,7 @@ func (me *BindingServiceClient) Update(ctx context.Context, id string, v *bindin
 
 	client := iam.NewIAMClient(me)
 
-	deployedBindings, err := me.getGroupPolicyBindingIDs(ctx, id)
+	deployedBindings, err := me.getGroupPolicyBindingUUIDs(ctx, id)
 	if err != nil {
 		return err
 	}
@@ -402,7 +402,7 @@ func (me *BindingServiceClient) Delete(ctx context.Context, id string) error {
 	if err != nil {
 		return err
 	}
-	policyIDs, err := me.getGroupPolicyBindingIDs(ctx, id)
+	policyIDs, err := me.getGroupPolicyBindingUUIDs(ctx, id)
 	policyUUIDs := map[string]string{}
 	for _, policy := range policyIDs {
 		policyUUID, _, _, err := policies.SplitID(policy, levelType, levelID)
--- a/dynatrace/api/iam/v2bindings/service.go
+++ b/dynatrace/api/iam/v2bindings/service.go
@@ -120,6 +120,23 @@ type BindingsResponse struct {
 	} `json:"policyBindings"`
 }
 
+func (me *BindingServiceClient) getGroupPolicyBindingIDs(ctx context.Context, id string) ([]string, error) {
+	groupID, levelType, levelID, err := splitID(id)
+	if err != nil {
+		return nil, err
+	}
+	client := iam.NewIAMClient(me)
+
+	type groupPolicyBindings struct {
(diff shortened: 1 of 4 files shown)


dynatrace/api/iam/v2bindings/service.go
```diff
@@ -120,6 +120,23 @@ type BindingsResponse struct {
 	} `json:"policyBindings"`
 }
 
+func (me *BindingServiceClient) getGroupPolicyBindingIDs(ctx context.Context, id string) ([]string, error) {
```
d0weinberger: What kind of `id` is this?

Kirdock: The ID of the resource that we're concatenating
```go
fmt.Sprintf("%s#-#%s#-#%s", binding.GroupID, levelType, levelID)
```

dynatrace/api/iam/v2bindings/service.go
```diff
@@ -201,31 +222,57 @@ func (me *BindingServiceClient) Update(ctx context.Context, id string, v *bindin
...
+	}
 
-	for _, policy := range policiesList {
-		policyUUID, _, _, _ := policies.SplitID(policy.ID, levelType, levelID)
-		if _, err = client.DELETE_MULTI_RESPONSE(ctx, fmt.Sprintf("%s/iam/v1/repo/%s/%s/bindings/%s/%s", me.endpointURL, levelType, levelID, policyUUID, groupID), []int{204, 400, 404}, false); err != nil {
-			return err
+	// update/delete
+	for _, existingPolicyID := range existingPolicyBindings {
+		found := false
+		for _, desiredPolicy := range v.Policies {
```
d0weinberger: You could prepare a map based on `v.Policies` with `policyUUID` as a key and then in the for-loop access it with `existingPolicyID`, then you wouldn't need the nested loop and only iterate over each policy in `v.Policies` only once.

d0weinberger (dismissed):
Looks good, just a bit too many nested for loops in `Update`. I think that can be made a bit more efficient.

dynatrace/api/iam/v2bindings/service.go
Kirdock: done 👍

d0weinberger (dismissed):

dynatrace/api/iam/v2bindings/service_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func TestAccV2Bindings(t *testing.T) {
+	if !api.AccEnvsGiven(t) {
+		return
+	}
+
+	t.Setenv("TF_VAR_ACCOUNT_ID", os.Getenv("DT_ACCOUNT_ID"))
```
arthurpitman: nit: This implies the test setup must use `DT_ACCOUNT_ID` rather than `DYNATRACE_ACCOUNT_ID`, or the test will fail.

dynatrace/api/iam/v2bindings/service.go
arthurpitman: Does this function return IDs or UUIDs? I would assume the latter from the return statement `return policyBindings.PolicyUuids, nil`, but the name suggests otherwise. Furthermore, I see we don't split the result of this function later on, suggesting they're already UUIDs.

dynatrace/api/iam/v2bindings/service_test.go
```diff
@@ -0,0 +1,61 @@
...
+			return provider.Provider(), nil
+		},
+	}
+
+	t.Run("Removal of policies works", func(t *testing.T) {
+		testCase := resource.TestCase{
+			ProviderFactories: providerFactories,
+			Steps: []resource.TestStep{
+				{Config: configCreate},
+				{Config: configUpdate}, // removes one policy, adds a new policy and updates an existing policy
```
d0weinberger, line 56: I'm not convinced that any policy is updated. I would guess that all are removed and three new ones are added, because the #name# will be different each time, right? So where is the link between the policy in `create.tf` and the policy in `update.tf` that makes an update possible?

arthurpitman: I set breakpoints and can see that indeed one is removed, one is modified and one is added in the update step. Could it be that the name is kept the same, as these are steps in the one test? I will have a look 👀

arthurpitman: Correct name is randomized when TF each file is read, however, in this instance the state is maintained between the steps as the resource names remain the same (no use of `#name#` here). So the actual IDs remain the same.

arthurpitman (approved):

dynatrace/api/iam/v2bindings/service_test.go
Kirdock: Yes, the name change doesn't matter here. It's not recreated. The removed policies are checked because for each step, Terraform does a plan, and if this is non-empty (e.g., there are policies defined in a group on the API side but not the resource side), the test would fail.

PR #963, opened 2026-01-22
refactor: use core client for iam client

Kirdock:
#### **Why** this PR?
The IAM resources don't perform the OAuth flow correctly (refreshing). To improve maintenance, we should use our core client

#### **What** has changed?
- The OAuth flow is now done properly.
- The maximum timeout for 429 and 504 retries is now 10 minutes (before unlimited)

#### **How** does it do it?
By using the core client for executing the requests instead of the custom implementation

#### How is it **tested**?
Current E2E tests cover it

Reduced the maximum timeout for requests

**Issue:** CA-17703


dynatrace/api/iam/iam_client.go
```diff
@@ -169,104 +172,52 @@ func (me *iamClient) request(ctx context.Context, url string, method string, exp
 	}
 }
 
-func (me *iamClient) _request(ctx context.Context, url string, method string, expectedResponseCodes []int, forceNewBearer bool, forceNewBearerRetryCount int, payload any, headers map[string]string) ([]byte, error) {
-	// httplog(fmt.Sprintf("[%s] %s", method, url))
-
+func (me *iamClient) _request(ctx context.Context, u string, method string, expectedResponseCodes []int, forceNewBearer bool, forceNewBearerRetryCount int, payload any, headers map[string]string) ([]byte, error) {
```
d0weinberger, line 175: I think `u` should still be `url`

dynatrace/api/iam/iam_client.go
```diff
@@ -58,33 +62,32 @@ type IAMClient interface {
...
-	}
-	if len(strings.TrimSpace(bearerToken)) == 0 {
-		return errors.New(msgInvalidOAuthCredentials)
+	httpClient := auth.NewOAuthClient(context.Background(), &oauthConfig)
+
+	opts := []rest2.Option{
+		rest2.WithHTTPListener(logging.HTTPListener("iam")),
+		rest2.WithRateLimiter(),
+		rest2.WithRetryOptions(&rest2.RetryOptions{
+			DelayAfterRetry: time.Second * 60,
```
d0weinberger, line 81: 60 seconds because this is basically what we would get from the retry header in the response, if we were to read it, right?

d0weinberger (dismissed):

dynatrace/api/iam/iam_client.go
Kirdock: yes

dynatrace/api/iam/iam_client.go
Kirdock: This conflicted with the url package import that's why I renamed it

arthurpitman (approved):

d0weinberger (approved):

PR #965, opened 2026-01-23
test(iam): Adds dynatrace_iam_user & dynatrace_iam_group E2E tests

d0weinberger:
#### **Why** this PR?

#### **What** has changed?
E2E tests have been added for 
- `dynatrace_iam_user`
- `dynatrace_iam_group`

##### `dynatrace_iam_user`:
Two steps:
1. A user with a group is created
2. The user is updated, i.e. the user is removed from the group.
##### `dynatrace_iam_group`:
Two steps:
1. A simple group is created
2. This updates the description, federated attributes and permissions of the group

#### How is it **tested**?
This is the test.

It doesn't.

**Issue:** CA-17713

### DIFF
--- a/dynatrace/api/iam/groups/testdata/update.tf
+++ b/dynatrace/api/iam/groups/testdata/update.tf
@@ -1,0 +1,21 @@
+variable "GROUP_NAME" {
+  description = "The name of the group."
+  type        = string
+}
+
+variable "ACCOUNT_ID" {
+  description = "The ID of the Dynatrace account."
+  type        = string
+}
+
+resource "dynatrace_iam_group" "my-group" {
+  name = var.GROUP_NAME
+  description = "A group created for e2e testing with specific permissions."
+  permissions {
+    permission {
+      name  = "account-viewer"
+      type  = "account"
+      scope = var.ACCOUNT_ID
+    }
+  }
+}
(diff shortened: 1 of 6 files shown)


dynatrace/api/iam/users/testdata/create.tf
```diff
@@ -0,0 +1,23 @@
...
+    type        = string
+}
+
+// Every user always belongs to at least "Default group with all users"
+data "dynatrace_iam_group" "all_users_group" {
+  name = "Default group with all users"
+}
+
+resource "dynatrace_iam_group" "my_group" {
+  name          = "#name#"
```
Kirdock: Should we use a variable like for the other tests?

dynatrace/api/iam/groups/testdata/update.tf
```diff
@@ -0,0 +1,16 @@
+variable "ACCOUNT_ID" {
+  description = "The ID of the Dynatrace account."
+  type        = string
+}
+
+resource "dynatrace_iam_group" "my-group" {
+  name = "e2e-test-group-with-permissions"
```
Kirdock: Is a group name unique? If yes, we should make it unique

Kirdock (approved):

PR #966, opened 2026-01-26
fix: add fixed waiting time for synthetic monitor updates

Kirdock:
#### **Why** this PR?
Synthetic monitors are eventually consistent, not only for creation but also for updates. Meaning a GET after a create or an update may lead to an incomplete result (some properties are updated, and some aren't)

#### **What** has changed?
The state is now properly set after the update for `dynatrace_network_monitor` and `dynatrace_http_monitor`.

#### **How** does it do it?
By adding a 5-second sleep duration after the update. It has gotten worse last week, and 3 seconds aren't enough anymore.

#### How is it **tested**?
The test should not be flaky anymore

They're less likely to run into `terraform plan` updates after `terraform apply` for `dynatrace_network_monitor` and `dynatrace_http_monitor`.

**Issue:**

### DIFF
--- a/dynatrace/api/v1/config/synthetic/monitors/http/service.go
+++ b/dynatrace/api/v1/config/synthetic/monitors/http/service.go
@@ -19,8 +19,10 @@ package http
 
 import (
 	"context"
+	"time"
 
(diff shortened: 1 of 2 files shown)


Kirdock:
Even with a huge waiting time this problem may still occur

PR #967, opened 2026-01-26
test(permission): Adds dynatrace_iam_permission E2E test

d0weinberger:
#### **Why** this PR?

#### **What** has changed?
E2E tests have been added for `dynatrace_iam_permission`

#### **How** does it do it?
Two steps:
1. A group is created and then a `dynatrace_iam_permission` resource is created which grants some permission to said group.
2. The permission resource is removed again.

#### How is it **tested**?
This is the test.

It doesn't.

**Issue:** CA-17713

### DIFF
--- a/dynatrace/api/iam/permissions/service_test.go
+++ b/dynatrace/api/iam/permissions/service_test.go
@@ -1,0 +1,64 @@
+//go:build integration
+
+/*
+ * @license
+ * Copyright 2026 Dynatrace LLC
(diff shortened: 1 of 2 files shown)


Kirdock:
I'm not exactly sure what the update should test. We have the ignore_changes, so anything related to this should not matter anyway

d0weinberger:
Then I'm just going to remove the update again

Kirdock (approved):

PR #968, opened 2026-01-26
test(iam): Adds tests for policy, boundary, and binding resources

d0weinberger:
#### **Why** this PR?

#### **What** has changed?
E2E tests have been added for 
- `dynatrace_iam_policy`
- `dynatrace_iam_policy_boundary`
- `dynatrace_iam_policy_bindings`

#### **How** does it do it?
For each resource there is a `create.tf` and an `update.tf` file.
So the respective resource is created and afterwards updated.

##### Limitation
I only included account policies. 
I did not want to assume an environment via env var for environment policies.

#### How is it **tested**?
This is the test.

It doesn't.

**Issue:** CA-17713

### DIFF
--- a/dynatrace/api/iam/bindings/service_test.go
+++ b/dynatrace/api/iam/bindings/service_test.go
@@ -1,0 +1,70 @@
+//go:build integration
+
+/*
+ * @license
+ * Copyright 2026 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package bindings_test
+
+import (
+	"os"
+	"testing"
+
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/testing/api"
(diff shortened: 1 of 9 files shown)


dynatrace/api/iam/bindings/service_test.go
```diff
@@ -0,0 +1,70 @@
...
+	if !api.AccEnvsGiven(t) {
+		return
+	}
+
+	accountID := os.Getenv("DT_ACCOUNT_ID")
+	//fallback to DYNATRACE_ACCOUNT_ID
+	if accountID == "" {
+		accountID = os.Getenv("DYNATRACE_ACCOUNT_ID")
+	}
+	t.Setenv("TF_VAR_ACCOUNT_ID", accountID)
```
Kirdock, line 42: As we use this more and more, should we move this to the testing folder?
We probably want to have the same thing for the environment name, too.
Speaking of which, as a follow-up, we could adjust the pipeline to include the environment name, maybe already as `TF_VAR` 🤔

dynatrace/api/iam/v2bindings/testdata/create.tf
Kirdock, line 1: Either here or in the other open PR: The file names here will result in a merge conflict with https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/962. Just to take care of, depending on which one is merged first

Kirdock (dismissed):

Kirdock:
On the one hand it's nice to use builtin logic for the names, but on the other hand, manually executing the tests without the testing framework may be a bit cumbersome. Depending on which files you are debugging/executing you need quite a lot of different variables. Should we maybe just have a random identifier variable (besides the account ID)?
Or maybe I'm just the only one that executes the tf files manually.
@arthurpitman your opinion?

dynatrace/api/iam/v2bindings/testdata/create.tf
d0weinberger: I removed my tests

Kirdock (approved):

PR #969, opened 2026-01-28
feat: resource updates for v330

Kirdock:
#### **Why** this PR?
Some APIs have changed with Dynatrace version 1.329. Therefore, the Terraform resources also need updating.

#### **What** has changed?
For most of the resources there are only documentation updates.
`dynatrace_mobile_app_enablement`: new optional field `experienceAnalytics`.
`dynatrace_openpipeline_v2_*_ingestsources`: The `eventType` of `BizeventAttributes` and `SdlcEventAttributes` is now required
`dynatrace_openpipeline_v2_*_pipelines`: The `eventType` of `BizeventAttributes` and `SdlcEventAttributes` is now required. New optional fields `group_role` and `routing`.
`dynatrace_host_anomalies_v2`: host `gc` (`highGcActivityDetection`) is now optional
Some missing preconditions were added (property A conflicts with property B, etc.) for `dynatrace_host_anomalies_v2`, `dynatrace_endpoint_detection_rules` and `dynatrace_generic_relationships`,  

#### How is it **tested**?
It's mostly about documentation.

They will see the updated documentation, will have breaking changes due to the openpipeline updates (new required fields) and can make use of the new properties mentioned above.

**Issue:** CA-17857

### DIFF
--- a/dynatrace/api/builtin/anomalydetection/infrastructure/hosts/settings/connection_lost_detection.go
+++ b/dynatrace/api/builtin/anomalydetection/infrastructure/hosts/settings/connection_lost_detection.go
@@ -18,26 +18,28 @@
 package hosts
 
 import (
+	"fmt"
+
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/terraform/hcl"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
 type ConnectionLostDetection struct {
-	Enabled             bool                                `json:"enabled"`                       // Detect host or monitoring connection lost problems
-	OnGracefulShutdowns *ConnectionLostDetectionSensitivity `json:"onGracefulShutdowns,omitempty"` // Graceful host shutdowns
+	Enabled             bool                                `json:"enabled"`                       // This setting is enabled (`true`) or disabled (`false`)
+	OnGracefulShutdowns *ConnectionLostDetectionSensitivity `json:"onGracefulShutdowns,omitempty"` // Graceful host shutdowns. Possible Values: `ALERT_ON_GRACEFUL_SHUTDOWN`, `DONT_ALERT_ON_GRACEFUL_SHUTDOWN`
 }
 
 func (me *ConnectionLostDetection) Schema() map[string]*schema.Schema {
 	return map[string]*schema.Schema{
 		"enabled": {
 			Type:        schema.TypeBool,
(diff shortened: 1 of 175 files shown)


dynatrace/api/builtin/anomalydetection/infrastructure/hosts/settings/connection_lost_detection.go
```diff
@@ -49,6 +51,16 @@ func (me *ConnectionLostDetection) MarshalHCL(properties hcl.Properties) error {
 	})
 }
 
+func (me *ConnectionLostDetection) HandlePreconditions() error {
+	if (me.OnGracefulShutdowns == nil) && (me.Enabled) {
+		return fmt.Errorf("'on_graceful_shutdowns' must be specified if 'enabled' is set to '%v'", me.Enabled)
+	}
+	if (me.OnGracefulShutdowns != nil) && (!me.Enabled) {
+		return fmt.Errorf("'on_graceful_shutdowns' must not be specified if 'enabled' is set to '%v'", me.Enabled)
```
tomazpu, line 59: NIT: The Error message could be better; both cases of validation provide the same error message.

dynatrace/api/builtin/monitoredentities/generic/relation/settings/enums.go
```diff
@@ -44,9 +44,9 @@ var Normalizations = struct {
 	Tolowercase   Normalization
 	Touppercase   Normalization
 }{
-	"Leavetextas_is",
-	"Tolowercase",
-	"Touppercase",
+	"Leave text as-is",
```
d0weinberger, line 47: That's a breaking change, right?

Kirdock: No. We don't validate enums and the API didn't change anything here. This is then only about the documentation, which was wrong in the first place

dynatrace/api/builtin/anomalydetection/infrastructure/hosts/settings/connection_lost_detection.go
Kirdock: It's not exactly the same. The first one is "must be specified if", the second one is "must 'not' be specified if"
I would rather keep it consistent between each resource than have manual updates to a specific one.

tomazpu (approved):

PR #974, opened 2026-02-03
fix(document): `dynatrace_direct_shares` supports documents with custom IDs

arthurpitman:
#### **Why** this PR?
The `dynatrace_direct_shares` had a validator requiring the specified `document_id` to be a UUID. As documents now support custom IDs that must not be UUIDs, this requirement is no longer valid.

The validator requiring `document_id` to be a UUID is removed.

#### How is it **tested**?
An explicit test creating a direct share for a document with a (non-UUID) custom ID is added.

Users can now share documents with custom IDs.

**Issue:** CA-17840

### DIFF
--- a/dynatrace/api/documents/directshares/testdata/terraform/example-custom-id.tf
+++ b/dynatrace/api/documents/directshares/testdata/terraform/example-custom-id.tf
@@ -1,0 +1,42 @@
+resource "dynatrace_iam_group" "this" {
+  name = "#name#"
+}
+
+resource "dynatrace_document" "this" {
+  custom_id = "#name#"
+  type      = "dashboard"
+  name      = "#name#"
+  content = jsonencode(
+    {
+      "version" : 1,
+      "variables" : [],
+      "tiles" : {
+        "0" : {
+          "type" : "markdown",
+          "title" : "",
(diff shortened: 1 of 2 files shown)


dynatrace/api/documents/directshares/testdata/terraform/example-custom-id.tf
```diff
@@ -0,0 +1,42 @@
+resource "dynatrace_iam_group" "this" {
+  name = "#name#"
+}
```
arthurpitman, line 3: Test is now improved to create the group used

dynatrace/api/documents/directshares/testdata/terraform/example-custom-id.tf
```diff
@@ -0,0 +1,42 @@
...
+
+resource "dynatrace_document" "this" {
+  custom_id = "my-custom-dashboard-id"
```
Kirdock: Should we also change this one to `#name#`? This can conflict with another pipeline running in parallel

arthurpitman: Updated 👍

Kirdock (approved):

PR #975, opened 2026-02-04
fix: not working request attribute fields

Kirdock:
#### **Why** this PR?
Some fields in `dynatrace_request_attribute` didn't work as expected.
Affected fields: cics_transaction_call_type, iib_node_type_condition, ims_transaction_call_type, server_variable_technology, span_attribute_key

#### **What** has changed?
Applying and planning now correctly returns the affected fields.

#### **How** does it do it?
The fields are now correctly mapped to and from the API (HCL to JSON and vice versa)

#### How is it **tested**?
A new example has been added that makes use of some other properties.

They can now properly use the fields cics_transaction_call_type, iib_node_type_condition, ims_transaction_call_type, server_variable_technology and span_attribute_key of a data source

**Issue:** NOISSUE (found during CA-17770)

### DIFF
--- a/dynatrace/api/v1/config/requestattributes/settings/data_source.go
+++ b/dynatrace/api/v1/config/requestattributes/settings/data_source.go
@@ -399,7 +399,7 @@ func (me *DataSource) MarshalJSON() ([]byte, error) {
 	if err := m.Marshal("cicsTransactionCallType", me.CICSTransactionCallType); err != nil {
 		return nil, err
 	}
-	if err := m.Marshal("iibNodeTypeCondition", me.IIBMethodNodeCondition); err != nil {
+	if err := m.Marshal("iibNodeTypeCondition", me.IIBNodeTypeCondition); err != nil {
 		return nil, err
 	}
 	if err := m.Marshal("imsTransactionCallType", me.IMSTransactionCallType); err != nil {
@@ -461,7 +461,7 @@ func (me *DataSource) UnmarshalJSON(data []byte) error {
 	if err := m.Unmarshal("cicsTransactionCallType", &me.CICSTransactionCallType); err != nil {
 		return err
 	}
-	if err := m.Unmarshal("iibNodeTypeCondition", &me.IIBMethodNodeCondition); err != nil {
+	if err := m.Unmarshal("iibNodeTypeCondition", &me.IIBNodeTypeCondition); err != nil {
 		return err
 	}
 	if err := m.Unmarshal("imsTransactionCallType", &me.IMSTransactionCallType); err != nil {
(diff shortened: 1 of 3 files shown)


dynatrace/api/v1/config/requestattributes/settings/data_source.go
```diff
@@ -396,19 +396,19 @@ func (me *DataSource) MarshalJSON() ([]byte, error) {
 	if err := m.Marshal("iibNodeType", me.IIBNodeType); err != nil {
 		return nil, err
 	}
-	if err := m.Marshal("cics_transaction_call_type", me.CICSTransactionCallType); err != nil {
+	if err := m.Marshal("cicsTransactionCallType", me.CICSTransactionCallType); err != nil {
 		return nil, err
 	}
-	if err := m.Marshal("iib_node_type_condition", me.IIBMethodNodeCondition); err != nil {
+	if err := m.Marshal("iibNodeTypeCondition", me.IIBMethodNodeCondition); err != nil {
```
d0weinberger: Shouldn't that be `IIBNodeTypeCondition`?

dynatrace/api/v1/config/requestattributes/settings/data_source.go
```diff
@@ -458,19 +458,19 @@ func (me *DataSource) UnmarshalJSON(data []byte) error {
 	if err := m.Unmarshal("iibNodeType", &me.IIBNodeType); err != nil {
 		return err
 	}
-	if err := m.Unmarshal("cics_transaction_call_type", &me.CICSTransactionCallType); err != nil {
+	if err := m.Unmarshal("cicsTransactionCallType", &me.CICSTransactionCallType); err != nil {
 		return err
 	}
-	if err := m.Unmarshal("iib_node_type_condition", &me.IIBMethodNodeCondition); err != nil {
+	if err := m.Unmarshal("iibNodeTypeCondition", &me.IIBMethodNodeCondition); err != nil {
```
d0weinberger: Here as well, I think that should be `IIBNodeTypeCondition`, no?

dynatrace/api/v1/config/requestattributes/settings/data_source.go
Kirdock: You are right! There were acutally two issues in this line. Fixed it 👍

dynatrace/api/v1/config/requestattributes/settings/data_source.go
Kirdock: You are right! There were acutally two issues in this line. Fixed it 👍

d0weinberger (approved):

PR #977, opened 2026-02-06
refactor: Centralize environment variables

arthurpitman:
#### **Why** this PR?

This PR centralizes all environment variable access into a single `envutils` package with typed wrappers to ensure that environment variables are all processed consistently, reducing duplication and increasing maintainability.

Every environment variable is defined in one place (`provider/envutils/env_vars.go`), with a GoDoc explaining the purpose of each one.

- Five typed wrappers are defined in `provider/envutils/envutils.go`: `BoolEnvVar`, `StringEnvVar`, `ClampedIntEnvVar`, `BoundedIntEnvVar`, and `MultiStringEnvVar`, each with a `Get()` method that reads the variable(s) using `os.LookupEnv(...)`, parses the value, and returns a typed result (or the default).

- Then, the PR progressively adds to `provider/envutils/env_vars.go`, where each environment variable is declared with its key, default, and bounds where applicable.

- The code is refactored to always read from environment variables rather than temporarily store values in package-level variables  

- `ConfigGetter.Get()` always checked the source URL environment variables regardless of which config key was requested. This meant source overrides for API tokens, client IDs, secrets, and other non-URL keys were silently ignored. The fix ensures each key resolves its own corresponding source environment variable.

#### How is it **tested**?
Tests are added for the `Get(...)` method of `BoolEnvVar`, `StringEnvVar`, `IntEnvVar`, `ClampedIntEnvVar`, `BoundedIntEnvVar`, and `MultiStringEnvVar`.

There should be no substantial change as the environment variables should continue to work as before.  However, there are a few behavioral changes worth noting:
- `DYNATRACE_PARALLEL` — Previously accepted any string; "false" disabled parallelism, anything else (including unset) enabled it. Now uses strconv.ParseBool, so only standard boolean strings ("true", "false", "1", "0", "t", "f", etc.) are recognized. Setting it to e.g. "yes" previously enabled parallelism; now it falls back to the default (true), which happens to be the same result — but for different reasons.
- `DYNATRACE_HEREDOC` — Same pattern. Previously "false" disabled heredoc, anything else enabled it. Now uses strconv.ParseBool. Non-standard values like "no" previously kept heredoc enabled (default); now they still fall back to the default (true). Functionally equivalent for all practical inputs.
- `DT_CACHE_DELETE_ON_LAUNCH` — Previously any non-empty, non-"false" string triggered cache deletion. Now uses strconv.ParseBool. Setting it to e.g. "yes" or "delete" previously triggered deletion; now those values fall back to the default (false) and will not trigger deletion. This is a real behavioral change for users who set non-standard truthy values.
- `DYNATRACE_DASHBOARD_TESTS` — Previously any non-empty string activated dashboard tests. Now requires a valid boolean. Setting it to e.g. "verbose" previously activated tests; now it falls back to false. Real behavioral change for non-standard values.
- `DYNATRACE_HOST_MONITORING_STRICT_UPDATE_RETRIES` — Previously had no upper bound; any non-negative integer was accepted. Now clamped to [0, 60]. Values above 60 are silently reduced to 60. Real behavioral change for users setting values > 60.

**Issue:** CA-18065

### DIFF
--- a/dynatrace/api/automation/workflows/settings/task.go
+++ b/dynatrace/api/automation/workflows/settings/task.go
@@ -1,6 +1,6 @@
 /**
 * @license
-* Copyright 2020 Dynatrace LLC
+* Copyright 2026 Dynatrace LLC
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -19,9 +19,9 @@ package workflows
 
 import (
 	"encoding/json"
-	"os"
 	"regexp"
 
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider/envutils"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/terraform/hcl"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
@@ -54,10 +54,8 @@ func (me *Tasks) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
-var useTypeList = os.Getenv("DYNATRACE_WORKFLOW_TASKS_USE_TYPE_LIST") == "true"
-
 func (me *Tasks) Schema(prefix string) map[string]*schema.Schema {
-	if useTypeList {
+	if envutils.DynatraceWorkflowTasksUseTypeList.Get() {
 		return map[string]*schema.Schema{
 			"task": {
 				Type:        schema.TypeList,
--- a/dynatrace/api/builtin/host/monitoring/mode/service.go
+++ b/dynatrace/api/builtin/host/monitoring/mode/service.go
@@ -1,6 +1,6 @@
 /**
 * @license
-* Copyright 2020 Dynatrace LLC
+* Copyright 2026 Dynatrace LLC
 *
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
@@ -31,13 +31,13 @@ import (
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/rest"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/settings"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/settings/services/settings20"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider/envutils"
 )
 
 const SchemaVersion = "1.3"
 const SchemaID = "builtin:host.monitoring.mode"
 
 var WarnOnAgentOffline = os.Getenv("DYNATRACE_HOST_MONITORING_WARNINGS") == "true"
-var ExportOfflineHosts = os.Getenv("DYNATRACE_HOST_MONITORING_OFFLINE") == "true"
 var StrictUpdateRetries = os.Getenv("DYNATRACE_HOST_MONITORING_STRICT_UPDATE_RETRIES")
 
 func Service(credentials *rest.Credentials) settings.CRUDService[*mode.Settings] {
@@ -77,7 +77,7 @@ func (me *service) List(ctx context.Context) (api.Stubs, error) {
 			if len(entity.Properties.State) == 0 {
 				continue
 			}
-			if !ExportOfflineHosts && (entity.Properties.State == "OFFLINE" || entity.Properties.State == "SHUTDOWN" || entity.Properties.State == "MONITORING_DISABLED") {
+			if !envutils.DynatraceHostMonitoringOffline.Get() && (entity.Properties.State == "OFFLINE" || entity.Properties.State == "SHUTDOWN" || entity.Properties.State == "MONITORING_DISABLED") {
 				continue
 			}
 			if entity.Properties.MonitoringMode == "INFRASTRUCTURE" {
--- a/dynatrace/api/builtin/host/monitoring/mode/service.go
+++ b/dynatrace/api/builtin/host/monitoring/mode/service.go
@@ -21,8 +21,6 @@ import (
 	"context"
 	"fmt"
 	"net/url"
-	"os"
-	"strconv"
 	"strings"
 	"time"
 
@@ -36,8 +34,6 @@ import (
 
 const SchemaVersion = "1.3"
 const SchemaID = "builtin:host.monitoring.mode"
-
-var StrictUpdateRetries = os.Getenv("DYNATRACE_HOST_MONITORING_STRICT_UPDATE_RETRIES")
 
 func Service(credentials *rest.Credentials) settings.CRUDService[*mode.Settings] {
 	return &service{
@@ -156,12 +152,7 @@ func (me *service) Update(ctx context.Context, id string, v *mode.Settings) erro
 func (me *service) update(ctx context.Context, id string, v *mode.Settings) error {
 	_, err := me.service.Create(ctx, v)
 	if err == nil {
-		remainingRetries := 0
-		if len(StrictUpdateRetries) > 0 {
-			if iStrictUpdateRetries, err := strconv.Atoi(StrictUpdateRetries); err == nil && iStrictUpdateRetries >= 0 {
-				remainingRetries = iStrictUpdateRetries
-			}
-		}
+		remainingRetries := envutils.DynatraceHostMonitoringStrictUpdateRetries.Get()
 		readMode := ""
 		for remainingRetries > 0 && readMode != string(v.MonitoringMode) {
 			getSettings := new(mode.Settings)
--- a/dynatrace/api/builtin/host/monitoring/mode/service.go
+++ b/dynatrace/api/builtin/host/monitoring/mode/service.go
@@ -37,7 +37,6 @@ import (
 const SchemaVersion = "1.3"
 const SchemaID = "builtin:host.monitoring.mode"
 
-var WarnOnAgentOffline = os.Getenv("DYNATRACE_HOST_MONITORING_WARNINGS") == "true"
 var StrictUpdateRetries = os.Getenv("DYNATRACE_HOST_MONITORING_STRICT_UPDATE_RETRIES")
 
 func Service(credentials *rest.Credentials) settings.CRUDService[*mode.Settings] {
@@ -179,7 +178,7 @@ func (me *service) update(ctx context.Context, id string, v *mode.Settings) erro
 	}
 	// we don't want to error out just because the Agent isn't online
 	if strings.Contains(err.Error(), "Monitoring mode can't be changed while OneAgent is offline") {
-		if WarnOnAgentOffline {
+		if envutils.DynatraceHostMonitoringWarnings.Get() {
 			return rest.Warning{Message: fmt.Sprintf("The host '%s' is currently offline - changing the monitoring mode was not possible", id)}
 		}
 		return nil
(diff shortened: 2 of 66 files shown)


provider/envutils/envutils.go
```diff
@@ -0,0 +1,133 @@
...
+// Get returns the environment variable's boolean value or its default value if unset or invalid.
+func (b BoolEnvVar) Get() bool {
+	v, found := os.LookupEnv(b.Key)
+	if !found {
+		return b.DefaultValue
+	}
+
+	value, err := strconv.ParseBool(v)
+	if err != nil {
+		return b.DefaultValue
```
d0weinberger, line 40: Should there be a warning emitted in this case? `Could not parse value 'invalid' for env var 'XYZ'. Falling back to default 'false'`

provider/envutils/envutils.go
```diff
@@ -0,0 +1,133 @@
...
+// Get returns the clamped integer value of the environment variable or its default value if unset or invalid.
+func (c ClampedIntEnvVar) Get() int {
+	v, found := os.LookupEnv(c.Key)
+	if !found || len(v) == 0 {
+		return c.DefaultValue
+	}
+
+	value, err := strconv.Atoi(v)
+	if err != nil {
+		return c.DefaultValue
```
d0weinberger, line 79: Same question about the warning here, and also below, when Max or Min are assumed

provider/envutils/envutils.go
arthurpitman: It would be nice, but would go beyond the scope of this PR. This one is only about centralizing the environment variables to allow us to understand which actually exist. While emitting warnings would be nice in theory, it leads to a number of problems:
- In my opinion, environment variables should not be used to configure Terraform providers. Users may not even have the ability to set environment variables.
- Given the way providers run, how would warnings be emitted? Considering the current design, this would require a huge number of changes to propagate them back to the user without overwhelming them with repetitive warnings.

My actual feeling is that, as a follow-up, we should try to remove at least 90% of these 💪

dynatrace/api/automation/workflows/settings/task.go
```diff
@@ -259,13 +258,3 @@ func (me *Task) UnmarshalHCL(decoder hcl.Decoder) error {
 	}
 	return nil
 }
-
-func (me *Task) HandlePreconditions() error {
```
d0weinberger, line 263: This change is not related to the environment variables. Was this an intentional change? Why is it now okay to remove the pointer?

provider/envutils/env_vars.go
```diff
@@ -0,0 +1,556 @@
...
+// DynatraceDashboardTests enables dashboard test behavior.
+var DynatraceDashboardTests = BoolEnvVar{
+	Key:          "DYNATRACE_DASHBOARD_TESTS",
+	DefaultValue: false,
+}
+
+// --- DQL ---
+
+// DynatraceDQLPollSleepDuration is the sleep duration in milliseconds between DQL poll attempts. Values outside [Min, Max] are clamped to the nearest boundary.
+var DynatraceDQLPollSleepDuration = ClampedIntEnvVar{
```
d0weinberger: Judging from how this one is used in [datasources/dql/data_source.go](https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/977/changes#diff-63fc5a087389cf25c491c4f036c461f8984619feb22c84ea999498ef8e4f129c) , it should be bounded, not clamped.

dynatrace/api/automation/workflows/settings/task.go
```diff
@@ -97,7 +96,7 @@ type Task struct {
 	Action       string               `json:"action" pattern:"^.+:.+$"`
 	Description  *string              `json:"description,omitempty"` // A description for this task
 	Input        map[string]any       `json:"input"`
-	Active       *bool                `json:"active,omitempty"` // Specifies whether a task should be skipped as a no operation or not
+	Active       bool                 `json:"active"` // Specifies whether a task should be skipped as a no operation or not
```
Kirdock: Incorrectly resolved merge conflict. Please fix

dynatrace/api/automation/workflows/settings/task.go
```diff
@@ -88,7 +87,7 @@ func (me *Tasks) UnmarshalHCL(decoder hcl.Decoder) error {
 	if err := decoder.DecodeSlice("task", me); err != nil {
 		return err
 	}
-	*me = hcl.FilterEmpty(*me, Task{Active: new(true), Input: map[string]any{}})
+	*me = hcl.FilterEmpty(*me, Task{Active: true, Input: map[string]any{}})
```
Kirdock: Incorrectly resolved merge conflict. Please fix

dynatrace/api/automation/workflows/settings/task.go
```diff
@@ -259,13 +258,3 @@ func (me *Task) UnmarshalHCL(decoder hcl.Decoder) error {
 	}
 	return nil
 }
-
```
Kirdock, line 262: Incorrectly resolved merge conflict. Please fix

dynatrace/api/builtin/host/monitoring/mode/service.go
```diff
@@ -157,12 +152,7 @@ func (me *service) Update(ctx context.Context, id string, v *mode.Settings) erro
 func (me *service) update(ctx context.Context, id string, v *mode.Settings) error {
 	_, err := me.service.Create(ctx, v)
 	if err == nil {
-		remainingRetries := 0
-		if len(StrictUpdateRetries) > 0 {
-			if iStrictUpdateRetries, err := strconv.Atoi(StrictUpdateRetries); err == nil && iStrictUpdateRetries >= 0 {
-				remainingRetries = iStrictUpdateRetries
-			}
-		}
+		remainingRetries := envutils.DynatraceHostMonitoringStrictUpdateRetries.Get()
```
Kirdock, line 155: [nit]: We now have a max of 60. Previously: nothing

dynatrace/api/v1/config/customtags/list/list.go
```diff
@@ -47,7 +43,7 @@ func List(ctx context.Context, client rest.Client) (api.Stubs, error) {
 		return nil, err
 	}
 
-	maxConcurrent := settings.GetIntEnvCtx(ctx, DYNATRACE_MAX_CONCURRENT_CUSTOM_TAG_LIST_REQUESTS, DefaultMaxConcurrent, MaxConcurrentMinValue, MaxConcurrentMaxValue)
+	maxConcurrent := envutils.DynatraceMaxConcurrentCustomTagListRequests.Get()
```
Kirdock, line 46: This changes a bit in behavior. Is this intended?
Also: `dynatrace/settings/get_env.go` now has some unused functions

dynatrace/api/v1/config/credentials/aws/settings/enums.go
```diff
@@ -50,7 +50,8 @@ var Types = struct {
 }
 
 // ConnectionStatus The status of the connection to the AWS environment.
-//  * `CONNECTED`: There was a connection within last 10 minutes.
+//   - `CONNECTED`: There was a connection within last 10 minutes.
+//
```
Kirdock: Why the change here and only one bullet point? `*` => `-`

dynatrace/api/v1/config/dashboards/settings/json_dashboard.go
```diff
@@ -185,7 +184,7 @@ func enrm(m map[string]any, bc string, fordiff bool) {
 	}
 	if bc == "tiles[#].filterConfig.chartConfig.series[#].dimensions[#]" {
 		ensure(m, "values", []string{})
-		if DYNATRACE_DASHBOARD_TESTS && fordiff {
+		if envutils.DynatraceDashboardTests.Get() && fordiff {
```
Kirdock, line 187: Change in behavior, but I would assume that the intention was only for testing, which can be removed now, I guess.

dynatrace/rest/logging/logger.go
```diff
@@ -32,27 +33,26 @@ type RESTLogger struct {
...
-	if stdoutLog {
+	if envutils.DynatraceLogHTTP.Get() == "stdout" {
 		tflog.Debug(ctx, fmt.Sprint(append(append([]any{}, "[HTTP]"), v...)...))
 	}
 	l.Log.Print(v...)
 }
 
 func (l *RESTLogger) Printf(ctx context.Context, format string, v ...any) {
-	if stdoutLog {
+	if envutils.DynatraceLogHTTP.Get() == "stdout" {
```
Kirdock: [nit] Should we introduce a constant here for `stdout` or small helper function?

dynatrace/api/v1/config/credentials/aws/settings/enums.go
arthurpitman: Manually fixed: I now removed one space and a new line to make it consistent

dynatrace/api/automation/workflows/settings/task.go
arthurpitman: No, thanks - should be fixed now!

dynatrace/api/v1/config/customtags/list/list.go
arthurpitman: Honestly, I'm not quite sure: settings.GetIntEnvCtx would read the value from the context if it was defined with that key, otherwise fall back to the environment variable. But I don't believe it was ever written to the context, or am I missing something?

dynatrace/api/builtin/host/monitoring/mode/service.go
arthurpitman: That is true - 60 is an arbitrary value. But leaving it unbounded also isn't great. What value makes sense?

dynatrace/api/automation/workflows/settings/task.go
arthurpitman: Should be fixed 😓

provider/envutils/env_vars.go
arthurpitman: Fixed 👍

dynatrace/rest/logging/logger.go
arthurpitman: Done 👍

dynatrace/api/builtin/host/monitoring/mode/service.go
d0weinberger: 60 retries with 20 seconds of sleep between retries = 20 minutes max, which seems plenty

dynatrace/rest/logging/logger.go
```diff
@@ -71,8 +75,8 @@ func (odw *onDemandWriter) Write(p []byte) (n int, err error) {
 }
 
 func initLogger() *RESTLogger {
-	restLogFileName := os.Getenv("DYNATRACE_LOG_HTTP")
-	if len(restLogFileName) > 0 && restLogFileName != "false" && !stdoutLog {
+	restLogFileName := envutils.DynatraceLogHTTP.Get()
+	if len(restLogFileName) > 0 && restLogFileName != "false" && restLogFileName != "stdout" {
```
d0weinberger: [nit] shouldn't that also just be ` !shouldLogHTTPToStdout() `?

arthurpitman: I think i'll add a second helper function: `shouldLogHTTPToFile()`

dynatrace/rest/logging/logger.go
arthurpitman: Now this uses constants, the minimal change. Maybe I'll take another look at this environment variable in a future PR.

dynatrace/rest/logging/logger.go
arthurpitman: Actually, I went to the minimal change: use constants

d0weinberger (dismissed):

d0weinberger (approved):

Kirdock (approved):

PR #978, opened 2026-02-06
refactor: Remove unneeded code

arthurpitman:
#### **Why** this PR?
This code does nothing.

It is deleted.

#### How is it **tested**?
Not applicable

Not applicable

**Issue:** Not applicable.

### DIFF
--- a/dynatrace/export/enums.go
+++ b/dynatrace/export/enums.go
@@ -57,38 +57,6 @@ func (rt ResourceType) IsPotentialCircularDependencyTo(referringResourceID strin
 	return rt == ResourceTypes.JSONDashboard && referredToResourceType == ResourceTypes.JSONDashboardBase && referringResourceID != referredToResourceID
 }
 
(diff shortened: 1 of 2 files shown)


d0weinberger (dismissed):

Kirdock (commented):
There is actually quite more with this.
If we remove this, the `voidResource` function is also not used anymore and with the removal of this, the `resource.Type.VoidResource(...` function is also not used anymore.

Should we do the full removal?

tomazpu (dismissed):

arthurpitman:
> There is actually quite more with this. If we remove this, the `voidResource` function is also not used anymore and with the removal of this, the `resource.Type.VoidResource(...` function is also not used anymore.
> 
> Should we do the full removal?

Very good catch - let's remove the it all!

d0weinberger (approved):

Kirdock (approved):

PR #981, opened 2026-02-11
feat(openpipeline-v2): add pipeline group resources

Kirdock:
#### **Why** this PR?
To support the upcoming OpenPipeline pipeline-groups schema

#### **What** has changed?
New pipeline-groups resources added for each kind

#### **How** does it do it?
Generating via schema

#### How is it **tested**?
E2E that are part of another branch/PR. https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/983

They can make use of pipeline groups

**Issue:** CA-16606

### Notes
Some OpenPipeline resources are out of date, making them incompatible with pipeline groups. Therefore, preparing an OpenPipeline resource update


dynatrace/api/builtin/openpipeline/bizevents/pipelinegroups/settings/enums.go
```diff
@@ -0,0 +1,56 @@
...
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+ */
+
+package pipelinegroups
+
+type StageConfigType string
+
+var StageConfigTypes = struct {
```
d0weinberger, line 22: Don't we handle camel case correctly?
Shouldn't we have `IncludeAll` here instead of `Includeall`?
Same question further below with the StageTypes

dynatrace/api/builtin/openpipeline/bizevents/pipelinegroups/settings/pipeline_group_composition.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+type PipelineGroupCompositions []*PipelineGroupComposition
+
+func (me *PipelineGroupCompositions) Schema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"pipeline_group_composition": {
+			Type:        schema.TypeList,
+			Required:    true,
+			MinItems:    1,
```
d0weinberger, line 34: We could have `MaxItems: 10` here.
Might be worth an improvement story?

dynatrace/api/builtin/openpipeline/bizevents/pipelinegroups/settings/stage_config.go
```diff
@@ -0,0 +1,73 @@
...
+
+func (me *StageConfig) MarshalHCL(properties hcl.Properties) error {
+	return properties.EncodeAll(map[string]any{
+		"exclude": me.Exclude,
+		"include": me.Include,
+		"type":    me.Type,
+	})
+}
+
+func (me *StageConfig) HandlePreconditions() error {
```
d0weinberger, line 61: Shouldn't there be preconditions checking that `exclude` is only set when `type` == `exclude` and `include` is only set when `type` == `include`?

dynatrace/api/builtin/openpipeline/bizevents/pipelinegroups/settings/stage_config.go
```diff
@@ -0,0 +1,73 @@
...
+type StageConfig struct {
+	Exclude []StageType     `json:"exclude,omitempty"` // exclude stages. Possible values: `costAllocation`, `dataExtraction`, `davis`, `metricExtraction`, `processing`, `productAllocation`, `securityContext`, `smartscapeEdgeExtraction`, `smartscapeNodeExtraction`, `storage`
+	Include []StageType     `json:"include,omitempty"` // include stages. Possible values: `costAllocation`, `dataExtraction`, `davis`, `metricExtraction`, `processing`, `productAllocation`, `securityContext`, `smartscapeEdgeExtraction`, `smartscapeNodeExtraction`, `storage`
+	Type    StageConfigType `json:"type"`              // Stage configuration type. Possible values: `exclude`, `include`, `includeAll`
+}
+
+func (me *StageConfig) Schema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"exclude": {
+			Type:        schema.TypeSet,
```
d0weinberger, line 34: These are sets, so we need special handling because of the bug in the SDK, right?

dynatrace/api/builtin/openpipeline/bizevents/pipelinegroups/settings/settings.go
```diff
@@ -0,0 +1,80 @@
...
+			MinItems:    1,
+			MaxItems:    1,
+		},
+		"display_name": {
+			Type:        schema.TypeString,
+			Description: "Display name",
+			Required:    true,
+		},
+		"member_pipelines": {
+			Type:        schema.TypeSet,
```
d0weinberger, line 48: Set. Special handling?

dynatrace/api/builtin/openpipeline/bizevents/pipelinegroups/settings/stage_config.go
Kirdock: We haven't considered non-basic basic types for the preconditions so far and as the include and exclude are of type set, we didn't consider it. But would be worth looking into that in a follow-up

dynatrace/api/builtin/openpipeline/bizevents/pipelinegroups/settings/enums.go
Kirdock: In this case I would say it doesn't really matter, as we don't use the enums.go file and the actual enum values are correct, right?

d0weinberger: True, it's more nit-picking in this case.

dynatrace/api/builtin/openpipeline/bizevents/pipelinegroups/settings/pipeline_group_composition.go
Kirdock: We haven't considered maxItems so far for any resource, but might be a thing to improve, yes

dynatrace/api/builtin/openpipeline/bizevents/pipelinegroups/settings/stage_config.go
Kirdock: No, these are basic types items and basic type aren't affected by the emtpy-set bug 👍

dynatrace/api/builtin/openpipeline/bizevents/pipelinegroups/settings/settings.go
Kirdock: No, these are basic types items and basic type aren't affected by the emtpy-set bug 👍

d0weinberger (dismissed):

dynatrace/api/builtin/openpipeline/bizevents/pipelinegroups/settings/pipeline_group_composition.go
Kirdock: Follow-up: CA-18164

dynatrace/api/builtin/openpipeline/bizevents/pipelinegroups/settings/stage_config.go
Kirdock: Follow-up: CA-18165

arthurpitman (approved):

PR #982, opened 2026-02-11
feat: update dynatrace_openpipeline_v2_* resources to 1.51

Kirdock:
#### **Why** this PR?
Some resources need to be updated.
Some were still on `1.34.1`, where they aren't compatible with pipeline groups

#### **What** has changed?
Resources are updated to match the current payload.

#### **How** does it do it?
- New properties for `field_extraction_entry` added
- New required property `extraction_type`. Fallback to the set default value added.
- New preconditions added for `field_extraction_entry` (due to the extraction_type and related conditional properties)
- Typo fixed for all resources (`Possible Values:...` => `Possible values:...`)
- Empty check for the empty-set-bug improved to be more future proof (if more attributes are added)

For the older resources that were still on `1.34.1`
- SDLC and bizevent attributes now require the `event_type` field to be set

#### How is it **tested**?
Existing tests for the empty-set-bug updated

They can now configure the resource to work with pipeline groups

**Issue:** CA-18100

### DIFF
--- a/dynatrace/api/builtin/openpipeline/bizevents/ingestsources/settings/field_extraction_entry.go
+++ b/dynatrace/api/builtin/openpipeline/bizevents/ingestsources/settings/field_extraction_entry.go
@@ -18,6 +18,9 @@
 package ingestsources
 
 import (
+	"fmt"
+
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/opt"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/terraform/hcl"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
@@ -47,8 +50,9 @@ func (me *FieldExtractionEntries) UnmarshalHCL(decoder hcl.Decoder) error {
 	// https://github.com/hashicorp/terraform-plugin-sdk/issues/895
 	// Only known workaround is to ignore these blocks
 	newEntries := FieldExtractionEntries{}
+	var emptyItem FieldExtractionEntry
 	for _, entry := range *me {
-		if entry.SourceFieldName != "" || entry.DestinationFieldName != nil || entry.DefaultValue != nil {
+		if *entry != emptyItem {
 			newEntries = append(newEntries, entry)
 		}
 	}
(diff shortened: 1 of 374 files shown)


dynatrace/api/builtin/openpipeline/bizevents/ingestsources/settings/field_extraction_entry.go
```diff
@@ -57,43 +61,87 @@ func (me *FieldExtractionEntries) UnmarshalHCL(decoder hcl.Decoder) error {
...
 		"default_value":          me.DefaultValue,
 		"destination_field_name": me.DestinationFieldName,
+		"extraction_type":        me.ExtractionType,
 		"source_field_name":      me.SourceFieldName,
 	})
 }
 
+func (me *FieldExtractionEntry) HandlePreconditions() error {
+	if (me.ConstantFieldName == nil) && (string(me.ExtractionType) == "constant") {
+		me.ConstantFieldName = opt.NewString("")
```
d0weinberger, line 121: I don't really understand this in the following 2 ifs. 
Why do we set nil fields to empty strings, and only given specific extraction types?

dynatrace/api/builtin/openpipeline/bizevents/ingestsources/settings/field_extraction_entry.go
```diff
@@ -57,43 +61,87 @@ func (me *FieldExtractionEntries) UnmarshalHCL(decoder hcl.Decoder) error {
...
+	}
+	if (me.ConstantValue == nil) && (string(me.ExtractionType) == "constant") {
+		me.ConstantValue = opt.NewString("")
+	}
+	if (me.SourceFieldName == nil) && (string(me.ExtractionType) == "field") {
+		me.SourceFieldName = opt.NewString("")
+	}
+	if (me.DefaultValue != nil) && (string(me.ExtractionType) != "field") {
+		return fmt.Errorf("'default_value' must not be specified if 'extraction_type' is set to '%v'", me.ExtractionType)
```
d0weinberger, line 130: Would the correct error message not be more like the following?
`"'default_value' must not be specified if 'extraction_type' is not set to 'field'"`

dynatrace/api/builtin/openpipeline/bizevents/ingestsources/settings/field_extraction_entry.go
```diff
@@ -57,43 +61,87 @@ func (me *FieldExtractionEntries) UnmarshalHCL(decoder hcl.Decoder) error {
...
+	}
+	if (me.DestinationFieldName != nil) && (string(me.ExtractionType) != "field") {
+		return fmt.Errorf("'destination_field_name' must not be specified if 'extraction_type' is set to '%v'", me.ExtractionType)
```
d0weinberger, line 133: same here

dynatrace/api/builtin/openpipeline/bizevents/pipelines/settings/field_extraction_entry.go
```diff
@@ -57,43 +61,87 @@ func (me *FieldExtractionEntries) UnmarshalHCL(decoder hcl.Decoder) error {
...
+		"constant_value":         me.ConstantValue,
 		"default_value":          me.DefaultValue,
 		"destination_field_name": me.DestinationFieldName,
+		"extraction_type":        me.ExtractionType,
 		"source_field_name":      me.SourceFieldName,
 	})
 }
 
+func (me *FieldExtractionEntry) HandlePreconditions() error {
+	if (me.ConstantFieldName == nil) && (string(me.ExtractionType) == "constant") {
```
d0weinberger, line 120: same questions here

dynatrace/api/builtin/openpipeline/bizevents/pipelines/settings/settings.go
```diff
@@ -78,7 +78,7 @@ func (me *Settings) Schema() map[string]*schema.Schema {
 		},
 		"group_role": {
```
d0weinberger, line 79: There is a modification policy set to `NEVER` for `group_role`, right? Shouldn't this entail `ForceNew: true`? Same for `routing` and actually `customId` as well. Or does this come in a follow-up? Because you mentioned that you are going to handle that as well

dynatrace/api/builtin/openpipeline/davis/events/ingestsources/settings/bizevent_attributes.go
```diff
@@ -41,7 +41,7 @@ func (me *BizeventAttributes) Schema() map[string]*schema.Schema {
 		"event_type": {
 			Type:        schema.TypeList,
 			Description: "Event type",
-			Optional:    true, // nullable
+			Required:    true,
```
d0weinberger, line 44: Breaking change here

Kirdock: Yes, this is intended and mentioned in the PR description. We can't set a valid default value here; therefore the breaking change.

dynatrace/api/builtin/openpipeline/bizevents/pipelines/settings/settings.go
Kirdock: Follow-up 👍

dynatrace/api/builtin/openpipeline/bizevents/ingestsources/settings/field_extraction_entry.go
Kirdock: If the type is set to `constant` then `ConstantFieldName` must be set. In our case, we always use the default empty value for all resources where this is the case, as this is not null.
So far, we haven't considered the `NOT_BLANK` constraint for all resources we have.

dynatrace/api/builtin/openpipeline/bizevents/ingestsources/settings/field_extraction_entry.go
Kirdock: I see both as valid. For the current structure, you just say that the field must not be specified if a certain type is set. In the current approach, you don't really care to which type it belongs (it may be several), but more that it doesn't belong to the current set type.

Because a field may belong to several set constants, the error message could expand like this
`'default_value' must not be specified if 'extraction_type' is not set to 'field' or 'fieldSet' or...`

d0weinberger (approved):

PR #985, opened 2026-02-16
chore: fix `dynatrace_disk_edge_anomaly_detectors` documentation

Kirdock:
#### **Why** this PR?
The description of some fields escaped their indentation, which incorrectly showed them as other required/optional fields instead of being a sub-bullet point list of the current field.
image

#### **What** has changed?
The documentation of a field doesn't escape its bullet point anymore.
image

#### **How** does it do it?
By adding some blanks for indentation if there is a bullet point list inside the description.

#### How is it **tested**?
Manual tests

They will see the updated documentation.

**Issue:** CA-17707


d0weinberger:
How did you check the result?

Kirdock:
> How did you check the result?

I generated the md file and pasted it into the doc preview tool https://registry.terraform.io/tools/doc-preview

d0weinberger (approved):

PR #986, opened 2026-02-17
test: add integration setup step

Kirdock:
This adds a setup step to the pipeline before the integration tests are executed, ensuring that long-running resources are created before the tests run.

#### **Why** this PR?
Some E2E tests (e.g., examples) refer to existing objects (e.g., management zones, request attributes, process groups, etc.). As this can easily break if someone just deletes something in the environment, we need to remove hardcoded IDs and create the needed dependencies on demand.
Because some resources take a long time to create, it's not feasible to create them in each example/test.
Therefore ,we need a setup that creates them once, before the actual tests run.

#### **What** has changed?
Required resources are created before the integration tests start.

#### **How** does it do it?
By using our the latest published provider to create the needed resources on demand.
If they don't exist, they will be created and if they exist, the step will be ignored.
The current implementation doesn't need a state file. It's either "import" or "create".

#### How is it **tested**?
Manual tests and verified the pipeline

N/A

**Issue:** CA-18113

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -65,9 +65,33 @@ jobs:
           name: lint-report
           path: lint-report.xml
 
+  run-setup:
+    name: Run Integration Setup
+    runs-on: ubuntu-latest
+    steps:
+      - name: ⬇️ Check out code into the Go module directory
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd #v6.0.2
+
+      - name: 🛠️ Setup Terraform
+        uses: hashicorp/setup-terraform@5e8dbf3c6d9deaf4193ca7a8fb23f2ac83bb6c85 #v4.0.0
+        with:
+          terraform_wrapper: false
+
+      - name: 🚀 Integration Setup - Terraform Init
+        working-directory: ./testsetup
+        run: terraform init
+
+      - name: 🚀 Integration Setup - Terraform Apply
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -65,9 +65,33 @@ jobs:
...
 
+  run-setup:
+    name: Run Integration Setup
+    runs-on: ubuntu-latest
+    steps:
+      - name: ⬇️ Check out code into the Go module directory
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd #v6.0.2
+
+      - name: 🛠️ Setup Terraform
+        uses: hashicorp/setup-terraform@b9cd54a3c349d3f38e8881555d616ced269862dd
```
d0weinberger: could you add a version comment, like in line 73?

d0weinberger:
> if they exist, the step will be ignored

How is this achieved in the PR? I don't see it

Kirdock:
> > if they exist, the step will be ignored
> 
> How is this achieved in the PR? I don't see it

It's in the Terraform file. With the combination of `data`, `import`, and `resource`, we have either an import (ignore) or a resource creation

.github/workflows/tests.yml
Kirdock: Done. I also added it to our other usage

tomazpu (approved):

PR #989, opened 2026-02-23
fix(rum): unmarshalling of user_action_properties

d0weinberger:
#### **Why** this PR?
The `user_action_properties` field of `dynatrace_calculated_web_metric` could not be unmarshalled.
The following error log occurred: 
`[WARN] **web.UserActionProperties user_action_properties NOT covered`

- `user_action_properties` changed from `TypeSet` to `TypeList`
- Receiver for `MarshalHCL` of `UserActionProperties` changed to pointer receiver
- `UserActionFilter` now has `UserActionProperties` instead of `*UserActionProperties`

#### How is it **tested**?
Manual testing and E2E test was added.

Users can use the `user_action_properties` field of the `dynatrace_calculated_web_metric` now.

**Issue:** CA-18231

### DIFF
--- a/dynatrace/api/v1/config/metrics/calculated/web/settings/user_action_filter.go
+++ b/dynatrace/api/v1/config/metrics/calculated/web/settings/user_action_filter.go
@@ -24,43 +24,43 @@ import (
 
 // User action filter of a calculated web metric.
 type UserActionFilter struct {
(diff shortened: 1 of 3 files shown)


Kirdock:
Seems like the test/example panics
```
Invalid address to set: []string{"user_action_filter", "0", "user_action_properties", "0", "from"}
```

Could you also take care of the TypeSet bug for the user_action_properties => property or follow-up?

Kirdock (approved):

PR #990, opened 2026-02-23
feat(workflows): Adds `analysis_ready` field to `davis_problem` workflow triggers

d0weinberger:
#### **Why** this PR?
We received feedback that a recently added field for Davis problem triggers for workflows has not yet been made available via the Terraform provider. This PR adds this field.

#### **What** has changed?
The `analysisReady` parameter can now be set via the Terraform provider.

#### **How** does it do it?
The optional boolean field with `false` default value is added to the HCL schema.

#### How is it **tested**?
An E2E test using the field is added.

Users can set the `analysisReady` parameter via the Terraform provider.

**Issue:** CA-18212

### DIFF
--- a/dynatrace/api/automation/workflows/settings/davis_problem_config.go
+++ b/dynatrace/api/automation/workflows/settings/davis_problem_config.go
@@ -31,6 +31,7 @@ type DavisProblemConfig struct {
 	OnProblemClose  bool                    `json:"onProblemClose" default:"false"` //
 	Categories      *DavisProblemCategories `json:"categories"`                     //
 	CustomFilter    string                  `json:"customFilter,omitempty"`         //
+	AnalysisReady   bool                    `json:"analysisReady" default:"false"`  // If set to `true`, the workflow will only be triggered after the initial root cause analysis run is completed
 }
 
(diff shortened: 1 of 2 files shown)


dynatrace/api/automation/workflows/settings/davis_problem_config.go
```diff
@@ -66,6 +67,12 @@ func (me *DavisProblemConfig) Schema(prefix string) map[string]*schema.Schema {
 			Description: "Additional DQL matcher expression to further filter events to match",
 			Optional:    true,
 		},
+		"analysis_ready": {
+			Type:        schema.TypeBool,
+			Description: "If set to `true`, the workflow will only be triggered after the initial root cause analysis run is completed",
```
Kirdock, line 72: This is manually adapted, I guess, or where is this taken from?
It's just because it doesn't match with the one in the spec, `Run only after first root cause analysis run completed`; I'm also fine with this one.

Kirdock (approved):

dynatrace/api/automation/workflows/settings/davis_problem_config.go
d0weinberger: Yes, you are right, I basically reworded the text from the spec a bit.

PR #996, opened 2026-02-24
feat: Support CyberArk in external vault config

arthurpitman:
#### **Why** this PR?
The API supports two further external vault configuration types, `CYBERARK_VAULT_USERNAME_PASSWORD` and `CYBERARK_VAULT_ALLOWED_LOCATION`. 

#### **What** has changed?
Fields required for these `sourceAuthMethod`s are added:
- `username_password_for_cpm` for `CYBERARK_VAULT_USERNAME_PASSWORD`
- `application_id`,  `safe_name`, `folder_name` and `account_name` for `CYBERARK_VAULT_USERNAME_PASSWORD` and `CYBERARK_VAULT_ALLOWED_LOCATION`

#### **How** does it do it?
Based on which fields are present, new values for `sourceAuthMethod` are inferred.

#### How is it **tested**?
Currently, manual testing based on the two examples included in this PR. To automate this, we'd need to have a certificate available in the pipeline and a way to clean up the credentials after the test, which is currently not possible due to the associated synthetic monitor. 

Users will be able to configure syncing credentials from a CyberArk vault.

**Issue:** CA-17530


dynatrace/api/v2/credentials/vault/settings/externalvault/config.go
```diff
@@ -214,8 +283,12 @@ func (me *Config) UnmarshalHCL(decoder hcl.Decoder) error {
 		me.SourceAuthMethod = SourceAuthMethods.AzureKeyVaultClientSecret
 	} else if me.RoleID != nil {
 		me.SourceAuthMethod = SourceAuthMethods.HashicorpVaultAppRole
+	} else if me.UsernamePasswordForCPM != nil {
```
d0weinberger, line 286: Should there be some precondition handling to mutually exclude setting RoleID with CyberArk params or CyberArk username password with other CyberArk params?
The problem is, you can basically set everything, but this logic here which determines which kind of auth method is actually used is not transparent to the user, is it?

arthurpitman: I would say the API should handle that. Problem is, we don't export `sourceAuthMethod`, but infer it.

d0weinberger (dismissed):

Kirdock (approved):

d0weinberger (approved):

PR #997, opened 2026-02-24
feat(workflows): Adds more optional fields to workflows

d0weinberger:
### Why this PR?
The `dynatrace_automation_workflow` resource was missing several fields supported by the Dynatrace Automation API. Users couldn’t configure important workflow properties like deployment state, execution limits, or workflow-level input parameters, forcing them to manage these settings outside of Terraform.

### What has changed?

*New workflow-level attributes* (all Optional, non-breaking):

| HCL Attribute | Type | Default | Description |
|---|---|---:|---|
| `is_deployed` | `bool` | `true` | Controls whether the workflow is active and billed, or a draft |
| `owner_type` | `string` | - | Enum: `USER` / `GROUP` - type of the workflow owner |
| `hourly_execution_limit` | `int` | `1000` | Cap on executions per hour |
| `input` | `string` (JSON) | - | Workflow-level input parameters available to all tasks |
| `task_defaults` | `string` (JSON) | - | Default settings applied to all tasks |
| `guide` | `string` | - | Markdown guide text for the workflow (max 10,000 chars) |
| `result` | `string` | - | Result identifier for the workflow |

### How does it do it?

All changes follow established patterns in this provider:

* Added struct fields to `Workflow` with correct JSON tags.
* Extended Terraform schema in `Schema()` with appropriate types, defaults, and validation (`StringInSlice`, `ValidateMaxLength`, `SuppressJSONorEOT`).
* Implemented `is_deployed` using the `*bool` + `GetOkExists` pattern (same as existing `isPrivate`) to avoid TF SDK default/zero-value pitfalls.
* Implemented `input` and `task_defaults` using the existing JSON-string pattern (like `Task.Input`): `map[string]any` in Go/API, `jsonencode()` in HCL.
* Updated `MarshalJSON` to include all new fields in the custom serialization struct (which injects `schemaVersion`).
* Updated `MarshalHCL` / `UnmarshalHCL` to encode/decode the new fields.

### How is it tested?
I added a new Terraform file for E2E testing using the newly introduced fields.

### How does it affect users?

* **No breaking changes** — all new fields are Optional and defaults align with API behavior; existing configurations keep working.
* **No state migration needed** — additive Optional fields don’t require schema version bumps.
* Users can now manage workflow deployment state, execution limits, input parameters, task defaults, and guides entirely through Terraform.

**Issue:** CA-18196


dynatrace/api/automation/workflows/settings/workflow.go
```diff
@@ -100,13 +121,45 @@ func (me *Workflow) Schema() map[string]*schema.Schema {
...
+			Type:             schema.TypeString,
+			Description:      "Default task settings as JSON. These defaults are applied to all tasks in the workflow",
+			Optional:         true,
+			DiffSuppressFunc: hcl.SuppressJSONorEOT,
+		},
+		"guide": {
+			Type:             schema.TypeString,
+			Description:      "Informational guide text for the workflow",
+			Optional:         true,
+			ValidateDiagFunc: ValidateMaxLength(10000),
```
Kirdock: We agreed not to validate field lengths, since the API might change, right?
Could you remove it?

dynatrace/api/automation/workflows/settings/workflow.go
```diff
@@ -72,12 +80,25 @@ func (me *Workflow) Schema() map[string]*schema.Schema {
 			DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool { return new == "" },
 			ValidateDiagFunc: ValidateUUID,
 		},
+		"owner_type": {
+			Type:             schema.TypeString,
+			Description:      "The type of the owner. Possible values are `USER` and `GROUP`",
+			Optional:         true,
+			DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool { return new == "" },
```
Kirdock: What's the reason for the `DiffSupressFunc` here? Is there a different behavior (sent data vs received data)?

dynatrace/api/automation/workflows/testdata/terraform/example-new-fields.tf
```diff
@@ -0,0 +1,192 @@
...
+  is_deployed            = false
+  owner_type             = "USER"
+  input = jsonencode({
+    "environment" : "production",
+    "threshold" : 42
+  })
+  task_defaults = jsonencode({
+    "timeout" : 3600
+  })
+  guide = chomp(<<-EOT
```
Kirdock: Do we really want to keep this huge markdown?
I don't see a real benefit in having it, and it would just reduce the readability and also maintainability of the example files. For testing, a short string would be enough

dynatrace/api/automation/workflows/testdata/terraform/example-new-fields.tf
Kirdock, line 1: Really nit, but is `new_fields` a good pick for a name? If new properties come in, will there be a `new-fields2.tf`? 😅

dynatrace/api/automation/workflows/settings/workflow.go
```diff
@@ -100,13 +121,45 @@ func (me *Workflow) Schema() map[string]*schema.Schema {
...
 	}
 }
 
 func (me *Workflow) MarshalHCL(properties hcl.Properties) error {
 	if me.Private == nil {
 		me.Private = opt.NewBool(true)
 	}
+	if me.IsDeployed == nil {
+		me.IsDeployed = opt.NewBool(true)
+	}
```
Kirdock: So, the value on the API side can still be undefined, even if it has a default value?
When running locally, without `is_deployed`, the value is always defined for me

dynatrace/api/automation/workflows/settings/workflow.go
```diff
@@ -167,34 +265,59 @@ func (me *Workflow) UnmarshalHCL(decoder hcl.Decoder) error {
...
 		me.Private = opt.NewBool(isPrivate.(bool))
 	}
+
+	// Same pattern as `private` — `isDeployed` defaults to `true`
+	isDeployed, isDeployedExists := decoder.GetOkExists("is_deployed")
+	if !isDeployedExists {
+		me.IsDeployed = opt.NewBool(true)
+	} else {
+		me.IsDeployed = opt.NewBool(isDeployed.(bool))
+	}
```
Kirdock: Because of `Default: true`, the `is_deployed` should always be set, right?

Here I haven't set the `is_deployed`
Image

dynatrace/api/automation/workflows/settings/workflow.go
Kirdock: *Some findings regarding this behavior and a fix here https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/1012,

Kirdock (approved):

PR #998, opened 2026-02-24
feat: resource updates for v332

Kirdock:
#### **Why** this PR?
Some APIs have changed with Dynatrace version 1.332. Therefore, the Terraform resources also need updating.

#### **What** has changed?
Updated the descriptions and ForceNew fields of several resources.

#### **How** does it do it?
Updated/aligned the description (indentation and updates) for
- `dynatrace_ms365_email_connection`
- `dynatrace_declarative_grouping`
- `dynatrace_oneagent_default_mode`
- `dynatrace_process_availability`
- `dynatrace_vmware`

Changes:
- `dynatrace_opentelemetry_metrics`: New optional field `enable_mint_v_2_ingest`
- `dynatrace_devobs_data_masking`: New conditionally required field `comparison_type`. Added a workaround to be backwards compatible
- `dynatrace_vmware`: correctly set ForceNew for non-modifiable fields.
- `dynatrace_azure_connection`: The name can now be modified without recreating the resource. Correctly set ForceNew fields
- `dynatrace_aws_connection`: The name can now be modified without recreating the resource. Correctly set ForceNew fields

#### How is it **tested**?
For new properties, there are new examples.
Else, it's mostly about descriptions.

**Issue:** CA-18256


dynatrace/api/builtin/declarativegrouping/settings/enums.go
```diff
@@ -24,9 +24,9 @@ var ProcessItems = struct {
 	Executable     ProcessItem
 	Executablepath ProcessItem
 }{
-	"CommandLine",
-	"Executable",
-	"ExecutablePath",
+	"commandLine",
```
tomazpu, line 27: This is a breaking change, for existing configurations or?

tomazpu (approved):

dynatrace/api/builtin/declarativegrouping/settings/enums.go
Kirdock: No.
- The enums were wrong in the first place
- They weren't even used (this file is not used. This is just for visibility)
- They now represent the API. The API didn't change

PR #1017, opened 2026-03-10
fix: additional new line during export

Kirdock:
#### **Why** this PR?
Fields that ended with a new line via `<<-EOT` had an additional new line added, leading to a non-empty terraform plan for the exported resources. 
e.g., 
When this is a resource that is applied
```hcl
resource "my_res" "name" {
  prop = <<-EOT
    something here
  EOT
}
```
The export of that resource looked like this
```hcl
resource "my_res" "name" {
  prop = <<-EOT
    something here
  
  EOT
}
```

#### **What** has changed?
The additional new line is removed.

#### **How** does it do it?
The new line that is required for closing the EOT syntax is only added if there isn't already one.

#### How is it **tested**?
Manual tests, using the workflows resource

The correct value is now set during export if a string contained new lines at the end.

**Issue:** CA-18459

### DIFF
--- a/terraform/hclgen/primitive_entry.go
+++ b/terraform/hclgen/primitive_entry.go
@@ -109,7 +109,13 @@ func (me *primitiveEntry) Write(w *hclwrite.Body, indent string) error {
 	} else if strValP, ok := me.Value.(*string); ok && strValP != nil && isJSON(*strValP) {
 		w.SetAttributeRaw(me.Key, hclwrite.Tokens{&hclwrite.Token{Type: hclsyntax.TokenStringLit, Bytes: []byte(toJSONencode(*strValP, indent))}})
 	} else if strVal, ok := me.Value.(string); ok && !preventHeredoc && strings.Contains(strVal, "\n") {
-		mlstr := "<<-EOT\n" + indent + "  " + finalizeString(strVal, indent) + "\n" + indent + "EOT"
+		var eotIndent string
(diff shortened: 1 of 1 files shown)


d0weinberger (approved):

terraform/hclgen/primitive_entry.go
```diff
@@ -109,7 +109,13 @@ func (me *primitiveEntry) Write(w *hclwrite.Body, indent string) error {
 	} else if strValP, ok := me.Value.(*string); ok && strValP != nil && isJSON(*strValP) {
 		w.SetAttributeRaw(me.Key, hclwrite.Tokens{&hclwrite.Token{Type: hclsyntax.TokenStringLit, Bytes: []byte(toJSONencode(*strValP, indent))}})
 	} else if strVal, ok := me.Value.(string); ok && !preventHeredoc && strings.Contains(strVal, "\n") {
-		mlstr := "<<-EOT\n" + indent + "  " + finalizeString(strVal, indent) + "\n" + indent + "EOT"
+		var eotIndent string
+		// If we always add a newline before the EOT end, then we end up with an extra blank line at the end of the string if the string already ends with a newline.
+		// This would cause a non-empty terraform plan
+		if !strings.HasSuffix(strVal, "\n") {
+			eotIndent = "\n" + indent
```
arthurpitman, line 116: So, if we already have a new line at the end of the value, we also skip the indent? Is that what we want?

Kirdock: We could also remove it 🤔 
Here with and without indentation:

Image

Image

PR #1025, opened 2026-03-17
feat: Resource `dynatrace_direct_shares` uses client from core library

arthurpitman:
#### **Why** this PR?
We'd like to use the same REST client as other platform resources, which is in the core library. As a consequence, a client for direct shares was added to the core library.

This PR updates the service for the `dynatrace_direct_shares` resource to use this client.

- First, the core library is updated to a version including the direct shares client.
- Then the logic of service is updated to use this client.
- Then error handling is simplified to return `APIError`s directly.
- The `monaco` folder is removed as the code is no-longer used.

#### How is it **tested**?
- The existing test is updated. The logic of the resource remains the same.
- An additional test is added which creates and then updates a direct share.

- The `dynatrace_direct_shares` resource can now use a platform token or an OAuth client.
- The user can now export `dynatrace_direct_shares` resources.
- `dynatrace_direct_shares` resources can now be created with no recipients, something supported by the API.

**Issues:** CA-18450, CA-18424


dynatrace/api/documents/directshares/service.go
```diff
@@ -70,41 +94,118 @@ func (me *service) SchemaID() string {
...
 func (me *service) Delete(ctx context.Context, id string) error {
-	return me.client(ctx).DELETE(ctx, directshare.DirectShares, id)
+	client, err := me.client(ctx)
+	if err != nil {
+		return err
+	}
+	if err := client.Delete(ctx, id); err != nil {
+		return err
+	}
+	return nil
```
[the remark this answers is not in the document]
arthurpitman: Will be fixed by improving the generic resource Delete handler

dynatrace/api/documents/directshares/service.go
```diff
@@ -22,43 +22,67 @@ import (
...
+func (me *service) Get(ctx context.Context, id string, v *serviceSettings.DirectShare) (err error) {
+	client, err := me.client(ctx)
+	if err != nil {
+		return err
+	}
+
+	result, err := client.Get(ctx, id)
+	if err != nil {
 		return err
 	}
```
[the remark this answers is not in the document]
arthurpitman: Will be fixed by improving the generic resource Delete handler

dynatrace/api/documents/directshares/service.go
```diff
@@ -22,43 +22,67 @@ import (
...
+		return err
+	}
+
+	for _, obj := range recipientsResp.All() {
+		var rec serviceSettings.Recipient
+		if err := json.Unmarshal(obj, &rec); err != nil {
+			return err
+		}
+		v.Recipients = append(v.Recipients, &rec)
+	}
```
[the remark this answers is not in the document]
arthurpitman: Fixed

dynatrace/api/documents/directshares/service.go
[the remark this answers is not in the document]
Kirdock: https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/1027

dynatrace/api/documents/directshares/service.go
```diff
@@ -22,43 +22,69 @@ import (
...
 		return err
 	}
 
-	v.DocumentId = result.DocumentId
-	v.ID = result.ID
-	v.Access = strings.Join(result.Access, "-")
-	v.Recipients = result.Recipients
+	var dsResponse struct {
+		ID         string   `json:"id"`
+		DocumentId string   `json:"documentId"`
```
Kirdock: Is this actually the correct JSON tag? I'm just looking at `DirectShare` of the setting and here it's `document_id`

dynatrace/api/documents/directshares/service.go
```diff
@@ -70,41 +96,118 @@ func (me *service) SchemaID() string {
...
+	}
+	if err := json.Unmarshal(result.Data, &resp); err != nil {
 		return nil, err
 	}
-	return &api.Stub{ID: id}, nil
+	return &api.Stub{ID: resp.ID}, nil
 }
 
-func (me *service) Update(ctx context.Context, id string, v *directshares.DirectShare) (err error) {
+func (me *service) Update(ctx context.Context, id string, v *serviceSettings.DirectShare) (err error) {
```
Kirdock, line 213: What happens if the `document_id` changes? Should this one actually have a ForceNew?

dynatrace/api/documents/directshares/service.go
```diff
@@ -70,41 +96,118 @@ func (me *service) SchemaID() string {
...
 
-	return me.client(ctx).UPDATE(ctx, directshare.DirectShares, id, data)
+	var remoteRecipients serviceSettings.Recipients
+	for _, obj := range recipientsResp.All() {
+		var rec serviceSettings.Recipient
+		if err := json.Unmarshal(obj, &rec); err != nil {
+			return err
+		}
+		remoteRecipients = append(remoteRecipients, &rec)
+	}
```
Kirdock: [nit]: Should we move this into a separate function like `UnmarshalRecipients` as this snippet is used two times?

dynatrace/api/documents/directshares/service.go
Kirdock: Edit: I guess this doesn't matter that it's `document_id` as we are overriding the MarshalJSON

arthurpitman: Fixed

Kirdock (dismissed):

dynatrace/api/documents/directshares/service.go
```diff
@@ -22,89 +22,279 @@ import (
...
 func (me *service) Delete(ctx context.Context, id string) error {
-	return me.client(ctx).DELETE(ctx, directshare.DirectShares, id)
+	client, err := me.clientGetter(ctx, me.credentials)
+	if err != nil {
+		return err
+	}
+	if err := client.Delete(ctx, id); err != nil {
+		return err
+	}
+	return nil
```
tomazpu:
```suggestion
	return client.Delete(ctx, id)
```

tomazpu (dismissed):
Just one nit-simplification

templates/resources/direct_shares.md.tmpl
```diff
@@ -12,8 +12,6 @@ description: |-
 
 -> To utilize this resource, please define the environment variables `DT_CLIENT_ID`, `DT_CLIENT_SECRET`, `DT_ACCOUNT_ID` with an OAuth client including the following permissions: **Read direct-shares** (`document:direct-shares:read`), **Write direct-shares** (`document:direct-shares:write`), and **Delete direct-shares** (`document:direct-shares:delete`).
 
--> This resource is currently not covered by the export utility.
```
Kirdock, line 15: Now that we've removed that line, we should add an export section here, as with other resources, right?

arthurpitman: I added the note that it is not exported directly (fixed to code to make it so too)  and also added an example for exporting.

Kirdock (approved):

PR #1026, opened 2026-03-17
fix: correctly handle errors in automation resources

Kirdock:
#### **Why** this PR?
The error handling was wrong, so that certain cases were simply dead code.
Delete also produced an error if the resource didn't exist anymore

#### **What** has changed?
- 404 during delete is correctly handled
- Not needed status checks removed

#### **How** does it do it?
The error handling was wrong, so that certain cases were simply dead code.
- Removed status checks that weren't possible because they errored out before
- Correctly check status codes during delete

#### How is it **tested**?
Manual tests

Delete won't produce an error on 404 anymore

**Issue:** CA-18640


dynatrace/api/automation/scheduling_rules/service.go
```diff
@@ -141,33 +126,19 @@ func (me *service) Update(ctx context.Context, id string, v *scheduling_rules.Se
...
 }
 
 func (me *service) Delete(ctx context.Context, id string) error {
 	client, err := me.client(ctx)
 	if err != nil {
 		return err
 	}
-	response, err := client.Delete(ctx, automation.SchedulingRules, id)
-	if response.StatusCode == 204 || response.StatusCode == 404 {
-		return nil
```
d0weinberger, line 165: Why did 404 produce an error when it was checked here?

EDIT: Ah, because the `Delete` does not give a response in the error case, right? Then, actually, the core client fix would be enough, I would suspect?

Kirdock: The response is not set if the status code is not `>=200` and `<400`

arthurpitman (approved):

PR #1027, opened 2026-03-18
fix: not-found behavior for core clients

Kirdock:
#### **Why** this PR?
For core clients such as automation or segments, 404 errors during get and delete are treated as errors. This is not the Terraform way and should be changed so that Terraform correctly sees it as `needs creation` or `ignore 404 on delete`.

#### **What** has changed?
Core 404 errors are not correctly handled. Get shows "needs creation" and delete ignores 404

#### **How** does it do it?
By not just asserting on our own Error struct but also on the apiError of the core clients.

#### How is it **tested**?
Manual tests.

Better error handling for 404. Terraform plan will show the correct plan and not error out if a resource was deleted via the UI.

**Issue:** Follow-up based on CA-18640

### Out of scope
There are more places, where just `Error` is checked and not `apiError`. This PR fixes the GET/DELETE behavior.
Others are not that important for now and would require more refactoring work

### DIFF
--- a/dynatrace/rest/error.go
+++ b/dynatrace/rest/error.go
@@ -25,6 +25,8 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+
+	api2 "github.com/dynatrace/dynatrace-configuration-as-code-core/api"
 )
 
 type errorEnvelope struct {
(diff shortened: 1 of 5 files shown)


dynatrace/rest/error.go
```diff
@@ -47,6 +49,13 @@ type Error struct {
 	Method               string                `json:"-"`
 }
 
+// IsNotFoundError checks if an error is a 404 error
+// api_token_client client, hybrid_client, etc. are using Error while core clients like automation use api2.APIError
+func IsNotFoundError(err error) bool {
+	var restErr Error
+	return errors.As(err, &restErr) && restErr.Code == 404 || api2.IsNotFoundError(err)
```
d0weinberger: The PR makes sense to me, but I am hesitant to approve without automated tests, as this is such a central component in the provider.

Kirdock: I'll add one

Kirdock: added for Read, Delete and IsNotFoundError

d0weinberger (approved):

PR #1030, opened 2026-03-19
chore: generate settings schema support file

Kirdock:
#### **Why** this PR?
To have a summary of currently supported settings schemas.

#### **What** has changed?
- `supported-resources.md` is validated based on the existing resources.
- A `schemaSupport.json` file is attached to the release assets

#### **How** does it do it?
- Generates a JSON file containing the current supported schemas and resources
- Errors if `supported-resources.md` doesn't mention a schema
- Adds it as a check to the pipeline that every schema is defined
- Adds it to the release assets

#### How is it **tested**?
Unit tests

They can look up the supported schema version for each upcoming release

**Issue:** CA-18558


tools/schemasupport/json.go
```diff
@@ -0,0 +1,84 @@
...
+// formatMissingError returns an error message listing all schemaIds that have
+// no matching resource name.
+func formatMissingError(missing []string) string {
+	sort.Strings(missing)
+	var sb strings.Builder
+	sb.WriteString("the following schemaIds have no matching resource name in supported-resources.md:\n")
+	for _, id := range missing {
+		sb.WriteString("  - ")
+		sb.WriteString(id)
+		sb.WriteString("\n")
```
arthurpitman:
```suggestion
	sb.WriteString("the following schemaIds have no matching resource name in supported-resources.md:")
	for _, id := range missing {
	
		sb.WriteString("\n  - ")
		sb.WriteString(id)
```

I'd then add an extra new line if required in the format string where this is used (in `main.go`)

tools/schemasupport/json_test.go
```diff
@@ -0,0 +1,221 @@
...
+		{SchemaID: "schema-b", Version: "2.0"},
+	}
+	mapping := map[string][]string{
+		"schema-a": {"res_a"},
+		"schema-b": {"res_b1", "res_b2"},
+	}
+
+	rows, missing := expandRows(infos, mapping)
+	assert.Empty(t, missing)
+	assert.Len(t, rows, 3) // schema-a → 1 row, schema-b → 2 rows
```
arthurpitman: nit: this test could be tightened to assert on the rows, maybe also combined with `TestExpandRows_RowContents` below?

arthurpitman (approved):

PR #1031, opened 2026-03-20
fix(webapplication): `monitoring_data_path` is set to Computed

d0weinberger:
#### **Why** this PR?
This is done because if not set, it can be that the API returns a default value for it. If we did not set it to `Computed`, users would get a non-empty plan on every re-apply.

`monitoring_data_path` is set to Computed

#### How is it **tested**?
A test file was added with an application where `monitoring_data_path` is not set in the Terraform file but will be set by the API. It must not produce a non-empty plan on reapply.

Users that experienced non-empty plans on reapply should not experience this anymore.

**Issue:** CA-18699

### DIFF
--- a/dynatrace/api/v1/config/applications/web/testdata/terraform/example_b.tf
+++ b/dynatrace/api/v1/config/applications/web/testdata/terraform/example_b.tf
@@ -1,0 +1,86 @@
+# Example with monitoring_data_path set by API
+resource "dynatrace_web_application" "#name#" {
+  cost_control_user_session_percentage = 100
+  load_action_key_performance_metric   = "VISUALLY_COMPLETE"
+
+  name                              = "#name#"
+  type                              = "MANUALLY_INJECTED"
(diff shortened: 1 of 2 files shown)


Kirdock (approved):

dynatrace/api/v1/config/applications/web/testdata/terraform/example_b.tf
Kirdock, line 1: Do we actually need this test? `monitoring_data_path` is also not set in example-a or does it depend on another property?

d0weinberger: If in example-a `monitoring_data_path` would be set by the API, we would have had failing tests due to non-empty plans earlier, right? I am not yet sure, what determines whether the property is set by the API, but example-b causes it to be set.

EDIT: Yes, it depends on the type of the web application. `MANUALLY_INJECTED` --> The API will return a value for the `monitoring_data_path`.

PR #1032, opened 2026-03-20
refactor: Don't transform API errors to REST errors in services using core clients

arthurpitman:
#### **Why** this PR?
With changes introduced in #1027, the provider also correctly handles not-found errors represented as the core library's `APIError` type. Therefore, there is no advantage in transforming these errors into `rest.Error` errors. This change was done for direct shares in #1025 and in most places in automation resources in #1026. This PR handles the other resources that use core clients, and updates the list functions for thje automation resources.

The code transforming these errors is removed in:
- `dynatrace/api/documents/document/service.go`
- `dynatrace/api/grail/segments/service.go`
- `dynatrace/api/slo/service.go`

Furthermore, the list functions are updated in:
- `dynatrace/api/automation/workflows/service.go`
- `dynatrace/api/automation/scheduling_rules/service.go`
- `dynatrace/api/automation/business_calendars/service.go`

#### How is it **tested**?
Existing tests.

No effect.

**Issue:** NA

### DIFF
--- a/dynatrace/api/automation/business_calendars/service.go
+++ b/dynatrace/api/automation/business_calendars/service.go
@@ -20,7 +20,6 @@ package business_calendars
 import (
 	"context"
 	"encoding/json"
-	"errors"
 
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api"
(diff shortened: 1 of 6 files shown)


Kirdock:
There are actually some places left, where we create a `rest.Error` and I also just noticed that we also still have them in place for the List function of the automation resources 🤔.

Kirdock (approved):

PR #1039, opened 2026-03-30
fix(dynatrace_automation_workflow): ignore computed task positions from API responses

Kirdock:
#### **Why** this PR?
The position of a task is optional and the API returns a position. Therefore, the field is actually computed.
Applying a config without a position for a task, leads to a non-empty plan afterwards.

#### **What** has changed?
The position from the API is now ignored if it's not set in the state.

#### **How** does it do it?
After unmarshalling workflows in Get, it removes task positions that the API supplies by default, but the user didn't set in the state. This avoids TypeSet hash churn and spurious Terraform plan differences by preserving positions only when present in the state.

The reason here is simply that the TypeSet hash changes every time if a field of an item is set to computed, even if nothing has changed (current state with `position {x=...y = ..}`. => `known after apply`).
If the position is set by the user in the Terraform file, then it's correctly compared to the API set one.
DiffSuppress can't help here because even if the task doesn't change, the bug of TypeSet happens, which adds a new empty item, which leads to a non-empty plan.

#### How is it **tested**?
Nest test added that has a task without positions set. Applying this one should not lead to a non-empty plan anymore.

A task without `position` won't lead to a non-empty plan anymore.

### Limitation:
Minor: If the user updates a task and adds the position field with the actual values set by the API, this will lead to an update instead of an empty plan.

**Issue:** CA-19023

### DIFF
--- a/dynatrace/api/automation/workflows/service.go
+++ b/dynatrace/api/automation/workflows/service.go
@@ -57,7 +57,33 @@ func (me *service) Get(ctx context.Context, id string, v *workflows.Workflow) er
 		return err
 	}
 
-	return json.Unmarshal(response.Data, &v)
+	if err := json.Unmarshal(response.Data, &v); err != nil {
+		return err
+	}
+	deComputeTaskPosition(ctx, v)
+
+	return nil
+}
+
+// deComputeTaskPosition removes API set defaults of task positions.
+// The reason here is simply because the TypeSet hash changes every time if a field of an item is set to computed even if nothing changed.
+// If the position is set by the user in the Terraform file, then it's correctly compared to the API set one.
+// DiffSuppress can't help here because even if the task doesn't change, the bug of TypeSet happens, which adds a new empty item which leads to a non-empty plan
+func deComputeTaskPosition(ctx context.Context, v *workflows.Workflow) {
+	positionTaskNames := map[string]bool{}
+	if stateConfig, ok := ctx.Value(settings.ContextKeyStateConfig).(*workflows.Workflow); ok {
+		for _, st := range stateConfig.Tasks {
+			if st.Position != nil {
+				positionTaskNames[st.Name] = true
+			}
+		}
+	}
+	for _, task := range v.Tasks {
+		if !positionTaskNames[task.Name] {
+			// => if not in state: set position (returned from API) to nil
+			task.Position = nil
+		}
+	}
 }
 
 func (me *service) SchemaID() string {
--- a/dynatrace/api/automation/workflows/testdata/terraform/example-g.tf
+++ b/dynatrace/api/automation/workflows/testdata/terraform/example-g.tf
@@ -1,0 +1,18 @@
+resource "dynatrace_automation_workflow" "task_without_position" {
+  description            = ""
+  type                   = "STANDARD"
+  hourly_execution_limit = 1000
+  owner_type             = "USER"
+  private                = false
+  title                  = "#name#"
+  tasks {
+    task {
+      name        = "http_request_1"
+      description = "Issue an HTTP request to any API change"
+      action      = "dynatrace.automations:http-function"
+      active      = true
+    }
+  }
+  trigger {
+  }
+}
--- a/dynatrace/api/automation/workflows/settings/task.go
+++ b/dynatrace/api/automation/workflows/settings/task.go
@@ -153,8 +153,10 @@ func (me *Task) Schema(prefix string) map[string]*schema.Schema {
 			Description: "Layouting information about the task tile when visualized. If not specified Dynatrace will position the task tiles automatically",
 			MinItems:    1,
 			MaxItems:    1,
-			Optional:    true,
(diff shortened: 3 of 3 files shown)


dynatrace/api/automation/workflows/testdata/terraform/example-g.tf
```diff
@@ -0,0 +1,18 @@
...
+  tasks {
+    task {
+      name        = "http_request_1"
+      description = "Issue an HTTP request to any API change"
+      action      = "dynatrace.automations:http-function"
+      active      = true
+    }
+  }
+  trigger {
+  }
```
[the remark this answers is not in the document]
Kirdock: It's actually the opposite. Removing it will lead to a non-empty state after apply

dynatrace/api/automation/workflows/service.go
```diff
@@ -57,7 +57,33 @@ func (me *service) Get(ctx context.Context, id string, v *workflows.Workflow) er
...
+// The reason here is simply because the TypeSet hash changes every time if a field of an item is set to computed even if nothing changed.
+// If the position is set by the user in the Terraform file, then it's correctly compared to the API set one.
+// DiffSuppress can't help here because even if the task doesn't change, the bug of TypeSet happens, which adds a new empty item which leads to a non-empty plan
+func deComputeTaskPosition(ctx context.Context, v *workflows.Workflow) {
+	positionTaskNames := map[string]bool{}
+	if stateConfig, ok := ctx.Value(settings.ContextKeyStateConfig).(*workflows.Workflow); ok {
+		for _, st := range stateConfig.Tasks {
+			if st.Position != nil {
+				positionTaskNames[st.Name] = true
+			}
```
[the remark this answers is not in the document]
Kirdock: It's not the prior state, it's the current state.
If it's applied without a position and then updated with another position, the new position is seen as a change and the apply is working and leading to a non-empty plan. The state correctly contains the new position field

dynatrace/api/automation/workflows/testdata/terraform/example-g.tf
```diff
@@ -0,0 +1,18 @@
+resource "dynatrace_automation_workflow" "task_without_position" {
+  description            = ""
+  type                   = "STANDARD"
+  hourly_execution_limit = 1000
+  owner_type             = "USER"
+  private                = false
+  title                  = "TEST WORKLOW PLAN"
```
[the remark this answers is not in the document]
Kirdock: There isn't any uniquenes check for the workflow name, but still we should keep it consistent, yes

d0weinberger (approved):

tomazpu (commented):
I think the limitation should be documented in the resource description.

tomazpu (approved):

Kirdock:
> I think the limitation should be documented in the resource description.

@tomazpu 
Do you have any ideas for what we should put in there? I would guess that anything in there could just confuse the user 🤔

dynatrace/api/automation/workflows/service.go
```diff
@@ -57,7 +57,33 @@ func (me *service) Get(ctx context.Context, id string, v *workflows.Workflow) er
...
+
+	return nil
+}
+
+// deComputeTaskPosition removes API set defaults of task positions.
+// The reason here is simply because the TypeSet hash changes every time if a field of an item is set to computed even if nothing changed.
+// If the position is set by the user in the Terraform file, then it's correctly compared to the API set one.
+// DiffSuppress can't help here because even if the task doesn't change, the bug of TypeSet happens, which adds a new empty item which leads to a non-empty plan
+func deComputeTaskPosition(ctx context.Context, v *workflows.Workflow) {
+	positionTaskNames := map[string]bool{}
```
arthurpitman, line 73: Do we need the `bool` here? We only ever add entries to the map with `true`, so wouldn't a `map[string]struct{}{}` suffice?

dynatrace/api/automation/workflows/service.go
```diff
@@ -57,7 +57,33 @@ func (me *service) Get(ctx context.Context, id string, v *workflows.Workflow) er
...
+	if stateConfig, ok := ctx.Value(settings.ContextKeyStateConfig).(*workflows.Workflow); ok {
+		for _, st := range stateConfig.Tasks {
+			if st.Position != nil {
+				positionTaskNames[st.Name] = true
+			}
+		}
+	}
+	for _, task := range v.Tasks {
+		if !positionTaskNames[task.Name] {
```
arthurpitman, line 82: In that case then, here we'd just need to look if `positionTaskNames` contains the key.

arthurpitman (approved):

dynatrace/api/automation/workflows/service.go
arthurpitman: More of a nit

Kirdock: It's a bit simpler to use instead of `ok, val := positionTaskNames[task.Name]; !ok` and as the defautl value of bool is false, I wouldn't see any issues here

PR #1041, opened 2026-04-01
fix: cache IAM and Platform REST clients to prevent OAuth 429

Kirdock:
#### **Why** this PR?
There may be 429 for every request to platform, which could break the whole Terraform plan/apply

#### **What** has changed?
Should not run into 429 errors anymore, as this changes from O(n) to 1 (within 5 minutes; token duration).

#### **How** does it do it?
Introduce caching for OAuth/rest clients to avoid repeated token requests and potential 429s.

#### How is it **tested**?
Tests should still work

Less OAuth requests and less likely to run into 429

**Issue:** CA-18762

### DIFF
--- a/dynatrace/rest/platform_request.go
+++ b/dynatrace/rest/platform_request.go
@@ -25,6 +25,7 @@ import (
 	"net/url"
 	"regexp"
 	"strings"
+	"sync"
 
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/rest/logging"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider/version"
@@ -38,6 +39,11 @@ var eligiblePlatformRequests = map[string]string{
 	"/api/v2/settings/objects": "/platform/classic/environment-api/v2/settings/objects",
 	"/api/v2/settings/schemas": "/platform/classic/environment-api/v2/settings/schemas",
 }
+
+// cachedPlatformClients caches clients for OAuth requests
+// The point of it is that there aren't x requests to get an access tokens as this could run into 429
+var cachedPlatformClients = map[string]*rest.Client{}
+var clientMutex sync.Mutex
 
 type platform_request request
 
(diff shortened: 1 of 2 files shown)


dynatrace/rest/platform_request.go
```diff
@@ -84,22 +90,30 @@ func CreatePlatformTokenClient(u string, credentials *Credentials) (*rest.Client
 	return configureCommonRestClient(rest.NewClient(parsedURL, NewBearerTokenBasedClient(credentials.OAuth.PlatformToken), opts...))
 }
 
-func CreatePlatformClient(ctx context.Context, platformURL string, credentials *Credentials) (*rest.Client, error) {
+func CreatePlatformClient(_ context.Context, platformURL string, credentials *Credentials) (*rest.Client, error) {
```
Laubi, line 93: I don't think that caching per platform URL is a sufficient way to cache. Do two provider initialization with the same platform url but different client credentials then share the cache?

Kirdock: At least up to now, we haven't had any issues, as this was present before
https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/970/changes#diff-2f571835658297e2f0367b78e088f1c41b17e358691d38ffc52c0680b5f4e1b8L45
Now with the change that the same context isn't used


But I'll double-check if they share the cache 👍

Kirdock: I just verified it with two provider blocks.
With the built provider, these are two separate caches
(log line was placed right after assigning the client to the cache)
image

Example used:
```hcl
provider "dynatrace" {
  dt_env_url       = var.DYNATRACE_ENV_URL
  account_id       = var.DYNATRACE_ACCOUNT_ID
  iam_endpoint_url = var.ACCOUNT_URL
  dt_api_token     = var.DYNATRACE_API_TOKEN
  client_id        = var.DYNATRACE_CLIENT_ID
  client_secret    = var.DYNATRACE_CLIENT_SECRET
}

provider "dynatrace" {
  dt_env_url       = var.DYNATRACE_ENV_URL
  account_id       = var.DYNATRACE_ACCOUNT_ID
  iam_endpoint_url = var.ACCOUNT_URL
  dt_api_token     = var.DYNATRACE_API_TOKEN
  client_id        = var.DYNATRACE_CLIENT_ID
  client_secret    = var.DYNATRACE_CLIENT_SECRET

  alias = "secondary"
}


resource "dynatrace_iam_group" "group" {
  name = "tf-group-1"
}

resource "dynatrace_iam_group" "group11" {
  name = "tf-group-11"
}

resource "dynatrace_iam_group" "group2" {
  name = "tf-group-2"

  provider = dynatrace.secondary
}

resource "dynatrace_iam_group" "group22" {
  name = "tf-group-22"

  provider = dynatrace.secondary
}
```

But note that this is different when running it (running the provider and providing the ENV vs built binary)
If you just run the provider locally instead of building it, it fills the cache once, because it's the same process.

Kirdock: V2 (slower one) with a retry mechanism: https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/1042

Kirdock:
Will do a proper way in the future

PR #1042, opened 2026-04-01
feat: add retry mechanism for OAuth 429

Kirdock:
#### **Why** this PR?
There may be 429 for every request to platform, which could break the whole Terraform plan/apply

#### **What** has changed?
429 returned by OAuth token requests are now retried

#### **How** does it do it?
This adds special handling for OAuth requests in case of 429. It now retries based on the set retry headers with a fallback to an exponential backoff with a maximum retry amount of 10

#### How is it **tested**?
Tests added and integration tests are still working

Less OAuth requests and less likely to run into 429

## Notes
V2 of https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/1041
Still, this one (with retry) doesn't solve the cause (too many not needed requests) and could lead to increasing execution times by a factor more than 10 compared to the cache.

**Issue:** CA-18762

### DIFF
--- a/dynatrace/api/iam/iam_client.go
+++ b/dynatrace/api/iam/iam_client.go
@@ -58,7 +58,7 @@ func NewIAMClient(a Authenticator) IAMClient {
 		ClientSecret: a.ClientSecret(),
 		TokenURL:     a.TokenURL(),
 	}
-	httpClient := auth.NewOAuthClient(context.Background(), &oauthConfig)
+	httpClient := auth.NewOAuthClient(rest.NewContextWithOAuthRetryClient(context.Background()), &oauthConfig)
 
 	opts := []rest2.Option{
 		rest2.WithHTTPListener(logging.HTTPListener("iam")),
--- a/dynatrace/rest/retry_transport.go
+++ b/dynatrace/rest/retry_transport.go
@@ -1,0 +1,193 @@
+/*
+ * @license
+ * Copyright 2026 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package rest
+
+import (
+	"bytes"
+	"context"
+	"io"
+	"math"
+	"math/rand/v2"
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/rest/logging"
+	"golang.org/x/oauth2"
+)
+
+const (
+	defaultMaxRetries  = 5
+	defaultBaseBackoff = 5 * time.Second
+	defaultMaxBackoff  = 60 * time.Second
+)
+
+// RetryTransport is an http.RoundTripper that automatically retries requests
+// when the server returns HTTP 429 (Too Many Requests) or HTTP 503 (Service Unavailable).
+// MaxRetries, BaseBackoff, and MaxBackoff can be set to non-zero values to override the
+// defaults, which is useful in tests to keep execution time short.
+type RetryTransport struct {
+	Transport   http.RoundTripper
+	MaxRetries  int
+	BaseBackoff time.Duration
+	MaxBackoff  time.Duration
+}
+
+// transport returns the configured RoundTripper, falling back to http.DefaultTransport when nil.
+func (t *RetryTransport) transport() http.RoundTripper {
+	if t.Transport != nil {
+		return t.Transport
+	}
+	return http.DefaultTransport
+}
+
+func (t *RetryTransport) maxRetries() int {
+	if t.MaxRetries > 0 {
+		return t.MaxRetries
+	}
+	return defaultMaxRetries
+}
+
+func (t *RetryTransport) baseBackoff() time.Duration {
+	if t.BaseBackoff > 0 {
+		return t.BaseBackoff
+	}
+	return defaultBaseBackoff
+}
+
+func (t *RetryTransport) maxBackoff() time.Duration {
+	if t.MaxBackoff > 0 {
+		return t.MaxBackoff
+	}
+	return defaultMaxBackoff
+}
+
+func isRetriableStatus(statusCode int) bool {
+	return statusCode == http.StatusTooManyRequests || statusCode == http.StatusServiceUnavailable
+}
+
+// bufferRequestBody reads and closes req.Body, returning its contents.
+// Returns nil, nil when the body is absent.
+func bufferRequestBody(req *http.Request) ([]byte, error) {
+	if req.Body == nil {
+		return nil, nil
+	}
+	body, err := io.ReadAll(req.Body)
+	if closeErr := req.Body.Close(); err == nil {
+		err = closeErr
+	}
+	return body, err
+}
+
+// drainAndClose drains body into /dev/null and closes it, allowing TCP
+// connection reuse. The drain error takes precedence over the close error.
+func drainAndClose(body io.ReadCloser) error {
+	_, copyErr := io.Copy(io.Discard, body)
+	closeErr := body.Close()
+	if copyErr != nil {
+		return copyErr
+	}
+	return closeErr
+}
+
+func (t *RetryTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	// Buffer the request body so it can be replayed on retries.
+	bodyBytes, err := bufferRequestBody(req)
+	if err != nil {
+		return nil, err
+	}
+
+	maxRetries := t.maxRetries()
+
+	for attempt := 0; ; attempt++ {
+		// Clone the request and restore the body for each attempt.
+		clone := req.Clone(req.Context())
+		if bodyBytes != nil {
+			clone.Body = io.NopCloser(bytes.NewReader(bodyBytes))
+		}
+
+		resp, err := t.transport().RoundTrip(clone)
+		if err != nil {
+			return nil, err
+		}
+
+		if !isRetriableStatus(resp.StatusCode) || attempt >= maxRetries {
+			return resp, nil
+		}
+
+		ctx := clone.Context()
+		wait := t.sleepDuration(resp, attempt)
+		logging.Logger.Printf(ctx, "[RetryTransport] Received HTTP %d, retrying in %s (attempt %d/%d)", resp.StatusCode, wait, attempt+1, maxRetries)
+		if err := drainAndClose(resp.Body); err != nil {
+			return nil, err
+		}
+		select {
+		case <-time.After(wait):
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		}
+	}
+}
+
+// getRetryAfterHeaderAsSleepTime parses the Retry-After response header and returns the
+// indicated wait duration. Returns 0 if the header is absent,
+// cannot be parsed, or indicates a non-positive delay.
+func getRetryAfterHeaderAsSleepTime(resp *http.Response) time.Duration {
+	ra := resp.Header.Get("Retry-After")
+	if ra == "" {
+		return 0
+	}
+	if seconds, err := strconv.Atoi(ra); err == nil && seconds > 0 {
+		d := time.Duration(seconds) * time.Second
+		return d
+	}
+	if t, err := http.ParseTime(ra); err == nil {
+		if d := time.Until(t); d > 0 {
+			return d
+		}
+	}
+	return 0
+}
+
+// computeBackoffSleepTime returns the exponential back-off duration for the given attempt
+// using a 1.5× multiplier (base × 1.5^attempt), with up to 10%
+// random jitter to avoid thundering-herd problems.
+func computeBackoffSleepTime(base time.Duration, attempt int) time.Duration {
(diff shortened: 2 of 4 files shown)


dynatrace/rest/retry_transport.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+		if t, err := http.ParseTime(ra); err == nil {
+			if d := time.Until(t); d > 0 {
+				return d
+			}
+		}
+	}
+	// exponential backoff: defaultBackoff * 2^attempt
+	return time.Duration(float64(defaultBackoff) * math.Pow(2, float64(attempt)))
+}
```
arthurpitman: I would split this function into two:
- `getRetryAfterHeaderAsSleepTime(resp *http.Response) time.Duration`
- `computeBackoffSleepTime(attempt int) time.Duration`

dynatrace/rest/retry_transport.go
```diff
@@ -0,0 +1,104 @@
...
+		// Restore the body for each attempt.
+		if bodyBytes != nil {
+			req.Body = io.NopCloser(bytes.NewReader(bodyBytes))
+		}
+
+		resp, err = t.Transport.RoundTrip(req)
+		if err != nil {
+			return nil, err
+		}
+		if resp.StatusCode != http.StatusTooManyRequests {
```
arthurpitman: I was wondering if this is the only thing we should retry. What about `503`?

dynatrace/rest/retry_transport.go
```diff
@@ -0,0 +1,104 @@
...
+			return resp, nil
+		}
+
+		if attempt == maxRetries {
+			break
+		}
+
+		wait := backoff(resp, attempt)
+		resp.Body.Close()
```
arthurpitman: Do we also need to ensure the body is read to completion?

dynatrace/api/iam/iam_client.go
```diff
@@ -58,7 +58,7 @@ func NewIAMClient(a Authenticator) IAMClient {
 		ClientSecret: a.ClientSecret(),
 		TokenURL:     a.TokenURL(),
 	}
-	httpClient := auth.NewOAuthClient(context.Background(), &oauthConfig)
+	httpClient := auth.NewOAuthClient(rest.NewContextWithOAuthRetryClient(context.Background()), &oauthConfig)
```
tomazpu, line 61: I can see that even before we used an empty context. The question I ask myself is, since we're starting to implement a retry that can go on for a longer time(2560s), we should use a non-empty ctx, so if Terraform cancels the process, the cleanup happens earlier. Also, maybe we want to set the max time with context?

tomazpu: I understand, you did not introduce this. But I think we should fix it and also inject ctx like in the platform client.

dynatrace/rest/retry_transport.go
```diff
@@ -0,0 +1,104 @@
...
+const (
+	maxRetries     = 10
+	defaultBackoff = 5 * time.Second
+)
+
+type RetryTransport struct {
+	Transport http.RoundTripper
+}
+
+func (t *RetryTransport) RoundTrip(req *http.Request) (*http.Response, error) {
```
tomazpu, line 107: Should we log every time we retry?

dynatrace/rest/retry_transport.go
Kirdock: We haven't had any issues with SSO so far and 429 would be the only one that we would expect.
So I would say we don't need the 503 and can implement it if needed

dynatrace/rest/retry_transport.go
Kirdock: We don't need to read the body if we simply do a retry. It's enough to just close it.

dynatrace/rest/retry_transport.go
```diff
@@ -0,0 +1,104 @@
...
+		if err != nil {
+			return nil, err
+		}
+		if resp.StatusCode != http.StatusTooManyRequests {
+			return resp, nil
+		}
+
+		if attempt == maxRetries {
+			break
+		}
```
tomazpu: ` if resp.StatusCode != http.StatusTooManyRequests || attempt == maxRetries {                                                   
             return resp, nil                                                                                                          
}  `

tomazpu: I would combine both exit conditions

dynatrace/rest/retry_transport.go
```diff
@@ -0,0 +1,104 @@
...
+	"math"
+	"net/http"
+	"strconv"
+	"time"
+
+	"golang.org/x/oauth2"
+)
+
+const (
+	maxRetries     = 10
```
tomazpu: If they're settable we could have a faster test

dynatrace/api/iam/iam_client.go
Kirdock: This would require quite a bit of refactoring and affect several resources, and another point is that potential caching would not work because Terraform cancels the context for each resource.

dynatrace/rest/retry_transport.go
Kirdock: Could do that but it would slightly reduce the code coverage as the last return can never be reached

dynatrace/rest/retry_transport.go
Laubi: Maybe we should just handle it right now, so we don't have to worry about it in the future 🤔 
503 (or generally speaking, 5xx errors) can always happen. They shouldn't, but they do

dynatrace/rest/retry_transport.go
Laubi: This is not entirely true
https://pkg.go.dev/net/http#Client.Do
> If the returned error is nil, the [Response](https://pkg.go.dev/net/http#Response) will contain a non-nil Body which the user is expected to close. If the Body is not both read to EOF and closed, the [Client](https://pkg.go.dev/net/http#Client)'s underlying [RoundTripper](https://pkg.go.dev/net/http#RoundTripper) (typically [Transport](https://pkg.go.dev/net/http#Transport)) may not be able to re-use a persistent TCP connection to the server for a subsequent "keep-alive" request.

so this technically works, but is quite bad. the underlaying tcp connection cannot be reused.
1. Kills the existing connection
2. Forces a new TCP handshake
3. Forces a new TLS handshake

Please use a pattern like 
```go
io.Copy(io.Discard, resp.Body)
resp.Body.Close()
```

dynatrace/rest/retry_transport.go
```diff
@@ -0,0 +1,104 @@
...
+			return time.Duration(seconds) * time.Second
+		}
+		if t, err := http.ParseTime(ra); err == nil {
+			if d := time.Until(t); d > 0 {
+				return d
+			}
+		}
+	}
+	// exponential backoff: defaultBackoff * 2^attempt
+	return time.Duration(float64(defaultBackoff) * math.Pow(2, float64(attempt)))
```
Laubi: At 10 retries this is a backoff of 85 minutes, note sure if we should wait that long. Maybe we should go with 1.5 and only 5 retries?

dynatrace/rest/retry_transport.go
```diff
@@ -0,0 +1,104 @@
...
+		time.Sleep(wait)
+	}
+
+	return resp, nil
+}
+
+// backoff determines how long to wait before the next retry.
+// It first checks the Retry-After header; if absent it falls back to exponential backoff.
+func backoff(resp *http.Response, attempt int) time.Duration {
+	if ra := resp.Header.Get("Retry-After"); ra != "" {
```
Laubi: We should set an upper bound, right? Also, we should add tests for edge cases, like 0 and negative values. Just to ensure that if the API returns invalid values, we handle them gracefully.

dynatrace/rest/retry_transport.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+	}
+
+	var resp *http.Response
+	var err error
+
+	for attempt := 0; attempt <= maxRetries; attempt++ {
+		// Restore the body for each attempt.
+		if bodyBytes != nil {
+			req.Body = io.NopCloser(bytes.NewReader(bodyBytes))
```
Laubi: https://pkg.go.dev/net/http#RoundTripper

> // RoundTrip should not modify the request, except for
>	// consuming and closing the Request's Body. RoundTrip may
>	// read fields of the request in a separate goroutine. Callers
>	// should not mutate or reuse the request until the Response's
>	// Body has been closed.

Clone the request per attempt:
```go
clone := req.Clone(req.Context())
clone.Body = io.NopCloser(bytes.NewReader(bodyBytes))
resp, err = t.Transport.RoundTrip(clone)
```

dynatrace/rest/retry_transport.go
Laubi: nit: Maybe we should add some jitter?

dynatrace/rest/retry_transport.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+	}
+	// exponential backoff: defaultBackoff * 2^attempt
+	return time.Duration(float64(defaultBackoff) * math.Pow(2, float64(attempt)))
+}
+
+func NewContextWithOAuthRetryClient(ctx context.Context) context.Context {
+	return context.WithValue(ctx, oauth2.HTTPClient, &http.Client{
+		Transport: &RetryTransport{
+			Transport: http.DefaultTransport,
```
Laubi: Why have the transport as a field if you always set it to defaultTransport?

dynatrace/rest/retry_transport.go
```diff
@@ -0,0 +1,104 @@
...
+		req.Body.Close()
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	var resp *http.Response
+	var err error
+
+	for attempt := 0; attempt <= maxRetries; attempt++ {
```
Laubi: We need to adhere to the context cancellation. 

the context may be canceled at any time during the retry, so we need to react to it. 
Possible solution is selecting whatever happens first:
```go
select {
case <-time.After(wait):
case <-req.Context().Done():
    resp.Body.Close()
    return nil, req.Context().Err()
}
```

Laubi (changes requested):

dynatrace/rest/retry_transport.go
Laubi: I think we can use synctest: https://go.dev/blog/synctest

dynatrace/rest/retry_transport.go
```diff
@@ -0,0 +1,104 @@
...
+	"golang.org/x/oauth2"
+)
+
+const (
+	maxRetries     = 10
+	defaultBackoff = 5 * time.Second
+)
+
+type RetryTransport struct {
+	Transport http.RoundTripper
```
Laubi: Should this be public?
If yes, it should 100% be nil-checked

dynatrace/rest/retry_transport.go
Laubi: Why can it never be reached?

Kirdock: It either returns on not 429 or max-attempts reached, which is inside the for-loop.

dynatrace/rest/retry_transport.go
Kirdock: Adjusted to not set it to default but still have the field to easily override it for tests

dynatrace/rest/retry_transport.go
Kirdock: Testing related. Nil check added

dynatrace/api/iam/iam_client.go
Kirdock: Decided to do this in a follow-up 👍

dynatrace/rest/retry_transport.go
```diff
@@ -0,0 +1,201 @@
...
+	defaultBaseBackoff = 5 * time.Second
+	defaultMaxBackoff  = 60 * time.Second
+)
+
+// RetryTransport is an http.RoundTripper that automatically retries requests
+// when the server returns HTTP 429 (Too Many Requests) or HTTP 503 (Service Unavailable).
+// MaxRetries, BaseBackoff, and MaxBackoff can be set to non-zero values to override the
+// defaults, which is useful in tests to keep execution time short.
+type RetryTransport struct {
+	// Transport is the underlying RoundTripper. If nil, http.DefaultTransport is used.
```
tomazpu: NIT: I am ok with the comment above the struct, but the comments for the fields give me no value, except that the code is harder to read, in my personal opinion.

tomazpu: If others agree i would ask you to only leave relevant comments that add value.

tomazpu (dismissed):

tomazpu (dismissed):

dynatrace/api/iam/iam_client.go
Kirdock: Follow-up: https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/1048

dynatrace/rest/retry_transport.go
```diff
@@ -0,0 +1,196 @@
...
+// computeBackoffSleepTime returns the exponential back-off duration for the given attempt
+// using a 1.5× multiplier (base × 1.5^attempt), with up to 10%
+// random jitter to avoid thundering-herd problems.
+func computeBackoffSleepTime(base time.Duration, attempt int) time.Duration {
+	backoff := time.Duration(float64(base) * math.Pow(1.5, float64(attempt)))
+	if backoff > 0 {
+		// Add up to 10% random jitter.
+		backoff += time.Duration(rand.Int64N(int64(backoff)/10 + 1))
+	}
+	return backoff
```
arthurpitman:
```suggestion
	// random number between 1.0 and 1.1 to add up to 10% jitter
	scale := 1 + 0.1*rand.Float64()
	return time.Duration(float64(base) * scale * math.Pow(1.5, float64(attempt)))
```

arthurpitman (dismissed):

Laubi (approved):

PR #1045, opened 2026-04-02
feat: platform extension config

Kirdock:
#### **Why** this PR?
There is a new platform endpoint for managing extensions. This one supports OAuth and platform tokens and has a slightly different structure (response, request, etc.)

#### **What** has changed?
New resource added `dynatrace_hub_extension_v2_config` for managing extension configurations.

#### How is it **tested**?
New integration and unit tests added for CRUD operations on extension configs

They can make use of the new endpoint and use OAuth and platform tokens for this resource.

**Issue:** CA-18565

## Differences to V1
- Extensions are not automatically installed anymore


terraform/hcl/credentials_test.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+		hcl.ReplaceCredPlaceholders(tfState, apiState)
+		// "password" exists in apiState but not in tfState → kept as placeholder
+		assert.Equal(t, "***789***", apiState["password"])
+	})
+
+	t.Run("map: apiState not a map returns apiState unchanged", func(t *testing.T) {
+		tfState := map[string]any{"key": "value"}
+		assert.Equal(t, "not-a-map", hcl.ReplaceCredPlaceholders(tfState, "not-a-map"))
+	})
```
arthurpitman, line 94: nit question: What about apiState a map, and tfState a string? Unchanged map should be returned, right?

terraform/hcl/credentials_test.go
```diff
@@ -0,0 +1,140 @@
...
+		tfState := []any{"a", "b"}
+		apiState := []any{"***111***"}
+		result := hcl.ReplaceCredPlaceholders(tfState, apiState).([]any)
+		assert.Equal(t, "***111***", result[0])
+	})
+
+	t.Run("slice: apiState not a slice returns apiState unchanged", func(t *testing.T) {
+		tfState := []any{"a"}
+		assert.Equal(t, "not-a-slice", hcl.ReplaceCredPlaceholders(tfState, "not-a-slice"))
+	})
```
arthurpitman, line 114: nit question: What about apiState a slice, and tfState a string? Unchanged slice should be returned, right?

terraform/hcl/credentials.go
```diff
@@ -0,0 +1,51 @@
...
+			bSlice[i] = ReplaceCredPlaceholders(aVal[i], bSlice[i])
+		}
+	case string:
+		bStr, ok := apiState.(string)
+		if ok && credPlaceholderPattern.MatchString(bStr) {
+			return aVal
+		}
+	}
+	return apiState
+}
```
arthurpitman: As we're touching the code, renaming might increase readability here
```suggestion
// ReplaceCredPlaceholders replaces every credential matching "***...***" in apiValue with the value from stateValue.
func ReplaceCredPlaceholders(stateValue, apiValue any) any {
	switch stateValue := stateValue.(type) {
	case map[string]any:
		apiMap, ok := apiValue.(map[string]any)
		if !ok {
			return apiValue
		}
		for key, apiMapValue := range apiMap {
			if stateMapValue, exists := stateValue[key]; exists {
				apiMap[key] = ReplaceCredPlaceholders(stateMapValue, apiMapValue)
			}
		}
	case []any:
		apiSlice, ok := apiValue.([]any)
		if !ok || len(stateValue) != len(apiSlice) {
			return apiValue
		}
		for i := range apiSlice {
			apiSlice[i] = ReplaceCredPlaceholders(stateValue[i], apiSlice[i])
		}
	case string:
		apiStr, ok := apiValue.(string)
		if ok && credPlaceholderPattern.MatchString(apiStr) {
			return stateValue
		}
	}
	return apiValue
}
```

dynatrace/api/extensions/monitoringconfigurations/testcases/recreate-scope-change/update.tf
```diff
@@ -0,0 +1,22 @@
...
+resource "dynatrace_hub_extension_v2_config" "config" {
+  name  = "com.dynatrace.extension.jmx-weblogic-cp"
+  scope = data.dynatrace_entities.hosts.entities[0].entity_id // update with HOST-... instead of "environment"
+  value = jsonencode(
+    {
+      "activationContext" : "LOCAL",
+      "activationTags" : [],
+      "enabled" : true,
+      "description" : "update",
+      "version" : "2.1.1", // version and description update
```
arthurpitman: Is the `version` here meant to be the same as in `create.tf`?

dynatrace/api/extensions/monitoringconfigurations/testcases/update-no-recreate/update.tf
```diff
@@ -0,0 +1,18 @@
+resource "dynatrace_hub_extension_v2_config" "config" {
+  name  = "com.dynatrace.extension.jmx-weblogic-cp"
+  scope = "environment"
+  value = jsonencode(
+    {
+      "activationContext" : "LOCAL",
+      "activationTags" : [],
+      "enabled" : true,
+      "description" : "update",
+      "version" : "2.1.1", // version and description update
```
arthurpitman, line 10: Also here `version` is the same

dynatrace/api/extensions/monitoringconfigurations/settings/settings.go
```diff
@@ -0,0 +1,127 @@
+/**
+* @license
+* Copyright 2020 Dynatrace LLC
```
arthurpitman:
```suggestion
* Copyright 2026 Dynatrace LLC
```

dynatrace/api/extensions/monitoringconfigurations/settings/settings.go
```diff
@@ -0,0 +1,127 @@
...
+			Type:        schema.TypeString,
+			Description: "The JSON encoded value for this monitoring configuration",
+			Required:    true,
+			ValidateFunc: func(i any, k string) (warnings []string, errs []error) {
+				v, ok := i.(string)
+				if !ok {
+					errs = append(errs, fmt.Errorf("expected type of %s to be string", k))
+					return warnings, errs
+				}
+				m := map[string]any{}
```
arthurpitman: Is using this less verbose than a dedicated struct?

terraform/hcl/credentials_test.go
Kirdock: Yes, the API state should be returned.
Basically, always the API state should be returned unless it's a secret.

terraform/hcl/credentials_test.go
Kirdock: Yes, the API state should be returned.
Basically, always the API state should be returned unless it's a secret.

dynatrace/api/extensions/monitoringconfigurations/testcases/recreate-scope-change/update.tf
Kirdock: I removed the comment. In this case, we only care about the scope change

dynatrace/api/extensions/monitoringconfigurations/testcases/update-no-recreate/update.tf
Kirdock: updated it 👍

dynatrace/api/extensions/monitoringconfigurations/settings/settings.go
Kirdock: changed to struct 👍

arthurpitman (approved):

PR #1046, opened 2026-04-07
fix(dynatrace_iam_policy_bindings_v2): allow multiple policy bindings with the same policy ID

Kirdock:
#### **Why** this PR?
It was no longer possible to have two policy bindings to a group with the same policy ID. This is possible via the API if the parameters/metadata differ. Now it deletes the policy first, as we can't provide a unique identifier for a policy

#### **What** has changed?
The same policy ID can now be used several times for bindings in `dynatrace_iam_policy_bindings_v2`

#### **How** does it do it?
By simply removing all existing bindings of a group and re-assigning the current set ones.

#### How is it **tested**?
A new test was added that checks if policy bindings with the same policy ID can be created.

They can now set the same policy ID with different parameters for a policy binding.

**Issue:** CA-19008

## Notes
It is possible to lock the OAuth client out through this resource but this is very unlikely because managing of policies and groups is on the account level and requires account permissions. If account permissions are set on a group, it also allows managing boundaries even if no policies are set.

### DIFF
--- a/dynatrace/api/iam/v2bindings/service.go
+++ b/dynatrace/api/iam/v2bindings/service.go
@@ -20,6 +20,7 @@ package v2bindings
 import (
 	"context"
 	"fmt"
+	"net/url"
 	"strings"
 
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api"
@@ -215,6 +216,13 @@ type bindingPayload = struct {
 }
 
 func (me *BindingServiceClient) Update(ctx context.Context, id string, v *bindings.PolicyBinding) error {
+	// Because a policy binding to a group is not unique (several policyIDs with different parameters can be assigned to a group)
+	// we are deleting everything first before adding new policies
+	// updating/deleting a policy binding would mean that we need to provide the whole parameters and metadata array
+	// the policy-binding identifier is basically the value (parameters and metadata) of the policy binding
+	if err := me.Delete(ctx, id); err != nil {
+		return err
+	}
 	groupID, levelType, levelID, err := splitID(id)
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


dynatrace/api/iam/v2bindings/service.go
```diff
@@ -215,58 +216,29 @@ type bindingPayload = struct {
...
-		if _, ok := existingPolicyBindings[policyUUID]; !ok {
-			payload := bindingPayload{
-				Parameters: desiredPolicy.Parameters,
-				Metadata:   desiredPolicy.Metadata,
-				Boundaries: desiredPolicy.Boundaries,
-			}
-			if _, err = client.POST(ctx, fmt.Sprintf("%s/iam/v1/repo/%s/%s/bindings/%s/%s", me.endpointURL, levelType, levelID, policyUUID, groupID), payload, 204, false); err != nil {
-				return err
-			}
+		if _, err = client.POST(ctx, fmt.Sprintf("%s/iam/v1/repo/%s/%s/bindings/%s/%s", me.endpointURL, levelType, levelID, policyUUID, groupID), payload, 204, false); err != nil {
```
tomazpu, line 240: There is a risk that the Delete will succeed and the Post will fail. This would result in all policy bindings being removed, i think we should document this, wdyt? 

It would be good to add this info to the resource documentation, and with the note you wrote on the PR.

tomazpu (dismissed):

dynatrace/api/iam/v2bindings/service.go
Kirdock: Updated with
> This resource re-assigns all policies bound to a group, so every policy that should remain bound must be specified in the configuration; otherwise, it will be unbound.
During this process, there is a brief window where the group has no policies assigned, which may temporarily cause permission issues for users in that group.
Locking out the OAuth client through this resource is theoretically possible but very unlikely, as managing policies and groups requires account-level permissions. If account permissions are set on a group, policy boundaries can still be managed even when no policies are assigned.

tomazpu (approved):

PR #1047, opened 2026-04-08
fix(dynatrace_webhook_notification): remove empty item for headers 

Kirdock:
#### **Why** this PR?
Removing/Adding header items in `dynatrace_webhook_notification` lead to an empty item (with conditional defaults applied), which resulted in an API error.

#### **What** has changed?
The empty item is now removed

#### **How** does it do it?
By using our existing utility to filter out empty items

#### How is it **tested**?
New tests added
Also added a new testing utility to read several configs with the same identifier instead of always generating a new one (`#name#` => random string)

They can now properly add/remove header items in `dynatrace_webhook_notification`

**Issue:** CA-18992

### DIFF
--- a/dynatrace/api/builtin/problem/notifications/webhook/service_test.go
+++ b/dynatrace/api/builtin/problem/notifications/webhook/service_test.go
@@ -23,8 +23,35 @@ import (
 	"testing"
 
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/testing/api"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
 func TestAccWebHookNotifications(t *testing.T) {
 	api.TestAcc(t)
+}
+
+func TestAccWebHookHeaderUpdate(t *testing.T) {
+	if !api.AccEnvsGiven(t) {
+		return
+	}
+
+	identifier := acctest.RandStringFromCharSet(10, acctest.CharSetAlpha)
+	createConfig := api.ReadTfConfigWithIdentifier(t, "testdata/create.tf", identifier)
+	updateConfig := api.ReadTfConfigWithIdentifier(t, "testdata/update.tf", identifier)
+
+	providerFactories := map[string]func() (*schema.Provider, error){
+		"dynatrace": func() (*schema.Provider, error) {
+			return provider.Provider(), nil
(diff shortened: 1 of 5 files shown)


dynatrace/api/builtin/problem/notifications/webhook/service_test.go
```diff
@@ -23,8 +23,35 @@ import (
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
 func TestAccWebHookNotifications(t *testing.T) {
 	api.TestAcc(t)
 }
+
+func TestWebHookHeaderUpdate(t *testing.T) {
```
tomazpu: Should the name also start with TestAcc ?

Kirdock: I would say it's not needed as we have a build tag anyway that indicates that this is an integration/acceptance test

sergiofbsilva:
Hello 👋 

I think we have hit a similar error ⬇️ will this PR fix this as well? 

> │ Error: API error: webHookNotification/headers/0/name: Must contain at least 1 non-whitespace character
> │ webHookNotification/headers/0/name: Size must be between 1 and 500
> │ 
> │   with module.alerting.dynatrace_webhook_notification.this["global/om-team/dynatrace/alerting/om_team_dt-demo-kt-event-filters.json-003"],
> │   on .terraform/modules/alerting/dynatrace/alerting/notification.jsm.tf line 1, in resource "dynatrace_webhook_notification" "this":
> │    1: resource "dynatrace_webhook_notification" "this" {

Thanks!

Kirdock:
HI @sergiofbsilva , yes, this will fix this exact problem 👍

dynatrace/testing/api/settingstest.go
```diff
@@ -50,26 +50,28 @@ func AccEnvsGiven(t *testing.T) bool {
...
-// RandomizeResource replaces "#name#" and "${randomize}" with a random string
-// Returns:
-// 	- The replaced config
-//  - The random string that was used
-func RandomizeResource(config string) (replacedConfig string, identifier string) {
-	name := acctest.RandStringFromCharSet(10, acctest.CharSetAlpha)
-	config = strings.ReplaceAll(config, "#name#", name)
-	return strings.ReplaceAll(config, "${randomize}", name), name
+// randomizeResource replaces "#name#" and "${randomize}" with a given identifier
+func randomizeResource(config string, identifier string) string {
```
arthurpitman: I think we should rename this as it no longer does anything random. Maybe `replaceIndentifiersInResource`?

arthurpitman (approved):

PR #1053, opened 2026-04-15
fix: TypeSet maintenance part 2

Kirdock:
#### **Why** this PR?
This is part 2 of several, where tests for all TypeSet fields are added and potential bugs are fixed.

#### **What** has changed?
This covers the `api/builtin` folder from A to M

#### **How** does it do it?

#### How is it **tested**?

**Issue:** CA-19022

### DIFF
--- a/dynatrace/api/builtin/dashboards/image/allowlist/service_test.go
+++ b/dynatrace/api/builtin/dashboards/image/allowlist/service_test.go
@@ -28,3 +28,7 @@ import (
 func TestAccDashboardsAllowlist(t *testing.T) {
 	api.TestAcc(t)
 }
+
+func TestAccTestCasesDashboardsAllowlist(t *testing.T) {
+	api.TestAccTestCases(t)
+}
--- a/dynatrace/api/builtin/deployment/oneagent/updates/testcases/update-maintenance-window-set/create.tf
+++ b/dynatrace/api/builtin/deployment/oneagent/updates/testcases/update-maintenance-window-set/create.tf
@@ -1,0 +1,26 @@
+resource "dynatrace_update_windows" "window" {
+  count = 2
+  name       = "#name#-${count.index}"
+  enabled    = true
+  recurrence = "ONCE"
+  once_recurrence {
+    recurrence_range {
+      end   = "2023-02-15T04:00:00Z"
+      start = "2023-02-15T02:00:00Z"
+    }
+  }
+}
+
+resource "dynatrace_oneagent_updates" "update" {
+  scope          = "environment"
+  target_version = "latest"
+  update_mode    = "AUTOMATIC_DURING_MW"
+  maintenance_windows {
+    dynamic "maintenance_window" {
+      for_each = toset(dynatrace_update_windows.window.*.id)
+      content {
+        maintenance_window = maintenance_window.value
+      }
+    }
+  }
+}
--- a/dynatrace/api/builtin/deployment/oneagent/updates/testcases/update-maintenance-window-set/create.tf
+++ b/dynatrace/api/builtin/deployment/oneagent/updates/testcases/update-maintenance-window-set/create.tf
@@ -1,0 +1,27 @@
+resource "dynatrace_update_windows" "window" {
+  count = 3
+  name       = "#name#-${count.index}"
+  enabled    = true
+  recurrence = "ONCE"
+  once_recurrence {
+    recurrence_range {
+      end   = "2023-02-15T04:00:00Z"
+      start = "2023-02-15T02:00:00Z"
+    }
+  }
+}
+
+resource "dynatrace_oneagent_updates" "update" {
+  scope          = "environment"
+  target_version = "latest"
+  update_mode    = "AUTOMATIC_DURING_MW"
+  maintenance_windows {
+    maintenance_window {
+      maintenance_window = dynatrace_update_windows.window[0].id
+    }
+    # to update
+    maintenance_window {
+      maintenance_window = dynatrace_update_windows.window[1].id
+    }
+  }
+}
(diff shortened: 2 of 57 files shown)


dynatrace/api/builtin/deployment/oneagent/updates/testcases/update-maintenance-window-set/create.tf
```diff
@@ -0,0 +1,26 @@
+resource "dynatrace_update_windows" "window" {
+  count = 2
```
arthurpitman, line 2: do we need the complexity of using count here?

Kirdock: Complexity? I just see it as quite nice because you don't have to specify this resource twice..
I could remove the for_each and dymanic block below though 🤔

dynatrace/api/builtin/monitoredentities/generic/relation/settings/source_filter.go
```diff
@@ -86,6 +91,11 @@ func (me *SourceFilter) MarshalHCL(properties hcl.Properties) error {
 }
 
 func (me *SourceFilter) HandlePreconditions() error {
+	var emptyItem SourceFilter
+	// ignore/don't throw on empty item that is coming from the TypeSet bug
+	if reflect.DeepEqual(*me, emptyItem) {
```
arthurpitman, line 96: this just allows us to keep the rest of the logic the same, right?

dynatrace/api/builtin/deployment/oneagent/updates/testcases/update-maintenance-window-set/create.tf
arthurpitman: I just had to look at it several times before I really understood what was going on. But I learnt something :)

dynatrace/api/builtin/monitoredentities/generic/relation/settings/source_filter.go
Kirdock: Exactly 👍

dynatrace/api/builtin/deployment/oneagent/updates/testcases/update-maintenance-window-set/create.tf
Kirdock: I updated it. Now it doesn't use the dynamic block and toSet.
Instead, I'm now creating 3 windows. In "create" I'm using window 1 and 2 and in update I'm using 1 and 3

dynatrace/api/builtin/dashboards/image/allowlist/service_test.go
```diff
@@ -28,3 +28,7 @@ import (
 func TestAccDashboardsAllowlist(t *testing.T) {
 	api.TestAcc(t)
 }
+
+func TestAccTestCasesDashboardsAllowlist(t *testing.T) {
+	api.TestAccTestCases(t)
```
d0weinberger, line 33: When would these test cases actually fail? I guess, if applying create.tf or update.tf fails with an error, or if the plan is non-empty after applying create.tf or update.tf, right?
I think it would make sense to just add this as a comment to `api.testAccTestCases` and also `api.TestAcc`. Maybe even rename the function `TestAcc` to something more descriptive, although I do not have a good suggestion right now, and I could not get Claude to suggest me something useful.

Kirdock: Fail conditions could be
- API errors during create, update, and cleanup (GET, POST, PUT, DELETE)
- Inconsistencies (GET after apply)
- Non-empty plan after create and update

> I think it would make sense to just add this as a comment to api.testAccTestCases and also api.TestAcc

So, we should basically describe what `resource.Test` is doing?

Even though, this would be a follow-up, as this is not part of the changes of this PR

dynatrace/api/builtin/logmonitoring/logevents/testcases/update-metadata-item-set/update.tf
```diff
@@ -0,0 +1,20 @@
...
+  event_template {
+    description = "Created by Terraform"
+    event_type  = "INFO"
+    title       = "{content}"
+    metadata {
+      item {
+        metadata_key   = "terraform.key1"
+        metadata_value = "terraform.value1"
+      }
+      item {
```
d0weinberger, line 15: Could you add the comment here as well to make clear what changed?

Kirdock: Done 👍

Kirdock: Even though, the testcase folder name also contains the sets that are changing

dynatrace/api/builtin/dashboards/image/allowlist/service_test.go
d0weinberger: >Fail conditions could be
> * API errors during create, update, and cleanup (GET, POST, PUT, DELETE)
> * Inconsistencies (GET after apply)
> * Non-empty plan after create and update

Yeah, having exactly this as a comment would help with understanding the tests. Fair point about this being rather a follow-up. It's more of a nit, really.

d0weinberger (approved):

arthurpitman (approved):

dynatrace/api/builtin/dashboards/image/allowlist/service_test.go
Kirdock: Added to https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/1064

PR #1057, opened 2026-04-15
fix: TypeSet maintenance part 4

Kirdock:
#### **Why** this PR?
This is part 4 of several, where tests for all TypeSet fields are added and potential bugs are fixed.

#### **What** has changed?
This covers the folders in `api` from `cluster` to `slo`

#### **How** does it do it?

#### How is it **tested**?

**Issue:** CA-19022


dynatrace/api/documents/directshares/testcases/update-recipients-set/update.tf
d0weinberger, line 46: [nit] you could also add a comment here that this is the updated part of the resource, to be consistent with other test files.

dynatrace/api/iam/v2bindings/settings/policy_binding.go
```diff
@@ -87,21 +87,7 @@ func (me *PolicyBinding) UnmarshalHCL(decoder hcl.Decoder) error {
...
-				continue
-			}
-			if len(policy.ID) == 0 {
-				continue
-			}
-			finalPolicies = append(finalPolicies, policy)
-		}
-		me.Policies = finalPolicies
-	}
+	me.Policies = hcl.FilterEmpty(me.Policies, Policy{})
```
d0weinberger, line 90: I'm just wondering whether there ever is a case where we actually want to allow an empty entry in a set

d0weinberger (dismissed):
Some nits and comments, but apart from that fine.

dynatrace/api/iam/v2bindings/settings/policy_binding.go
Kirdock: In this case it isn't even possible because of the required flags in this HCL schema but in general, even if it is allowed, we can't differentiate then if it's coming from the resource or the bug

dynatrace/api/documents/directshares/testcases/update-recipients-set/update.tf
Kirdock: Updated 👍

d0weinberger (approved):

PR #1059, opened 2026-04-16
fix: TypeSet maintenance part 5

Kirdock:
#### **Why** this PR?
This is part 5 of several, where tests for all TypeSet fields are added and potential bugs are fixed.

#### **What** has changed?
This covers the `V1` folders in `api`

#### **How** does it do it?

#### How is it **tested**?

**Issue:** CA-19022

### DIFF
--- a/dynatrace/api/v1/config/synthetic/monitors/browser/service_test.go
+++ b/dynatrace/api/v1/config/synthetic/monitors/browser/service_test.go
@@ -29,3 +29,9 @@ import (
 func TestAccBrowserMonitors(t *testing.T) {
 	api.TestAcc(t, api.TestAccOptions{ExternalProviders: map[string]resource.ExternalProvider{"time": {Source: "hashicorp/time"}}})
 }
+
+func TestAccTestCasesBrowserMonitors(t *testing.T) {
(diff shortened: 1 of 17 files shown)


dynatrace/api/v1/config/synthetic/monitors/browser/service_test.go
```diff
@@ -29,3 +29,9 @@ import (
 func TestAccBrowserMonitors(t *testing.T) {
 	api.TestAcc(t, api.TestAccOptions{ExternalProviders: map[string]resource.ExternalProvider{"time": {Source: "hashicorp/time"}}})
 }
+
+func TestAccTestCasesBrowserMonitors(t *testing.T) {
+	t.Skip("Skipping because the eventual consistency time is way too high")
```
d0weinberger, line 34: What would the path to eventually enabling this test case be?

d0weinberger (approved):

dynatrace/api/v1/config/synthetic/monitors/browser/service_test.go
Kirdock: DEM-20089

PR #1060, opened 2026-04-16
fix(dynatrace_calculated_web_metric): update didn't work

Kirdock:
#### **Why** this PR?
Update doesn't seem to work on any property of the `dynatrace_calculated_web_metric` resource

#### **What** has changed?
It's now possible to change all properties.

#### **How** does it do it?
The API actually only accepts `enabled` for updates. If something else needs to be updated, the resource must be re-created. Therefore, adding `ForceNew: true` to every property, except `enabled`. Also adding a static sleep after delete because the API seems to have a cache, which could cause issues.

#### How is it **tested**?
A new test was added that checks if an update actually works

They can now update all properties of `dynatrace_calculated_web_metric`.

**Issue:** NOISSUE


Kirdock:
V2 with retry instead of sleep: https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/1060

Kirdock:
choosing a different solution

PR #1065, opened 2026-04-17
fix: Use correct payload type for `default_sampling_ratio` in `dynatrace_dql`

arthurpitman:
#### **Why** this PR?
The type for `default_sampling_ratio` in `dynatrace_dql` is set to `float`. In the payload and the associated type assert, it was `int`, causing a panic.

The type-assert and type of `default_sampling_ratio` in the payload has been changed to `float64`.

#### How is it **tested**?
A rudimentary test, `TestAccDatasourceDQL`,  has been added and now passes with the fix.

Users can set `default_sampling_ratio` when using the `dynatrace_dql` data source.

**Issue:** CA-19165

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -140,6 +140,7 @@ jobs:
           - { name: "builtin (n-o)", path: "$(go list ./dynatrace/api/builtin/... | grep -E '^.+/dynatrace/api/builtin/[n-o][^/]*/?')" }
           - { name: "builtin (p-s)", path: "$(go list ./dynatrace/api/builtin/... | grep -E '^.+/dynatrace/api/builtin/[p-s][^/]*/?')" }
           - { name: "builtin (t-z)", path: "$(go list ./dynatrace/api/builtin/... | grep -E '^.+/dynatrace/api/builtin/[t-z][^/]*/?')" }
+          - { name: "data sources", path: "./datasources/..." }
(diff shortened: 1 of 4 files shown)


datasources/dql/testdata/terraform/example_a.tf
```diff
@@ -0,0 +1,16 @@
...
+  max_result_records = 10
+  max_result_bytes = 10485760
+  query = <<-EOT
+    data record(id = 1) | fields id
+  EOT
+}
+
+output "events" {
+  value = data.dynatrace_dql.events
+}
```
Kirdock: At the moment, does this output have any benefit?

[nit]: We could test the output of the data source. E.g., like this

```go
var id string
	resource.Test(t, resource.TestCase{
		ProviderFactories: map[string]func() (*schema.Provider, error){
			"dynatrace": func() (*schema.Provider, error) {
				return provider.Provider(), nil
			},
		},
		Steps: []resource.TestStep{
			{
				Config: f,
				Check: func(state *terraform.State) error {
					id = state.RootModule().Outputs["events"].Value.(string)
					return nil
				},
			},
		},
	})
	assert.Equal(t, "1", id)
```

with an output like this
```hcl
output "events" {
  value = jsondecode(data.dynatrace_dql.events.records)[0].id
}
```

Kirdock (approved):

PR #1067, opened 2026-04-20
feat: resource updates for 336

Kirdock:
#### **Why** this PR?
Some APIs have changed with Dynatrace version 336. Therefore, the Terraform resources also need updating.

#### **What** has changed?
New fields and descriptions

#### **How** does it do it?

Description updates:
- `dynatrace_extension_execution_remote`
- `dynatrace_aws_connection`
- `dynatrace_azure_connection`
- `dynatrace_log_agent_feature_flags`
- 

Changes:
- `dynatrace_automation_approval`: New optional field `external_approvals_enabled`
- `dynatrace_container_builtin_rule `: New optional fields (required in schema but we use the mentioned default) `ignore_open_shift_etcd_namespace`, `ignore_open_shift_ingress_canary_namespace`, `ignore_open_shift_kube_apiserver_namespace`, `ignore_open_shift_machine_config_operator_namespace`, `ignore_open_shift_monitoring_namespace`, `ignore_open_shift_ovn_kubernetes_namespace`

#### How is it **tested**?
For new properties, there are new examples.
Else, it's mostly about descriptions.

**Issue:** CA-19065


d0weinberger:
>Some APIs have changed with Dynatrace version 334.44

You mean 336?

dynatrace/api/builtin/automation/approval/testdata/terraform/example_a.tf
```diff
@@ -1,3 +1,4 @@
-resource "dynatrace_automation_approval" "#name#" {
+resource "dynatrace_automation_approval" "wf_approval" {
```
d0weinberger, line 1: Why the change from `#name#`?

Kirdock: Because there is no point in using `#name#` for the resource name and because of invalid characters, I can't simply run `terraform plan/apply`

d0weinberger (approved):

PR #1061, opened 2026-04-16
chore: Update to Go 1.26

Laubi:
#### **Why** this PR?

Go released in February with multiple improvements to the syntax, standard library, and security improvements. 
This PR introduces Go 1.26 from 1.25. 

#### **What** has changed?

`go fix` has ben introduced to generate the changeset. An additional commit contains all additional manual changes.

#### How does it affect **users**?

**Issue:** noissue

PR #1068, opened 2026-04-20
refactor: Simplify pointer creation and a couple of casts

arthurpitman:
#### **Why** this PR?
A few things to tidy up while reviewing #1061.

- Simplify the creation of several pointers
- Remove a couple of redundant casts

#### How is it **tested**?
NA

NA

**Issue:** NA

### DIFF
--- a/dynatrace/api/builtin/anomalydetection/databases/settings/load_drops.go
+++ b/dynatrace/api/builtin/anomalydetection/databases/settings/load_drops.go
@@ -63,7 +63,7 @@ func (me *LoadDrops) UnmarshalHCL(decoder hcl.Decoder) error {
 		"minutes_abnormal_state": &me.MinutesAbnormalState,
 	})
 	if me.LoadDropPercent == nil && me.Enabled {
-		me.LoadDropPercent = new(float64(0))
+		me.LoadDropPercent = new(0.0)
 	}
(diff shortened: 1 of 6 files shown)


Laubi (approved):

dynatrace/api/builtin/anomalydetection/databases/settings/load_drops.go
```diff
@@ -63,7 +63,7 @@ func (me *LoadDrops) UnmarshalHCL(decoder hcl.Decoder) error {
 		"minutes_abnormal_state": &me.MinutesAbnormalState,
 	})
 	if me.LoadDropPercent == nil && me.Enabled {
-		me.LoadDropPercent = new(float64(0))
+		me.LoadDropPercent = new(0.0)
```
Kirdock, line 66: Does something speak against using `new(float64)`?

Kirdock: Nvm, when it's about setting a default with a value we have, `0.0` may be more consistent across other files with different defaults.

arthurpitman: Correct - the majority are already like this

PR #1078, opened 2026-04-22
test: make builtin configs more environment independent

Kirdock:
#### **Why** this PR?
It's not possible to execute some E2E in other environments because of hardcoded IDs.

#### **What** has changed?
Builtin configs (except OpenPipeline) can now be executed on a different environment

#### **How** does it do it?
- By replacing hardcoded location IDs with a datasource or a resource
- By skipping deprecated OpenPipeline schemas
- By changing a generic settings example to use a builtin setting
- By changing examples so that they can be applied (e.g., dynatrace_activegate_token)

#### How is it **tested**?
Tests updated and are working

N/A

**Issue:** CA-17770

### DIFF
--- a/dynatrace/api/builtin/davis/anomalydetectors/testcases/update-analyzer-input-field-set/create.tf
+++ b/dynatrace/api/builtin/davis/anomalydetectors/testcases/update-analyzer-input-field-set/create.tf
@@ -37,5 +37,10 @@ resource "dynatrace_davis_anomaly_detectors" "detector" {
     }
   }
   execution_settings {
+    actor = dynatrace_iam_service_user.user.id
   }
+}
+
+resource "dynatrace_iam_service_user" "user" {
+  name = "#name#"
 }
--- a/dynatrace/api/builtin/host/monitoring/mode/testdata/terraform/example_a.tf
+++ b/dynatrace/api/builtin/host/monitoring/mode/testdata/terraform/example_a.tf
@@ -1,4 +1,8 @@
-resource "dynatrace_host_monitoring_mode" "#name#" {
-  host_id         = "HOST-E532D48B0DBD5A6B"
+resource "dynatrace_host_monitoring_mode" "mode" {
+  host_id         = data.dynatrace_entities.hosts.entities[0].entity_id
   monitoring_mode = "FULL_STACK"
+}
+
+data "dynatrace_entities" "hosts" {
+  type = "HOST"
 }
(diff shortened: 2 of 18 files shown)


dynatrace/api/builtin/host/monitoring/mode/testdata/terraform/example_a.tf
```diff
@@ -1,4 +1,8 @@
-resource "dynatrace_host_monitoring_mode" "#name#" {
-  host_id         = "HOST-E532D48B0DBD5A6B"
+resource "dynatrace_host_monitoring_mode" "mode" {
+  host_id         = data.dynatrace_entities.hosts.entities[0].entity_id
   monitoring_mode = "FULL_STACK"
-}
\ No newline at end of file
+}
+
+data "dynatrace_entities" "hosts" {
```
arthurpitman, line 6: While this is a definite improvement, is it certain that a host is available in every environment?

dynatrace/api/builtin/davis/anomalydetectors/testcases/update-analyzer-input-field-set/create.tf
```diff
@@ -37,5 +37,10 @@ resource "dynatrace_davis_anomaly_detectors" "detector" {
     }
   }
   execution_settings {
+    actor = dynatrace_iam_service_user.user.id
   }
 }
+
+resource "dynatrace_iam_service_user" "user" {
+  name = "#name#"
```
arthurpitman, line 45: While using `#name#` can't really hurt, is it required? I was thinking the name is not required to be unique.

arthurpitman (approved):

dynatrace/api/builtin/host/monitoring/mode/testdata/terraform/example_a.tf
Kirdock: No, but we must have one to test the resource. It's not possible with some placeholder value like HOST-00000.
We also can't create one ourselves, since this is handled by the OneAgent, right?

dynatrace/api/builtin/davis/anomalydetectors/testcases/update-analyzer-input-field-set/create.tf
Kirdock: It's not required yes, but with this, we are more aligned with other resources/tests/examples

dynatrace/api/builtin/host/monitoring/mode/testdata/terraform/example_a.tf
arthurpitman: I think yes, it requires something to be monitored, e.g. through a OneAgent install

PR #1080, opened 2026-04-27
fix: remove double logging

Kirdock:
#### **Why** this PR?
We had some double logging lines for requests and responses.

#### **What** has changed?
Removed double logging

#### **How** does it do it?
By removing the additional logging with the round tripper and relying on the core client logging, which we already have.

#### How is it **tested**?
New test added for logging (IAM, platform, classic)

They will see less duplicated logging lines.

**Issue:** CA-19071


dynatrace/rest/logging/http_listener.go
arthurpitman: While we're touching this area, should we change the if to use an or, and handle both conditions, as the action is the same?

dynatrace/rest/logging/http_listener.go
```diff
@@ -26,7 +26,19 @@ import (
...
+		return
+	}
+	if request.Body == nil {
+		Logger.Printf(ctx, "[%s] [%s] [REQUEST] %s %s", prefix, id, request.Method, request.URL.String())
+		return
+	}
+
+	body, _ := io.ReadAll(request.Body)
+	Logger.Printf(ctx, "[%s] [%s] [REQUEST] %s %s", prefix, id, request.Method, request.URL.String())
+	Logger.Printf(ctx, "           [%s] [PAYLOAD] %s", id, string(body))
```
arthurpitman: Should this line also include the prefix? After all, multiple resources might create logs concurrently.

Kirdock: We still have the request ID that we set on both lines, though, which is the same and unique for a request and the related response.
Still, I can add it if needed

dynatrace/rest/logging/http_listener.go
arthurpitman: No indentation here or in line 53, compared with the lines below

dynatrace/rest/legacy_request.go
```diff
@@ -113,10 +114,10 @@ func (me *legacy_request) Raw() ([]byte, error) {
 	}
 
 	if len(data) > 0 {
-		Logger.Printf(me.ctx, "[legacy  ] [%s] [REQUEST ] %s %s", me.id, me.method, url)
-		Logger.Printf(me.ctx, "           [%s] [PAYLOAD ] %s", me.id, string(data))
+		logging.Logger.Printf(me.ctx, "[legacy] [%s] [REQUEST] %s %s", me.id, me.method, url)
```
arthurpitman: Is `[legacy]` like the prefix?

Kirdock: yes

dynatrace/rest/legacy_request.go
```diff
@@ -113,10 +114,10 @@ func (me *legacy_request) Raw() ([]byte, error) {
...
+		logging.Logger.Printf(me.ctx, "           [%s] [PAYLOAD] %s", me.id, string(data))
 	} else {
-		Logger.Printf(me.ctx, "[legacy  ] [%s] [REQUEST ] %s %s", me.id, me.method, url)
+		logging.Logger.Printf(me.ctx, "[legacy] [%s] [REQUEST] %s %s", me.id, me.method, url)
```
arthurpitman: Isn't this line technically done either way? In which case, it could be refactored to be before the `if` and then the `else` could be removed?

Kirdock: I didn't want to open the box of pandora here and refactor every tiny bit of technical debt. The double logging was the main issue here 🫠. Everything else, like an additional prefix in logs, etc. is out of scope

dynatrace/rest/legacy_request.go
```diff
@@ -262,8 +263,8 @@ func (s *legacy_request) execute(ctx context.Context, callback func() (*http.Res
 	for response.StatusCode == http.StatusTooManyRequests && currentIteration < maxIterationCount {
 
 		if limit, humanReadableTimestamp, timeInMicroseconds, err := s.extractRateLimitHeaders(response); err == nil {
-			Logger.Printf(ctx, "Rate limit of %s requests/min reached (iteration: %d)", limit, currentIteration+1)
-			Logger.Printf(ctx, "Attempting to sleep until %s", humanReadableTimestamp)
+			logging.Logger.Printf(ctx, "Rate limit of %s requests/min reached (iteration: %d)", limit, currentIteration+1)
+			logging.Logger.Printf(ctx, "Attempting to sleep until %s", humanReadableTimestamp)
```
arthurpitman, line 265: Is this still useful code?

Kirdock: Yes, for the legacy_request we don't have the core client

arthurpitman (dismissed):
Can be merged as is, but let's do a follow-up to make this nicer.

dynatrace/rest/logging/http_listener.go
Kirdock: Refactored it a bit

dynatrace/rest/logging/http_listener.go
Kirdock: I now added the prefix to every request and response

dynatrace/rest/logging/http_listener.go
Kirdock: Removed the indentation for all because I added the prefix to all requests and responses now

dynatrace/rest/legacy_request.go
Kirdock: If you mean the whole legacy stuff and retry, I simply don't even know why we have it and it's not documented at all

dynatrace/rest/logging/http_listener.go
```diff
@@ -76,7 +70,7 @@ func HTTPListener(prefix string) *crest.HTTPListener {
 		Callback: func(response crest.RequestResponse) {
 			ctx := requestContext(response)
 			logRequest(ctx, response.ID, response.Request, prefix)
-			logResponse(ctx, response.ID, response.Response)
+			logResponse(ctx, response.ID, response.Response, prefix)
 		},
```
arthurpitman, line 74: Just looking at the code here, I'm not sure what the `prefix` is really for, but should it be added to the response log lines?

dynatrace/rest/logging/http_listener.go
```diff
@@ -60,34 +72,10 @@ func requestContext(response crest.RequestResponse) context.Context {
 }
 
 func HTTPListener(prefix string) *crest.HTTPListener {
```
arthurpitman, line 74: Just looking at the code here, I'm not sure what the `prefix` is really for, but should it be added to the response log lines?

Kirdock: The prefix is simply iam/platform/classic/etc., so basically each different core client we create

arthurpitman (approved):

PR #1081, opened 2026-04-27
fix: flaky dashboard general creation

Kirdock:
#### **Why** this PR?
When creating dashboards and user groups, the data source of `dynatrace_dashboard_general` may not be up to date yet. This may lead to an error during create

#### **What** has changed?
It should not fail anymore if the referenced resource was just created

#### **How** does it do it?
By adding custom retry logic in the resource (max 10s)

#### How is it **tested**?
Test now succeeds without time_sleep

More consistent creation of `dynatrace_dashboard_general`

**Issue:** NOISSUE

### DIFF
--- a/dynatrace/api/builtin/dashboards/general/service.go
+++ b/dynatrace/api/builtin/dashboards/general/service.go
@@ -18,15 +18,83 @@
 package general
 
 import (
+	"context"
+	"errors"
+	"net/http"
+	"slices"
+	"strings"
+	"time"
+
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api"
 	general "github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/builtin/dashboards/general/settings"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/rest"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/settings"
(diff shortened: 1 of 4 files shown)


dynatrace/api/builtin/dashboards/general/service.go
```diff
@@ -18,15 +18,79 @@
...
+	service settings.CRUDService[*general.Settings]
+}
+
 func Service(credentials *rest.Credentials) settings.CRUDService[*general.Settings] {
-	return settings20.Service[*general.Settings](credentials, SchemaID, SchemaVersion)
+	return &service{settings20.Service[*general.Settings](credentials, SchemaID, SchemaVersion)}
+}
+func (me *service) Create(ctx context.Context, v *general.Settings) (*api.Stub, error) {
+	var apiStub *api.Stub
+	err := retry.RetryContext(ctx, 10*time.Second, func() *retry.RetryError {
```
arthurpitman: Should the retry timeout be configurable via `v.DefaultTimeout` or similar like we do for Azure or AWS HA?

Kirdock: I would say it isn't needed, and we don't need to confuse the customer.
The default case, compared to e.g, the AWS connector, is that it should work.
If they need a configurable timeout, they can also make use of the `time_sleep` resource that adds a create duration.

arthurpitman (approved):

PR #1083, opened 2026-04-28
feat(extensions): data source for latest version

Kirdock:
#### **Why** this PR?
At the moment, it's not possible to get the latest installed version of an extension (for the platform-based resource).

#### **What** has changed?
There is a new data source `dynatrace_hub_extension_v2_latest_version` for getting the latest version of an extension.

#### **How** does it do it?
By adding a data source that calls `/extensions/{extension-name}` and returns the latest version that's inside the returned array via semver check.
As for now, there is no fixed guarantee that the returned array is sorted.

#### How is it **tested**?
New tests added

They can get the latest version of an extension.

**Issue:** CA-19280


datasources/extensions/latest_version/data_source_test.go
```diff
@@ -0,0 +1,46 @@
...
+	resource.Test(t, resource.TestCase{
+		ProviderFactories: map[string]func() (*schema.Provider, error){
+			"dynatrace": func() (*schema.Provider, error) {
+				return provider.Provider(), nil
+			},
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: datasourceConfig,
+				Check:  resource.TestCheckOutput("latest_version", "2.1.1"),
```
tomazpu, line 42: Can we hardcode this?

datasources/extensions/latest_version/data_source.go
```diff
@@ -0,0 +1,118 @@
...
+			// skip entries that are not valid semver
+			continue
+		}
+		if latestVersion == nil || v.GreaterThan(latestVersion) {
+			latestVersion = v
+		}
+	}
+	if latestVersion == nil {
+		d.SetId("")
+		return nil
```
tomazpu, line 109: If there is no version installed, we don't error, so we don't block the flow. But should we log something?

datasources/extensions/latest_version/data_source_test.go
Kirdock: What do you mean?

datasources/extensions/latest_version/data_source.go
Kirdock: Yes and no.
If an extension is installed, there must be a version, so there must be a result. If an extension doen't exist, the API returns a 404.
On the other side, I aligned it with all the other data sources that we have.
Erroring out would be the better/correct approach here, but I decided to have aligned functionality across all data sources (even though this branch should never happen).
We can do a follow-up, though, to revisit all existing data sources.

Kirdock: I'll create a follow-up to address all usages
CA-19314

tomazpu (approved):

arthurpitman (approved):

PR #1090, opened 2026-04-29
fix: dynatrace_calculated_web_metric eventual errors

Kirdock:
#### **Why** this PR?
There are different places where eventual consistency errors can occur. Not only during Create but also during get, update and delete.

#### **What** has changed?
CRUD operations should be more consistent.

#### **How** does it do it?
By adding certain retry conditions for all CRUD operations

#### How is it **tested**?
Current tests should be less flaky.

More consistent usage of the `dynatrace_calculated_web_metric` resource

**Issue:** CA-19218

### DIFF
--- a/dynatrace/api/v1/config/metrics/calculated/web/service.go
+++ b/dynatrace/api/v1/config/metrics/calculated/web/service.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


dynatrace/api/v1/config/metrics/calculated/web/service.go
```diff
@@ -72,62 +85,54 @@ func (me *service) Validate(ctx context.Context, v *mysettings.CalculatedWebMetr
...
-			}
-		}
-		time.Sleep(2 * time.Second)
-	}
-
-	if err != nil {
-		return err
-	}
-	return nil
+	return retry.RetryContext(ctx, retrycommon.DurationUntilDeadlineOrDefault(ctx, defaultTimeout), func() *retry.RetryError {
```
tomazpu, line 132: We don't want to check after the delete if it's gone?

Kirdock: I would say, if the delete works (status 200), we can safely say that this is gone

tomazpu (approved):

PR #1099, opened 2026-05-11
test: use data source for process groups

Kirdock:
#### **Why** this PR?
To make tests more independent across different environments and not rely on a variable for a process group ID.

#### **What** has changed?
Instead of setting the process group ID via variable/env, a data source is now used. Previously, during the research, the wrong/deprecated data source was used for process groups that faded out after 3 days. The data source `dynatrace_entity` doesn't have this problem.

#### **How** does it do it?
By making use of the `dynatrace_entity` data source instead of using a variable that needs to be provided.
It doesn't matter which ID is taken. It's only important that it's a valid process group ID.

#### How is it **tested**?
Tests still pass that rely on a valid process group ID.

N/A (minor doc update)

**Issue:** CA-17769

### DIFF
--- a/dynatrace/api/builtin/alerting/connectivityalerts/testdata/terraform/example_a.tf
+++ b/dynatrace/api/builtin/alerting/connectivityalerts/testdata/terraform/example_a.tf
@@ -1,8 +1,8 @@
-variable "PROCESS_GROUP_ID" {
-  type = string
+data "dynatrace_entity" "process_group" {
+  entity_selector = "type(\"PROCESS_GROUP\")"
 }
 
 resource "dynatrace_connectivity_alerts" "alert" {
   connectivity_alerts = false
-  process_group_id    = var.PROCESS_GROUP_ID
+  process_group_id    = data.dynatrace_entity.process_group.id
 }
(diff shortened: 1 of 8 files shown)


dynatrace/api/builtin/alerting/connectivityalerts/testdata/terraform/example_a.tf
```diff
@@ -1,8 +1,8 @@
-variable "PROCESS_GROUP_ID" {
-  type = string
+data "dynatrace_entity" "process_group" {
+  entity_selector = "type(\"PROCESS_GROUP\")"
 }
 
 resource "dynatrace_connectivity_alerts" "alert" {
   connectivity_alerts = false
-  process_group_id    = var.PROCESS_GROUP_ID
+  process_group_id    = data.dynatrace_entity.process_group.id
```
d0weinberger, line 7: What exactly does the data source return? The first entity id it can get that fits the `entity_selector`?

Kirdock: yes, exactly

Laubi (approved):

d0weinberger (approved):

PR #1100, opened 2026-05-11
fix(dynatrace_http_monitor): inconsistent GET

Kirdock:
#### **Why** this PR?
Because of eventual consistency reasons, the GET endpoint doesn't return all the data. This leads to failing tests and also inconsistencies when applying this resource (Get after apply).

#### **What** has changed?
Get after create/update of this resource is more consistent.

#### **How** does it do it?
Added retry logic to better assume when an object is ready to be fetched.
Similar to https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/1095

Because of eventual consistency
- `tags` may not be returned.
- Get may return 404
- Even if Get returns non-404 and the expected tags, we can't rely on it because server nodes may not be synced.

That's why we need to
1. Retry on 404
2. Retry if tags are not present (if there should be any)
3. Retry several times, even if 1. and 2. succeeded. Server nodes may still be out of sync.

#### How is it **tested**?
Unit tests added.
E2E tests should be more consistent (less flaky)

Get after create/update should not lead to a non-empty plan anymore

**Issue:** CA-19438

### DIFF
--- a/dynatrace/api/v1/config/synthetic/monitors/http/service.go
+++ b/dynatrace/api/v1/config/synthetic/monitors/http/service.go
@@ -34,6 +34,10 @@ import (
 const SchemaID = "v1:synthetic:monitors:http"
 const BasePath = "/api/v1/synthetic/monitors"
 
+const DesiredReadySuccesses = 5
+
+var ErrConsistencyRetry = errors.New("eventual consistency check")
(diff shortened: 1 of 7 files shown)


Kirdock:
Won't do

PR #1103, opened 2026-05-13
feat: dynatrace process grouping rules resource

Kirdock:
#### **Why** this PR?
The resources dynatrace_process_group_simple_detection, dynatrace_declarative_grouping and dynatrace_process_group_detection are deprecated and merged into a new settings schema. Therefore, we need to support the new one.

#### **What** has changed?
New resource added `dynatrace_process_grouping_rules` that supports the new settings schema.
Marked the other resources as deprecated and added the migration guide (environment-only side)

#### **How** does it do it?
Creating the resource based on the settings schema.
Also adding insertAfter to it, as this was also supported for the old resources.

#### How is it **tested**?
New integration tests added.

They can migrate their old process grouping rule resources and use the new one.

**Issue:** CA-19451


dynatrace/api/builtin/processgroupingrules/settings/settings.go
```diff
@@ -0,0 +1,90 @@
...
+	return map[string]*schema.Schema{
+		"custom_technology_name": {
+			Type:        schema.TypeString,
+			Description: "Note: Reported only in full-stack, infrastructure and discovery modes.",
+			Optional:    true, // nullable
+		},
+		"enabled": {
+			Type:        schema.TypeBool,
+			Description: "This setting is enabled (`true`) or disabled (`false`)",
+			Required:    true,
```
d0weinberger, line 47: Since it's not nullable, but the default value is set (to `true`), wouldn't that mean Optional: true, Default: true?

Kirdock: I would say we completely ignore default values in the schema. They are usually not correct (e.g., don't fulfill constraints) and the defaults are only used for migration purposes of settings, not for filling them if not given.

Also, this would be another manual adjustment that would lead to maintenance.

dynatrace/api/builtin/processgroupingrules/settings/instance_id_source.go
```diff
@@ -0,0 +1,81 @@
...
+		"name":              me.Name,
+		"property":          me.Property,
+	})
+}
+
+func (me *InstanceIdSource) HandlePreconditions() error {
+	if (me.Name == nil) && (me.Property != nil && slices.Contains([]string{"ENVIRONMENT_VARIABLE", "JAVA_SYS_PROP"}, string(*me.Property))) {
+		return fmt.Errorf("'name' must be specified if 'property' is set to '%v'", me.Property)
+	}
+	if (me.Name != nil) && (me.Property != nil && !slices.Contains([]string{"ENVIRONMENT_VARIABLE", "JAVA_SYS_PROP"}, string(*me.Property))) {
```
d0weinberger: I think I messed up a case with the preconditions. If `property` is null, we actually would not want `name` to be set, right?

Kirdock: yes
The name is only set if
```json
"precondition": {
	"expectedValues": [
		"ENVIRONMENT_VARIABLE",
		"JAVA_SYS_PROP"
	],
	"property": "property",
	"type": "IN"
}
```
Meaning, empty/null should lead to an error

Kirdock: Updated 👍

d0weinberger (approved):

PR #1105, opened 2026-05-18
fix(dynatrace_metric_events): drop stale entity_filter on type switch

d0weinberger:
#### **Why** this PR?

Fixes an API rejection when switching `dynatrace_metric_events` `query_definition.type` from `METRIC_KEY` to `METRIC_SELECTOR`. An empty `entityFilter` that the provider auto-creates on the initial Create (when no `entity_filter` block is configured) survives in state because the schema's `DiffSuppressFunc` masks its removal — and is then shipped in the type-switch Update payload,  which the API rejects.
 
#### **What** has changed and **How** does it do it?

The fix normalises an empty `EntityFilter` to nil at the top of `UnmarshalHCL`, before the existing auto-create. 
`METRIC_KEY` without an `entity_filter` block keeps working; `METRIC_SELECTOR` cleanly omits `entityFilter`. 
Previously the only workaround was destroying and recreating the resource. 

#### How is it **tested**?

Introduced a lifecycle test scaffolding (typed mock service capturing Create/Update payloads) that future bug regression tests for this resource can reuse. 

 - [x] `go test -tags=unit ./dynatrace/api/builtin/anomalydetection/metricevents/...` passes. 
 - [x] Verified the new test fails with the fix reverted (captured Update payload contains `&EntityFilter{}`). 
 - [x] `TF_ACC` acceptance suite for `dynatrace_metric_events` runs green against a tenant. 
 - [x] Manual smoke test: create with `type = METRIC_KEY` and no `entity_filter` block, edit to `type = METRIC_SELECTOR`, apply. 

Users will not really notice a change, other than that switching the type of a `dynatrace_metric_events` resource from `METRIC_KEY` to `METRIC_SELECTOR` works now.

**Issue:**
CA-19482 (sub-issue)

### DIFF
--- a/dynatrace/api/builtin/anomalydetection/metricevents/resource_unit_test.go
+++ b/dynatrace/api/builtin/anomalydetection/metricevents/resource_unit_test.go
@@ -1,0 +1,155 @@
+//go:build unit
+
+/**
+* @license
+* Copyright 2026 Dynatrace LLC
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+ */
+
+package metricevents_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api"
+	metricevents "github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/builtin/anomalydetection/metricevents/settings"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/export"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/rest"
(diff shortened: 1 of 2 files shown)


dynatrace/api/builtin/anomalydetection/metricevents/resource_unit_test.go
Kirdock, line 1: So this basically mocks the `service`, right?


[nit (no need to adjust)]: something like this would also be sufficient, right?
```go
val := &metricevents.Settings{}
err := val.UnmarshalHCL(testing2.MockDecoder{Elements: map[string]any{
	"enabled": true,
	"summary": "test event",
	"event_template": []any{
		map[string]any{
			"title":       "title",
			"description": "description",
			"event_type":  "CUSTOM_ALERT",
		},
	},
	"model_properties": []any{
		map[string]any{
			"alert_condition":    "ABOVE",
			"alert_on_no_data":   false,
			"dealerting_samples": 5,
			"samples":            5,
			"violating_samples":  3,
			"type":               "STATIC_THRESHOLD",
			"threshold":          0.0,
		},
	},
	"query_definition": []any{
		map[string]any{
			"type":            "METRIC_SELECTOR",
			"metric_selector": "builtin:host.cpu.usage:avg",
			// Empty entity_filter block — what survives in `d` when the
			// DiffSuppressFunc masks the removal of an auto-created filter.
			"entity_filter": []any{
				map[string]any{},
			},
		},
	},
}})
require.NoError(t, err)

assert.Nil(t, val.QueryDefinition.EntityFilter, "empty entity_filter must be dropped when type=METRIC_SELECTOR; otherwise the API rejects the payload")
```
On the other hand, the MockDecoder hasn't implemented decodeAll

Kirdock (approved):

PR #1107, opened 2026-05-19
fix: export workflow task positions

Kirdock:
#### **Why** this PR?
Task positions weren't exported

#### **What** has changed?
Task positions are exported again.

#### **How** does it do it?
By now removing deCompute for positions for export.

#### How is it **tested**?
Unit tests added

They can export task positions again

**Issue:** CA-19515

### DIFF
--- a/dynatrace/api/automation/workflows/service.go
+++ b/dynatrace/api/automation/workflows/service.go
@@ -70,16 +70,26 @@ func (me *service) Get(ctx context.Context, id string, v *workflows.Workflow) er
 // If the position is set by the user in the Terraform file, then it's correctly compared to the API set one.
 // DiffSuppress can't help here because even if the task doesn't change, the bug of TypeSet happens, which adds a new empty item which leads to a non-empty plan
 func deComputeTaskPosition(ctx context.Context, v *workflows.Workflow) {
-	positionTaskNames := map[string]bool{}
-	if stateConfig, ok := ctx.Value(settings.ContextKeyStateConfig).(*workflows.Workflow); ok {
-		for _, st := range stateConfig.Tasks {
-			if st.Position != nil {
-				positionTaskNames[st.Name] = true
-			}
-		}
+	// We only remove the position of the API task if the related task in the state and doesn't have positions
+	// We keep the permisisons if:
+	// - the task is in the state and has a position (we assume that the user set it, so we keep it). Drift detection works.
+	// - the task is not in the state
+	//   - Either because we removed one or on the API side something was added, and now it needs to be removed (showing with posision). Drift detection doesn't matter much, the task will be removed.
+	//   - Or if an import block with generate resource is used (settings.ContextKeyStateConfig is set but empty). Not in state => generate with position. No drift detection, because generate.
+
+	stateConfig, ok := ctx.Value(settings.ContextKeyStateConfig).(*workflows.Workflow)
+	if !ok {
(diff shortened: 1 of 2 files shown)


dynatrace/api/automation/workflows/service_unit_test.go
```diff
@@ -0,0 +1,385 @@
...
+					Name:     "task1",
+					Position: &workflowSettings.TaskPosition{X: 10, Y: 20},
+					Action:   "namespace:type",
+					Input:    map[string]any{},
+				},
+			},
+		}, v)
+	})
+
+	t.Run("Keeps task positions when no state config is present (export)", func(t *testing.T) {
```
arthurpitman, line 143: Should we add a test case for when the task in the state has no position and thus the position in value should be removed?

dynatrace/api/automation/workflows/service.go
```diff
@@ -70,16 +84,19 @@ func (me *service) Get(ctx context.Context, id string, v *workflows.Workflow) er
...
-				positionTaskNames[st.Name] = true
-			}
-		}
+	for _, st := range stateConfig.Tasks {
+		positionTaskNames[st.Name] = st.Position != nil
 	}
+
 	for _, task := range v.Tasks {
-		if !positionTaskNames[task.Name] {
+		if hasPosition, ok := positionTaskNames[task.Name]; ok && !hasPosition {
```
arthurpitman, line 106: what should happen to task in the value that is not in the state? Or is it OK, because once it is in the state it will be considered in subsequent gets?

arthurpitman: If it is OK, should we add a test case to assert that position is maintained?

arthurpitman: Resolved in conversation

d0weinberger: >what should happen to task in the value that is not in the state?

The `!ok` case, right? I was also again wondering reading over this line. Maybe it would be worth to add a comment why it is not necessary to handle that case?

Kirdock: To sum this a bit up:
- Cause: "position" inside a task may not be set. This is optional and intended.
- We can't really set something (Positions) inside a TypeSet (Tasks) to `Computed: true`, because this would lead to a TypeSet-hash change (because it's an unknown value), and then this is going to show a non-empty Terraform plan each and every time.
- We are removing the positions from the API response, but only if the task that we have locally doesn't have a position. This means that the position (locally) of a task is correctly validated/updated if set and is not validated (with a diff) at all if not set

With our export functionality, we don't have a state and we want to export the whole API response, with the positions. Therefore, we aren't setting the API position value to "nil".

If there are new tasks, added via the API, that are not present in the local state, we don't remove the position.
The behavior of this is that Terraform plan will show a task (with positions) that needs to be removed, because the API state doesn't match local.

This is also needed for an import block with generate resource. While using an import block, the state is not nil (key is given). An import block will also add the positions, because there isn't anything in the state, therefore the positions should not be removed.

dynatrace/api/automation/workflows/service_unit_test.go
Kirdock: Added two tests
- `Keeps task positions if the API response task does not match a local one`
- `Removes task position when matching state task does not have a position`

d0weinberger (approved):

PR #1110, opened 2026-05-20
fix(export): computed nested fields weren't exported

Kirdock:
#### **Why** this PR?
Nested computed fields like in workflows (trigger->schedule->timezone) aren't exported. They are missing in the generated resource.

#### **What** has changed?
Nested exported fields are exported.

#### **How** does it do it?
By adding the same functionality for writing computed fields as on the root level.
The cause is that the parent export implementation in hclgen.go isn't aligned with the child (resoruce_entry.go). On root level computed fields are written but not if the computed field is in a child. Aligning the logic in resource_entry.go with hclgen.go

#### How is it **tested**?
Manual tests

They can export nested computed fields.

**Issue:** CA-19536

### DIFF
--- a/terraform/hclgen/hclgen.go
+++ b/terraform/hclgen/hclgen.go
@@ -423,27 +423,24 @@ func (me *HCLGen) export(m map[string]any, schema map[string]*schema.Schema, w i
 			resourceName,
 		},
 	)
-	body := bs.Body()
-	for _, entry := range ents {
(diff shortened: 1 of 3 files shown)


d0weinberger (commented):
LGTM, but could you add some test cases?

tomazpu (approved):

d0weinberger (approved):

PR #1115, opened 2026-06-01
feat: implement Validate on settings20 service

d0weinberger:
#### **Why** this PR?
The settings20 service's `Validate` method was previously a no-op stub with a comment saying Settings 2.0 doesn't offer validation - that is no longer true, and the stub would silently swallow the call.

#### **What** has changed?
`settings20.service[T].Validate` is now a real implementation instead of a no-op. The method signature has also been corrected to match the existing `settings.Validator[T]` interface, which already required a `context.Context` parameter that the stub was missing.

#### **How** does it do it?
POSTs the settings payload to `/api/v2/settings/objects?validateOnly=true`. The endpoint validates the payload against the schema and runs synchronous validators without persisting any object.

#### How is it **tested**?
A new integration test (`dynatrace/api/builtin/rum/hostheaders/validate_test.go`) lists all `builtin:rum.host-headers` objects after calling Validate, then asserts that the host header for which we called `Validate` is **not** in the list. `builtin:rum.host-headers` was chosen as the test vehicle because it has no custom validators, no side-effects, and a single plain-text field. A `t.Cleanup` block deletes any accidentally created objects to keep the environment clean if the assertion ever fails.

No user-facing change. `Validate` is an internal service method not exposed through any Terraform resource or data source.

**Issue:** CA-19461

### DIFF
--- a/dynatrace/api/builtin/rum/hostheaders/validate_test.go
+++ b/dynatrace/api/builtin/rum/hostheaders/validate_test.go
@@ -1,0 +1,82 @@
+//go:build integration
+
+/**
+* @license
+* Copyright 2026 Dynatrace LLC
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+ */
+
+package hostheaders_test
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"testing"
+
+	hostheadersvc "github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/builtin/rum/hostheaders"
+	hostheaders "github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/builtin/rum/hostheaders/settings"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/rest"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/settings"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/settings/services/settings20"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


dynatrace/settings/services/settings20/service.go
```diff
@@ -344,8 +341,25 @@ func (me *service[T]) List(ctx context.Context) (api.Stubs, error) {
...
+	}
+	if insertAfter := settings.GetInsertAfter(v); insertAfter != nil {
+		soc.InsertAfter = *insertAfter
+	}
+
+	var req rest.Request
+	if me.skipRepairInput() {
+		req = me.client.Post(ctx, "/api/v2/settings/objects?validateOnly=true", []SettingsObjectCreate{soc}).Expect(200)
+	} else {
+		req = me.client.Post(ctx, "/api/v2/settings/objects?validateOnly=true&repairInput=true", []SettingsObjectCreate{soc}).Expect(200)
```
Laubi: [nit] I'd build the url in one place, and just add the query parameter if the skip is not set. One request instead of basically a duplicated line.

dynatrace/settings/services/settings20/service.go
```diff
@@ -344,8 +341,25 @@ func (me *service[T]) List(ctx context.Context) (api.Stubs, error) {
 	return stubs, nil
 }
 
-func (me *service[T]) Validate(v T) error {
-	return nil // Settings 2.0 doesn't offer validation
+func (me *service[T]) Validate(ctx context.Context, v T) error {
```
Laubi, line 347: Please add a brief doc about the behavior of `Validate`. Can be a followup

dynatrace/api/builtin/rum/hostheaders/validate_test.go
Kirdock, line 1: Some things regarding this
1. This one has a huge potential to run into conflicts with parallel pipeline executions. If this one (pipeline A) runs and pipeline B creates a host header, it can be deleted and pipeline B shows a non-empty plan (needs creation). It can also happen that pipeline B is creating a setting and then the test fails with `Validate must not create any settings objects`. I would say we don't need to validate that the object is not created
2. Please use `t.Context` instead of `context.Background`

Laubi (dismissed):
lg from my side, @Kirdock comments could be a followup for me

dynatrace/api/builtin/rum/hostheaders/validate_test.go
d0weinberger: Regarding 1: I changed the assertion. A random name is chosen for the header. If it pops up in the `List` afterwards, we assume it was created, leading to the test failing.

Regarding 2: Yes, but for the cleanup step I keep `context.Background`, because by the time it comes to cleanup, `t.Context` will already be cancelled.

Kirdock (dismissed):

Kirdock (approved):

PR #1116, opened 2026-06-01
feat: resource updates for 338

Kirdock:
#### **Why** this PR?
Some APIs have changed with Dynatrace version 338. Therefore, the Terraform resources also need updating.

#### **What** has changed?
New fields, descriptions, preconditions.

#### **How** does it do it?
Look into the commits for all the changes.

#### How is it **tested**?
Added new fields if possible.
For OpenPipeline I wasn't able to find a single schema that supports the new geo_lookup field (at least the API didn't return any `geoLookup` in the processors array.

**Issue:** CA-19449


dynatrace/api/builtin/failuredetection/environment/rules/settings/predicate.go
```diff
@@ -90,15 +93,19 @@ func (me *Predicate) MarshalHCL(properties hcl.Properties) error {
...
-func (me *Predicate) HandlePreconditions() {
-	if me.CaseSensitive == nil && slices.Contains([]string{"STRING_EQUALS", "STARTS_WITH", "ENDS_WITH", "CONTAINS"}, string(me.PredicateType)) {
+func (me *Predicate) HandlePreconditions() error {
+	if (me.CaseSensitive == nil) && (slices.Contains([]string{"STRING_EQUALS", "STARTS_WITH", "ENDS_WITH", "CONTAINS"}, string(me.PredicateType))) {
 		me.CaseSensitive = new(false)
 	}
+	if (me.CaseSensitive != nil) && (!slices.Contains([]string{"STRING_EQUALS", "STARTS_WITH", "ENDS_WITH", "CONTAINS"}, string(me.PredicateType))) {
+		return fmt.Errorf("'case_sensitive' must not be specified unless 'predicate_type' is one of ['STRING_EQUALS', 'STARTS_WITH', 'ENDS_WITH', 'CONTAINS']; got 'predicate_type'='%v'", me.PredicateType)
+	}
 	// ---- ManagementZones []string -> {"expectedValues":["MANAGEMENT_ZONES_CONTAINS_ALL"],"property":"predicateType","type":"IN"}
```
d0weinberger, line 103: I think we would support these preconditions now, wouldn't we?

dynatrace/api/builtin/hyperscalerauthentication/connections/aws/settings/aws_role_based_authentication_config.go
```diff
@@ -37,7 +37,7 @@ func (me *AwsRoleBasedAuthenticationConfig) Schema() map[string]*schema.Schema {
 		// },
 		"consumers": {
 			Type:        schema.TypeSet,
-			Description: "Dynatrace integrations that can use this connection. Possible values: `APP:dynatrace.biz.carbon`, `DA`, `NONE`, `SVC:com.dynatrace.bo`, `SVC:com.dynatrace.da`, `SVC:com.dynatrace.grail`, `SVC:com.dynatrace.openpipeline`",
```
d0weinberger, line 40: Huh, interesting. What did they do about existing configs having this app as consumer? Or were there none?

dynatrace/api/builtin/failuredetection/environment/rules/settings/predicate.go
Kirdock: No, as this affects a list and not a basic type, which we don't support yet

dynatrace/api/builtin/hyperscalerauthentication/connections/aws/settings/aws_role_based_authentication_config.go
Kirdock: Not sure. This is about the settings migrator. It would use/migrate the set default I guess
```json
"default": [
  "SVC:com.dynatrace.da"
],
```

dynatrace/api/builtin/openpipeline/bizevents/ingestsources/settings/davis_event_property.go
```diff
@@ -45,35 +48,53 @@ func (me *DavisEventProperties) UnmarshalHCL(decoder hcl.Decoder) error {
...
-		"value": me.Value,
+		"key":      me.Key,
+		"strategy": me.Strategy,
+		"value":    me.Value,
 	})
 }
 
+func (me *DavisEventProperty) HandlePreconditions() error {
+	if (me.Value == nil) && ((me.Strategy != nil && string(*me.Strategy) == "equals") || (me.Strategy == nil)) {
+		me.Value = new("")
```
d0weinberger, line 86: Is an empty string valid here? Why would we allow it? We wouldn't break anyone being restrictive, as it was previously a required field anyways

dynatrace/api/builtin/openpipeline/bizevents/ingestsources/settings/field_extraction_entry.go
```diff
@@ -105,24 +111,45 @@ func (me *FieldExtractionEntry) MarshalHCL(properties hcl.Properties) error {
...
 		"extraction_type":        me.ExtractionType,
 		"source_field_name":      me.SourceFieldName,
+		"strategy":               me.Strategy,
 	})
 }
 
 func (me *FieldExtractionEntry) HandlePreconditions() error {
-	if (me.ConstantFieldName == nil) && (string(me.ExtractionType) == "constant") {
+	empty := FieldExtractionEntry{}
+	if *me == empty {
```
d0weinberger, line 120: How does this happen anyway?

dynatrace/api/builtin/openpipeline/bizevents/ingestsources/settings/davis_event_property.go
Kirdock: Previously, the value was required and empty string was valid (type string without omitempty) at least from our side.
Again, it's empty string as we can't tell via plugin SDK if it's not set or set to empty string. Therefore, we set it to empty string, but only if the value can be provided

dynatrace/api/builtin/openpipeline/bizevents/ingestsources/settings/field_extraction_entry.go
Kirdock: Because of the TypeSet bug of the plugin SDK, which forwards an empty item.

dynatrace/api/builtin/failuredetection/environment/parameters/settings/compare_operation.go
```diff
@@ -71,17 +74,32 @@ func (me *CompareOperation) MarshalHCL(properties hcl.Properties) error {
...
+	}
+	if (me.TextValue != nil) && (!slices.Contains([]string{"STRING_EQUALS", "NOT_STRING_EQUALS", "STARTS_WITH", "NOT_STARTS_WITH", "CONTAINS", "NOT_CONTAINS", "ENDS_WITH", "NOT_ENDS_WITH"}, string(me.CompareOperationType))) {
+		return fmt.Errorf("'text_value' must not be specified unless 'compare_operation_type' is one of ['STRING_EQUALS', 'NOT_STRING_EQUALS', 'STARTS_WITH', 'NOT_STARTS_WITH', 'CONTAINS', 'NOT_CONTAINS', 'ENDS_WITH', 'NOT_ENDS_WITH']; got 'compare_operation_type'='%v'", me.CompareOperationType)
+	}
+	if (me.TextValue == nil) && (slices.Contains([]string{"STRING_EQUALS", "NOT_STRING_EQUALS", "STARTS_WITH", "NOT_STARTS_WITH", "CONTAINS", "NOT_CONTAINS", "ENDS_WITH", "NOT_ENDS_WITH"}, string(me.CompareOperationType))) {
+		return fmt.Errorf("'text_value' must be specified when 'compare_operation_type' is one of ['STRING_EQUALS', 'NOT_STRING_EQUALS', 'STARTS_WITH', 'NOT_STARTS_WITH', 'CONTAINS', 'NOT_CONTAINS', 'ENDS_WITH', 'NOT_ENDS_WITH']; got 'compare_operation_type'='%v'", me.CompareOperationType)
```
d0weinberger, line 100: This is now a breaking change. Before this PR, we would just set an empty string for textValue

Kirdock: Yes and no. We don't do validation but the schema actually defines min length of 1. That's why we don't set it to empty string here if the precondition is valid

dynatrace/api/builtin/openpipeline/davis/events/ingestsources/settings/geo_lookup_attributes.go
```diff
@@ -0,0 +1,66 @@
+/**
+* @license
+* Copyright 2025 Dynatrace LLC
```
d0weinberger: 2026, actually. [nit]

Kirdock: Damn, it was found 😅. Wanted to keep the maintenance low. Can update it

Kirdock: done 👍

d0weinberger (approved):

PR #1117, opened 2026-06-01
feat: add dynatrace_gcp_connection resource

d0weinberger:
#### **Why** this PR?
Dynatrace supports connecting to GCP environments via Service Account Impersonation for integrations such as Data Acquisition, OpenPipeline, and Business Observability. This PR exposes that capability as a Terraform resource so users can manage GCP connections as code alongside the rest of their Dynatrace configuration.

#### **What** has changed?
A new dynatrace_gcp_connection resource backed by the builtin:hyperscaler-authentication.connections.gcp Settings 2.0 schema. The resource is registered in the provider and the export descriptor, placing it alongside the existing AWS and Azure connection resources.

#### **How** does it do it?
The resource follows the same pattern as `dynatrace_aws_connection` and `dynatrace_azure_connection`: a thin service wrapper around `settings20.Service` with typed settings structs, HCL marshaling, and a HandlePreconditions guard that enforces the `type == service_account_impersonation` requirement.

#### How is it **tested**?
PR includes a GCP e2e acceptance test that provisions a service account and IAM binding before creating the connection. The test is skipped for now. It will be enabled in a follow-up.

For now, nothing. The resource is not wired in `provider.go` and cannot yet be used. This will happen in a follow-up, tying all pieces together.

**Issue:** CA-19463


dynatrace/api/builtin/hyperscalerauthentication/connections/gcp/testdata/terraform/e2e_test.tf
```diff
@@ -0,0 +1,48 @@
...
+
+# Grant DT GCP Principal access to the service account
+resource "google_service_account_iam_member" "wif_binding" {
+  service_account_id = google_service_account.gcp_service_account.name
+  member  = "serviceAccount:${data.dynatrace_gcp_principal.principal.principal}"
+  role    = "roles/iam.serviceAccountTokenCreator"
+}
+
+# Wait for IAM binding to propagate
+resource time_sleep "wait_for_iam_binding" {
```
d0weinberger, line 29: This is up for debate. Do we want to retry in the resource or do we want the users to take this into their own hands? I would argue for the latter.

Kirdock: We already added a retry mechanism for AWS, so I would vote for keeping it consistent and also add it here (can be a follow up, though)

Kirdock (approved):

Kirdock:
Can't really test it but should be fine, I guess 👍

Laubi (approved):

PR #1120, opened 2026-06-02
feat: resource updates for 339

Kirdock:
#### **Why** this PR?
Some APIs have changed with Dynatrace version 339. Therefore, the Terraform resources also need updating.

#### **What** has changed?
New fields, descriptions, preconditions.

#### **How** does it do it?
Look into the commits for all the changes.

#### How is it **tested**?
Added new fields.

**Issue:** CA-19449


dynatrace/api/builtin/anomalydetection/metricevents/settings/model_properties.go
```diff
@@ -99,14 +101,23 @@ func (me *ModelProperties) MarshalHCL(properties hcl.Properties) error {
...
+		return fmt.Errorf("'signal_fluctuation' must not be specified unless 'type' is set to 'AUTO_ADAPTIVE_THRESHOLD'; got 'type'='%v'", me.Type)
+	}
+	if (me.Threshold != nil) && (string(me.Type) != "STATIC_THRESHOLD") {
+		return fmt.Errorf("'threshold' must not be specified unless 'type' is set to 'STATIC_THRESHOLD'; got 'type'='%v'", me.Type)
+	}
+	if (me.Tolerance != nil) && (string(me.Type) != "SEASONAL_BASELINE") {
+		return fmt.Errorf("'tolerance' must not be specified unless 'type' is set to 'SEASONAL_BASELINE'; got 'type'='%v'", me.Type)
+	}
+	if (me.Tolerance == nil) && (string(me.Type) == "SEASONAL_BASELINE") {
+		return fmt.Errorf("'tolerance' must be specified when 'type' is set to 'SEASONAL_BASELINE'; got 'type'='%v'", me.Type)
```
d0weinberger, line 120: That's what the schema says, but previously we had just set it to 0.0 in case it was not there. With the stricter condition, we might break existing configs, right? Don't we set a default because of the constraint in the schema? Also, interesting, but old: There are other defaults defined in the schema than we take here. I'm sure there is a reason, but it always makes me wonder.

Kirdock: Yes, every added precondition is a breaking change and the schema actually defines a minimum of `0.1`, so `0.0` was wrong in the first place

d0weinberger: The precondition was not added now, in fact there isn't really a change in the schema. This is why I am pointing it out, because here we cannot say "The schema changed". It seems to have been a conscious decision, to set Tolerance to 0.0. But you are right, 0.0 would have been wrong anyways.

d0weinberger (approved):

PR #1123, opened 2026-06-02
feat: add dynatrace_gcp_principal data source

d0weinberger:
#### Why this PR?

The `dynatrace_gcp_principal` data source is a prerequisite for setting up GCP integrations: users need the
Dynatrace-managed GCP service account identity before they can grant it IAM permissions in their GCP project.
Without a dedicated data source, there is no way to retrieve this value via Terraform.

#### What has changed?

- New package datasources/hyperscalerauthentication/connections/gcpdynatraceprincipal containing the data source implementation, its internal settings type, and the API schema snapshot (schema.json).

#### How does it do it?

The data source calls the Settings 2.0 list endpoint for `builtin:hyperscaler-authentication.connections.gcp-dynatrace-principal`.
If no principal exists yet, it triggers creation as a side effect by submitting a dummy GCP connection payload (`ValidConnection`)
to the validate endpoint — which causes Dynatrace to provision the principal asynchronously.
It then polls the list endpoint every 5 seconds (up to 5 minutes) until the principal appears,
and returns its service account email.

#### How is it tested?

 - `TestAccDatasourceGcpPrincipal` (integration): runs a full Terraform plan/apply against a real environment and asserts the output matches the GCP service account email pattern (iam.gserviceaccount.com$).
 - `TestValidateIsSuccessfulForGcpConnection` (integration): submits `ValidConnection` directly to the validate endpoint, guarding against schema drift that would silently break principal creation.
 - Both tests are gated behind the `integration` build tag and require `DYNATRACE_ENV_URL` / `DYNATRACE_API_TOKEN`.

The tests are currently skipped, as neither the connection resource nor the principal datasource are wired in the provider yet. Enabling the tests will happen in the same pull request where the wiring happens, i.e. with the release of the feature.

Currently, it does _not_ affect users as the data source is not yet registered in the provider. This will happen in a follow-up.
Once the data source is registered, users can reference the Dynatrace GCP Principal in their Terraform configuration:

```
data "dynatrace_gcp_principal" "principal" {}

# e.g. grant it a role in GCP

resource "google_service_account_iam_member" "binding" {
  service_account_id = "..."
  role = "roles/iam.serviceAccountTokenCreator"
  member = "serviceAccount:${data.dynatrace_gcp_principal.principal.principal}"
}
```
The first apply may take a while if the principal does not yet exist in the environment.

**Issue:** CA-19461

### DIFF
--- a/datasources/hyperscalerauthentication/connections/gcpdynatraceprincipal/data_source.go
+++ b/datasources/hyperscalerauthentication/connections/gcpdynatraceprincipal/data_source.go
@@ -1,0 +1,136 @@
+/**
+* @license
+* Copyright 2020 Dynatrace LLC
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+ */
+
+package gcpdynatraceprincipal
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	gcpdynatraceprincipal "github.com/dynatrace-oss/terraform-provider-dynatrace/datasources/hyperscalerauthentication/connections/gcpdynatraceprincipal/settings"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/builtin/hyperscalerauthentication/connections/gcp"
+	gcpsettings "github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/builtin/hyperscalerauthentication/connections/gcp/settings"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/rest"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/settings"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/settings/services/settings20"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider/config"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider/logging"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+const (
+	SchemaVersion = "0.0.3"
+	SchemaID      = "builtin:hyperscaler-authentication.connections.gcp-dynatrace-principal"
+
+	principalCreationTimeout = 5 * time.Minute
+)
+
+var errPrincipalNotProvisionedYet = errors.New("no Dynatrace GCP Principal found yet")
+
+var ValidConnection = gcpsettings.Settings{
+	Name: "temp",
+	Type: "serviceAccountImpersonation",
+	ServiceAccountImpersonation: new(gcpsettings.ServiceAccountImpersonation{
+		Consumers: []gcpsettings.ConsumersOfServiceAccountImpersonation{"SVC:com.dynatrace.bo"},
+	}),
+}
+
+func ensurePrincipalExists(ctx context.Context, credentials *rest.Credentials) (*api.Stub, error) {
+	listService := settings20.Service[*gcpdynatraceprincipal.Settings](credentials, SchemaID, SchemaVersion)
+
+	stubs, err := listService.List(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(stubs) == 0 {
+		if v, ok := gcp.Service(credentials).(settings.Validator[*gcpsettings.Settings]); ok {
+			if err = v.Validate(ctx, &ValidConnection); err != nil {
+				return nil, fmt.Errorf("failed to trigger Dynatrace GCP Principal creation. Please reach out to support: %w", err)
+			}
+		}
+
+		err = retry.RetryContext(ctx, principalCreationTimeout, func() *retry.RetryError {
+			var listErr error
+			stubs, listErr = listService.List(ctx)
+			if listErr != nil {
+				return retry.NonRetryableError(fmt.Errorf("Dynatrace GCP Principal provisioning failed. Please reach out to support: %w", listErr))
+			}
+			if len(stubs) == 0 {
+				return retry.RetryableError(errPrincipalNotProvisionedYet)
+			}
+			return nil
+		})
+		if err != nil {
+			if errors.Is(err, errPrincipalNotProvisionedYet) {
+				return nil, fmt.Errorf("Dynatrace GCP Principal was not provisioned within %v. Please reach out to support", principalCreationTimeout)
+			}
+			return nil, err
+		}
+	}
+
+	// There can only be one principal, so we take the first (and only one) in the list.
+	return stubs[0], nil
+}
+
+// Attribute keys
+const (
+	attrPrincipal = "principal"
+)
+
+func DataSource() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: logging.EnableDSCtx(DataSourceRead),
+		Description: "Returns the Dynatrace GCP Principal",
+		Schema: map[string]*schema.Schema{
+			attrPrincipal: {
+				Type:        schema.TypeString,
+				Description: "Dynatrace GCP Principal",
+				Computed:    true,
+			},
+		},
+	}
+}
+
+func DataSourceRead(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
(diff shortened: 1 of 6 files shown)


datasources/hyperscalerauthentication/connections/gcpdynatraceprincipal/data_source.go
```diff
@@ -0,0 +1,134 @@
...
+			}
+			if len(stubs) == 0 {
+				select {
+				case <-time.After(principalCreationPollInterval):
+				case <-ctx.Done():
+					return retry.NonRetryableError(ctx.Err())
+				}
+				return retry.RetryableError(fmt.Errorf("no Dynatrace GCP Principal found yet"))
+			}
+			return nil
```
Kirdock, line 81: Doesn't the retryContext function take care of the polling? It also has some exponential backoff as far as I know.

```suggestion
			if len(stubs) == 0 {
				return retry.RetryableError(fmt.Errorf("no Dynatrace GCP Principal found yet"))
			}
			return nil
```

datasources/hyperscalerauthentication/connections/gcpdynatraceprincipal/data_source.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+const (
+	SchemaVersion = "0.0.3"
+	SchemaID      = "builtin:hyperscaler-authentication.connections.gcp-dynatrace-principal"
+
+	principalCreationTimeout      = 5 * time.Minute
```
d0weinberger: timeout and intervals are up for debate, I would say. Normally, the principal is created almost instantly.

datasources/hyperscalerauthentication/connections/gcpdynatraceprincipal/data_source_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"regexp"
+	"testing"
+
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/testing/api"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func TestAccDatasourceGcpPrincipal(t *testing.T) {
```
d0weinberger, line 32: The real problem I see with this test is that after the first run, the principal will be created in our test environment. We will not remove it afterwards. So, the creation of the principal is never really tested again after that. Not sure how to mitigate this.

datasources/hyperscalerauthentication/connections/gcpdynatraceprincipal/data_source.go
```diff
@@ -0,0 +1,134 @@
...
+				case <-time.After(principalCreationPollInterval):
+				case <-ctx.Done():
+					return retry.NonRetryableError(ctx.Err())
+				}
+				return retry.RetryableError(fmt.Errorf("no Dynatrace GCP Principal found yet"))
+			}
+			return nil
+		})
+		if err != nil {
+			return "", fmt.Errorf("no Dynatrace GCP Principal found. Please reach out to support: %w", err)
```
Kirdock: So eventually the error message could look like this, right?
`no Dynatrace GCP Principal found. Please reach out to support: no Dynatrace GCP Principal found yet`

datasources/hyperscalerauthentication/connections/gcpdynatraceprincipal/data_source.go
```diff
@@ -0,0 +1,134 @@
...
+	return &schema.Resource{
+		ReadContext: logging.EnableDSCtx(DataSourceRead),
+		Description: "Returns the Dynatrace GCP Principal",
+		Schema: map[string]*schema.Schema{
+			attrPrincipal: {
+				Type:        schema.TypeString,
+				Description: "Dynatrace GCP Principal",
+				Computed:    true,
+			},
+		},
```
Kirdock, line 110: Isn't that duplicated now with the Schema in `settings/settings.go`?
We can also keep it like this because of separation of concerns (settings fetch/response vs data source data)

datasources/hyperscalerauthentication/connections/gcpdynatraceprincipal/validate_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/settings/services/settings20"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Guards against schema drift in ValidConnection. The data source triggers Dynatrace GCP
+// Principal creation as a side effect of submitting ValidConnection to the validate endpoint;
+// if the payload is rejected, no principal will ever be created.
+func TestValidateIsSuccessfulForGcpConnection(t *testing.T) {
+	t.Skip("Enable this test as soon as the dynatrace_gcp_connection resource datasource is referenced in provider.go")
```
Kirdock: The test actually doesn't rely on Terraform, so we can execute/unskip it, right?
And shouldn't this be in the resource folder? This is actually not about the data source, right? On the other hand, this is more about the generic resource and not the GCP resource, so I'm fine with either.

In addition, should we add a cleanup for the principal or are we fine that it is created (also for TestAccDatasourceGcpPrincipal)?
Maybe worth adding it to the sequential_integration tests 🤔

datasources/hyperscalerauthentication/connections/gcpdynatraceprincipal/data_source_test.go
Kirdock: We could add a cleanup function where we delete the settings object afterward, like you did in your previous validate PR. Would this help? 🤔

datasources/hyperscalerauthentication/connections/gcpdynatraceprincipal/data_source.go
Kirdock: Then I would say 1 minute is enough as an upper limit, but no hard feelings for keeping 5 minutes

datasources/hyperscalerauthentication/connections/gcpdynatraceprincipal/validate_test.go
d0weinberger: You are right, I will unskip it.
I would like to keep it next to the data source, though, as it makes sure that exactly the specific payload we send to trigger principal creation is valid. 

I am against a clean up, though.

datasources/hyperscalerauthentication/connections/gcpdynatraceprincipal/data_source.go
d0weinberger: That looks a little silly, I agree

datasources/hyperscalerauthentication/connections/gcpdynatraceprincipal/data_source.go
d0weinberger: Hm, yes, it is duplicated. I actually changed the generated version, which didn't have `Computed`, but `ForceNew`, and `Required`, which makes sense if the principal is treated like a resource.

I would say: settings contains the part directly derived from the settings schema. data_source has its own structure, which makes sense for a data source.

datasources/hyperscalerauthentication/connections/gcpdynatraceprincipal/data_source.go
d0weinberger: I left it at 5 minutes for now. Too long doesn't hurt much. Too short does.

datasources/hyperscalerauthentication/connections/gcpdynatraceprincipal/data_source_test.go
d0weinberger: Hm, as discussed offline, currently not an option, I'm afraid.

datasources/hyperscalerauthentication/connections/gcpdynatraceprincipal/data_source.go
d0weinberger: Yes, I was not aware of that, but I changed it now. This makes the polling interval obsolete.

datasources/hyperscalerauthentication/connections/gcpdynatraceprincipal/data_source.go
d0weinberger: Handled. If we get a non-retryable error, we print it. If it's just the one retryable one, we don't.

datasources/hyperscalerauthentication/connections/gcpdynatraceprincipal/data_source.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	principal, err := ensurePrincipalExists(ctx, creds)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	d.SetId(principal)
```
Kirdock: Just thinking, we never need/use the objectId, right? Else we would have a problem or a breaking change or V2 in the future. But it also doesn't hurt to set it I would say

Kirdock (dismissed):

datasources/hyperscalerauthentication/connections/gcpdynatraceprincipal/data_source.go
d0weinberger: Changed it so use the settings object id here

Kirdock (approved):

PR #1125, opened 2026-06-03
chore: Exclude `.agents`, `.claude` and `.opencode` folders from source control

arthurpitman:
#### **Why** this PR?
While working with agents locally, I'd like to not have the configurations in source control.

`.agents`, `.claude` and `.opencode` folders are excluded from source control.

#### How is it **tested**?
NA

NA

**Issue:** NA

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,9 @@ test.log
 .DS_Store
 *.env
 __debug_bin*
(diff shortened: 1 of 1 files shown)


arthurpitman:
Closing this, as I'll re-work it

PR #1126, opened 2026-06-03
feat: add retry to GCP connection and make impersonation fields ForceNew

d0weinberger:
#### Why this PR?
- GCP service account impersonation depends on a user-provided service account ID whose IAM policy changes on the GCP side may not have propagated by the time the Dynatrace connection is created, causing the create to fail transiently. GCP
  connections previously had no retry on create (unlike Azure/AWS).
- Wrapping the GCP service in a struct (needed for the retry) would otherwise hide the underlying Validate capability that the GCP Principal data source relies on to trigger principal provisioning, so the wrapper must forward it.

#### What has changed?

  - dynatrace_gcp_connection create now retries while the API reports authentication failure.
  - New exported constant `RetryableAuthenticationErrorMessage` ("GCP authentication failed") in the gcp package.
  - Settings now expose a create `Timeouts()` with `DefaultCreateTimeout = 2 * time.Minute`.
  - `service_account_impersonation.service_account_id` is now `Required` + `ForceNew`; `consumers` is now `ForceNew`.
  - New `settings.ErrValidatorNotImplemented(schemaID)` helper, used by both the GCP service wrapper and the GCP Principal data source to replace an ad-hoc duplicated message.
  - New `envutils.DTGCPTestUnimpersonableServiceAccount` env var, a (skipped) acceptance test, its testdata-auth-failure/ config, and CI secret `DT_GCP_TEST_UNIMPERSONABLE_SERVICE_ACCOUNT` wiring in tests.yml.

#### How does it do it?

  - The plain delegating Service() is replaced by a wrapper whose Create runs through retry.RetryContext, retrying only on a 400 constraint violation carrying RetryableAuthenticationErrorMessage (via settings.ClassifyConstraintRetryError),
  bounded by the create timeout.
  - Update stays plain delegation: every impersonation field is ForceNew, so updates only ever rename the connection, which does not depend on IAM propagation.
  - The wrapper forwards Validate to the underlying service, falling back to ErrValidatorNotImplemented (an internal-provider-bug error that names the schema ID and points to support) if it ever does not implement validation.
  - The retry-trigger message and the test assertion share one constant, so an API wording change is caught.

#### How is it tested?

  - Acceptance test `TestAccGcpConnectionAuthenticationFailure` drives a connection with an unimpersonable, TF_VAR-injected service account ID and asserts via `ExpectError` that the API returns the message the retry keys on.
  - The test is currently `t.Skip`'d (and its CI secret pre-wired) pending the GCP connection resource being fully wired up in a follow-up PR; it carries a documented known-flake note about the retry budget coinciding with the context
  deadline.

####  How does it affect users?
  
 Currently, it doesn't, as `dynatrace_gcp_connection` is not yet wired up in the provider. Once it is, users will see the following changes:
 
  - `dynatrace_gcp_connection` create is now resilient to GCP IAM propagation delays, retrying for up to 2 minutes instead of failing immediately.
  - `service_account_id` is now required at creation and changing it (or consumers) forces resource replacement rather than an in-place update.

**Issue:** CA-19635

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -168,6 +168,7 @@ jobs:
           DT_CLIENT_ID: ${{ secrets.DT_CLIENT_ID }}
           DT_CLIENT_SECRET: ${{ secrets.DT_CLIENT_SECRET }}
           DT_ACCOUNT_ID: ${{ secrets.DT_ACCOUNT_ID }}
+          TF_VAR_DT_GCP_TEST_UNIMPERSONABLE_SERVICE_ACCOUNT: ${{ secrets.DT_GCP_TEST_UNIMPERSONABLE_SERVICE_ACCOUNT }}
         run: go test -tags=integration -v -p 1 ${{ matrix.config.path }}
 
   sequential-integration-tests:
--- a/dynatrace/api/builtin/hyperscalerauthentication/connections/gcp/service.go
+++ b/dynatrace/api/builtin/hyperscalerauthentication/connections/gcp/service.go
@@ -18,7 +18,13 @@
 package gcp
 
 import (
-	service "github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/builtin/hyperscalerauthentication/connections/gcp/settings"
+	"context"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
+
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api"
+	serviceSettings "github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/builtin/hyperscalerauthentication/connections/gcp/settings"
+	retrycommon "github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/builtin/hyperscalerauthentication/connections/retry"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/rest"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/settings"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/settings/services/settings20"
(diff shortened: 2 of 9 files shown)


.github/workflows/tests.yml
```diff
@@ -168,6 +168,7 @@ jobs:
           DT_CLIENT_ID: ${{ secrets.DT_CLIENT_ID }}
           DT_CLIENT_SECRET: ${{ secrets.DT_CLIENT_SECRET }}
           DT_ACCOUNT_ID: ${{ secrets.DT_ACCOUNT_ID }}
+          DT_GCP_TEST_UNIMPERSONABLE_SERVICE_ACCOUNT: ${{ secrets.DT_GCP_TEST_UNIMPERSONABLE_SERVICE_ACCOUNT }}
```
d0weinberger: Debate: I could also add a new tag, maybe `gcp_integration`, and add the secret there

dynatrace/api/builtin/hyperscalerauthentication/connections/gcp/service.go
Kirdock, line 1: We also need a Validate function here, right?

dynatrace/api/builtin/hyperscalerauthentication/connections/gcp/service_test.go
```diff
@@ -36,3 +42,52 @@ func TestGcpConnection(t *testing.T) {
...
+func TestAccGcpConnectionAuthenticationFailure(t *testing.T) {
+	t.Skip("Enable this test as soon as the GCP connection resource is wired.")
+
+	if !api.AccEnvsGiven(t) {
+		return
+	}
+
+	// Holds a service account ID that the Dynatrace Principal can never impersonate
+	const unimpersonableServiceAccountEnvVar = "DT_GCP_TEST_UNIMPERSONABLE_SERVICE_ACCOUNT"
+	serviceAccountID := os.Getenv(unimpersonableServiceAccountEnvVar)
```
Kirdock: We should put this to our envutils, right?

.github/workflows/tests.yml
Kirdock: I'm fine with the current solution, no need to overcomplicate it :)

dynatrace/api/builtin/hyperscalerauthentication/connections/gcp/service_test.go
```diff
@@ -36,3 +42,52 @@ func TestGcpConnection(t *testing.T) {
...
+	if serviceAccountID == "" {
+		t.Skipf("%s is not set; skipping GCP authentication failure test", unimpersonableServiceAccountEnvVar)
+	}
+
+	config, _ := api.ReadTfConfig(t, "testdata-auth-failure/gcp_connection.tf")
+	config = strings.ReplaceAll(config, "#service_account_id#", serviceAccountID)
```
Kirdock: Instead of using the env and replacing it, would something speak against using Terraform variables instead?
`variable "DT_GCP_TEST_UNIMPERSONABLE_SERVICE_ACCOUNT"`

=> `TF_VAR_DT_GCP_TEST_UNIMPERSONABLE_SERVICE_ACCOUNT: mySecret` (in workflow file)

dynatrace/api/builtin/hyperscalerauthentication/connections/gcp/service_test.go
d0weinberger: Oh yes, definitely!

dynatrace/api/builtin/hyperscalerauthentication/connections/gcp/service_test.go
d0weinberger: sounds also good. This means we don't need it in envutils then right?

dynatrace/api/builtin/hyperscalerauthentication/connections/gcp/service.go
d0weinberger: Yup!

dynatrace/api/builtin/hyperscalerauthentication/connections/gcp/service_test.go
Kirdock: yes

dynatrace/api/builtin/hyperscalerauthentication/connections/gcp/testdata-auth-failure/gcp_connection.tf
```diff
@@ -0,0 +1,21 @@
+variable "DT_GCP_TEST_UNIMPERSONABLE_SERVICE_ACCOUNT" {
+    type        = string
+    description = "The service account that should be used for the auth failure test. This service account must not allow impersonation by the DT GCP principal."
```
Kirdock, line 3: I guess we don't need to set it to sensitive or do we?

provider/envutils/env_vars.go
```diff
@@ -521,6 +521,16 @@ var TFAcc = StringEnvVar{
...
+// DTGCPTestUnimpersonableServiceAccount represents a GCP service account that no Dynatrace GCP
+// Principal can impersonate. Used in tests to check API behavior when triggering authorization
+// errors. The key is intentionally the TF_VAR_-prefixed name: the same value populates the
+// Terraform variable of the same name in the test config, so one env var drives both the
+// skip-guard and the .tf input.
+var DTGCPTestUnimpersonableServiceAccount = StringEnvVar{
+	Key:          "TF_VAR_DT_GCP_TEST_UNIMPERSONABLE_SERVICE_ACCOUNT",
+	DefaultValue: "",
+}
+
```
Kirdock, line 533: Do we even need and check this one? What happens for the test if the value is not provided?

dynatrace/api/builtin/hyperscalerauthentication/connections/gcp/service.go
d0weinberger: done

dynatrace/api/builtin/hyperscalerauthentication/connections/gcp/service_test.go
d0weinberger: done. I still use it as a skip guard for the test

provider/envutils/env_vars.go
d0weinberger: The test is useless without it. I guess it would fail because the value isn't provided and we require it. We also check `AccEnvsGiven` at the start of the test to check whether those vars are set. I would see it the same here

dynatrace/api/builtin/hyperscalerauthentication/connections/gcp/testdata-auth-failure/gcp_connection.tf
d0weinberger: I did, just to be safe

Kirdock (approved):

PR #1129, opened 2026-06-03
feat: Add initial release notes skill

arthurpitman:
#### **Why** this PR?
We'd like to leverage coding agents when writing release notes.

#### **What** has changed?
A skill `release-notes` is added.

#### **How** does it do it?
The skill's markdown describes how release notes should be generated.

#### How is it **tested**?
Tested with OpenCode.

NA

**Issue:** CA-19657


arthurpitman:
Not needed any more

PR #1130, opened 2026-06-04
fix: reserve 5 s buffer before context deadline in retry budget

d0weinberger:
#### Why this PR?

Users who misconfigure a hyperscaler connection (wrong IAM policies, unimpersonable service account, etc.) sit in the retry loop until the create timeout and rely entirely on the final error message to understand what went wrong. Because the retry budget previously coincided with the context deadline, a request in-flight at expiry could be cancelled mid-flight, surfacing a generic "context deadline exceeded" instead of the real constraint-violation message — leaving users with no actionable information. This affects all three hyperscaler connection types (GCP, Azure, AWS) since they all share the same retry package. The race was also observable as a flake in `TestAccGcpConnectionAuthenticationFailure`, where `ExpectError` occasionally missed the expected message.

#### What has changed?

- retry.go: new exported constant `RetryDeadlineBuffer = 5 * time.Second` and `DurationUntilDeadlineOrDefault` now subtracts it from the remaining context time when computing the retry budget.
- retry_test.go: `context.Background()` replaced with `t.Context()`; expected value in `_WithDeadline_Plenty` updated to account for the buffer; new test `_DeadlineWithinBuffer` added; pre-existing typo in a doc comment fixed.
- gcp/service_test.go: the "KNOWN FLAKE" comment block removed from `TestAccGcpConnectionAuthenticationFailure` — the root cause it described is now resolved.

#### How does it do it?

`DurationUntilDeadlineOrDefault` returns `time.Until(deadline) - RetryDeadlineBuffer` instead of the raw remaining time. The retry loop therefore exhausts its budget 5 seconds before the context deadline, guaranteeing the last attempt runs against a still-live context and its error propagates cleanly.

#### How is it tested?

- Unit tests in retry_test.go cover: no deadline (returns default), deadline with plenty of time (returns remaining - RetryDeadlineBuffer), deadline within the buffer (returns zero), and an already-expired deadline (returns zero).
- The integration test `TestAccGcpConnectionAuthenticationFailure` (which validated the flaky behaviour end-to-end) remains skipped pending GCP environment setup, but its flake-description comment has been removed as the fix is in place.

Users with an invalid configuration (e.g. wrong IAM policies or an unimpersonable service account) will now consistently see the real API error message when the create timeout is reached, instead of occasionally receiving a generic "context deadline exceeded" error that provides no actionable information. 

The effective create timeout is reduced by 5 seconds, which is the deliberate cost of this fix.

**Issue:** CA-19463

### DIFF
--- a/dynatrace/api/builtin/hyperscalerauthentication/connections/retry/retry.go
+++ b/dynatrace/api/builtin/hyperscalerauthentication/connections/retry/retry.go
@@ -26,17 +26,12 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
 )
 
-// RetryDeadlineBuffer is reserved before the context deadline so the final retry attempt can
-// complete against a still-live context. Without it, the retry budget would coincide with the
-// context deadline: a request in-flight at expiry would be cancelled and surface a generic
-// "context deadline exceeded" error instead of the real, actionable API error (and, for retries
-// driven by eventual consistency, would mask the underlying constraint violation).
+// RetryDeadlineBuffer is reserved before the context deadline so the final retry attempt
+// completes against a live context and returns the real API error, not "context deadline exceeded".
 const RetryDeadlineBuffer = 5 * time.Second
 
-// DurationUntilDeadlineOrDefault computes the retry budget: the duration until the context deadline
-// minus RetryDeadlineBuffer, or the provided defaultTimeout if no deadline is set. Reserving the
-// buffer ensures the retry loop gives up slightly before the context is cancelled, so the last
-// attempt finishes and returns the real error rather than a context-cancellation error.
+// DurationUntilDeadlineOrDefault returns the time until deadline minus RetryDeadlineBuffer,
+// or defaultTimeout if no deadline is set.
 func DurationUntilDeadlineOrDefault(ctx context.Context, defaultTimeout time.Duration) time.Duration {
 	dl, hasDeadline := ctx.Deadline()
 	if !hasDeadline {
--- a/dynatrace/api/builtin/hyperscalerauthentication/connections/retry/retry.go
+++ b/dynatrace/api/builtin/hyperscalerauthentication/connections/retry/retry.go
@@ -26,18 +26,28 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
 )
 
-// DurationUntilDeadlineOrDefault computes the duration until the context deadline,
-// or returns the provided defaultTimeout if no deadline is set.
+// RetryDeadlineBuffer is reserved before the context deadline so the final retry attempt can
+// complete against a still-live context. Without it, the retry budget would coincide with the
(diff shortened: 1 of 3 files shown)


dynatrace/api/builtin/hyperscalerauthentication/connections/retry/retry.go
```diff
@@ -26,18 +26,28 @@ import (
...
+// complete against a still-live context. Without it, the retry budget would coincide with the
+// context deadline: a request in-flight at expiry would be cancelled and surface a generic
+// "context deadline exceeded" error instead of the real, actionable API error (and, for retries
+// driven by eventual consistency, would mask the underlying constraint violation).
+const RetryDeadlineBuffer = 5 * time.Second
+
+// DurationUntilDeadlineOrDefault computes the retry budget: the duration until the context deadline
+// minus RetryDeadlineBuffer, or the provided defaultTimeout if no deadline is set. Reserving the
+// buffer ensures the retry loop gives up slightly before the context is cancelled, so the last
+// attempt finishes and returns the real error rather than a context-cancellation error.
```
Kirdock: Not sure if the documentation is overkill now, as I can read and understand the code way faster than reading the documentation, but maybe it's just me

Kirdock (dismissed):

dynatrace/api/builtin/hyperscalerauthentication/connections/retry/retry.go
arthurpitman:
```suggestion
/ DurationUntilDeadlineOrDefault returns the time until deadline minus RetryDeadlineBuffer,
// or defaultTimeout if no deadline is set.
```

dynatrace/api/builtin/hyperscalerauthentication/connections/retry/retry.go
```diff
@@ -26,18 +26,28 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
 )
 
-// DurationUntilDeadlineOrDefault computes the duration until the context deadline,
-// or returns the provided defaultTimeout if no deadline is set.
+// RetryDeadlineBuffer is reserved before the context deadline so the final retry attempt can
+// complete against a still-live context. Without it, the retry budget would coincide with the
+// context deadline: a request in-flight at expiry would be cancelled and surface a generic
+// "context deadline exceeded" error instead of the real, actionable API error (and, for retries
+// driven by eventual consistency, would mask the underlying constraint violation).
```
arthurpitman:
```suggestion
// RetryDeadlineBuffer is reserved before the context deadline so the final retry attempt
// completes against a live context and returns the real API error, not "context deadline exceeded".

```

arthurpitman (dismissed):
Looks good, I agree about the comments though so left some suggestions.

d0weinberger:
I simplified the comments based on Arthurs suggestions

arthurpitman (approved):

Kirdock (approved):

PR #1136, opened 2026-06-09
feat: maintenance windows

Kirdock:
#### Up for discussion
I went with `dynatrace_maintenance_windows` as this is the name of the schema, but we also have a deprecated resource named `dynatrace_maintenance_window` (without `s`).
Either we're fine with it or we rename the current one to v2/v3?

Outcome: Decided to use `dynatrace_maintenance_windows`. Previous deprecated resources are referring to this one now.

#### **Why** this PR?
Adds support for the new `builtin:maintenance-windows` schema, which will replace `dynatrace_maintenance`.

#### **What** has changed?
- New resource `dynatrace_maintenance_windows` added
- Deprecated `dynatrace_maintenance`

#### **How** does it do it?
By creating the resource based on the schema

#### How is it **tested**?
New example added

They can make use of the new resource/schema

**Issue:** CA-18270


templates/resources/maintenance.md.tmpl
```diff
@@ -1,13 +1,15 @@
...
 page_title: "dynatrace_maintenance Resource - terraform-provider-dynatrace"
-subcategory: "Alerting"
+subcategory: "Deprecated"
 description: |-
   The resource `dynatrace_maintenance` covers configuration for maintenance windows
 ---
 
 # dynatrace_maintenance (Resource)
 
+~> **Warning** This resource has been deprecated in favor of `dynatrace_maintenance_windows`.
```
arthurpitman: Shall we add a link also here, like in `dynatrace_maintenance_window`?

Kirdock: Done 👍

arthurpitman (approved):

PR #1137, opened 2026-06-09
refactor: replace dynatrace_gcp_principal data source with a resource

d0weinberger:
#### **Why** this PR?

- The `dynatrace_gcp_principal` data source had a side effect: reading it provisioned the Dynatrace GCP Principal if it didn't exist yet. Data sources must not create resources.
- As a consequence, a *read* required *write* permissions — an absurd contract for a data source.
- Decision: model provisioning as a resource instead, and drop the data source entirely for now (re-add later only if a genuine read-only use case appears).

#### **What** has changed?

- Removed the `dynatrace_gcp_principal` data source (`data_source.go`, its `schema.json`, and `testdata/datasource.tf`).
- Added a `dynatrace_gcp_principal` resource under `dynatrace/api/builtin/.../gcpdynatraceprincipal/`: a `CRUDService` (`service.go`), computed-only `settings.go`, `schema.json`, and `testdata/resource.tf`.
- Moved `ValidConnection` and its schema-drift guard test (`validate_test.go`) into the resource package; renamed the acceptance test to `service_test.go`.
- Repointed the GCP connection e2e config (`e2e_test.tf`) to provision the principal via the resource instead of the data source.
- Provider wiring (`provider.go`, `resource_descriptor.go`, `enums.go`) and doc templates are intentionally **not** included in this PR.

#### **How** does it do it?

- The resource takes **no input**. `Create` ignores its settings argument and calls `ensurePrincipalExists`, which lists existing principals and, if none exist, triggers provisioning via a `validateOnly` call against the GCP **connection** schema, then polls (up to a
5-minute timeout) until the Dynatrace-managed singleton principal appears.
- If a principal already exists, `Create` is a no-op that returns the existing stub. `Update` and `Delete` are no-ops, since the principal is a tenant-wide singleton that can't be deleted.
- The `principal` attribute is `Computed`-only, so Terraform rejects any attempt to set it in config — enforcing the no-input contract.
- A code comment documents the counter-intuitive permission model: write access is needed on the GCP *connection* schema, not the principal schema.

#### How is it **tested**?

- `service_test.go` — an integration-tagged acceptance test that provisions the principal and asserts the output matches a GCP service-account email (`iam.gserviceaccount.com$`). Currently `t.Skip`-ped until the resource is wired into `provider.go`.
- `validate_test.go` — an integration-tagged guard that submits `ValidConnection` to the connection's validate endpoint, failing if the API rejects the payload (which would mean principal creation can never be triggered).

- No user-facing change yet: the resource is not wired into the provider in this PR, and the data source is removed. Neither is exposed until the wiring/docs land in a follow-up.
- Once wired, users will provision the principal with a `dynatrace_gcp_principal` resource (no arguments) and reference its computed `principal` attribute, rather than reading a side-effecting data source.

**Issue:** CA-19461


dynatrace/api/builtin/hyperscalerauthentication/connections/gcpdynatraceprincipal/service.go
```diff
@@ -0,0 +1,142 @@
...
+	if len(stubs) == 0 {
+		// Provisioning is triggered by a validateOnly call against the GCP connection schema, not
+		// the principal schema. As a consequence the caller needs write access on the GCP connection
+		// schema (builtin:hyperscaler-authentication.connections.gcp) to provision the principal.
+		// Write access on the principal schema is not required at all — a caller denied write access
+		// on the principal schema can still trigger its creation here. This is a Dynatrace API
+		// characteristic; we cannot work around it here.
+		if v, ok := me.connectionService.(settings.Validator[*gcpconnection.Settings]); ok {
+			if err = v.Validate(ctx, &ValidConnection); err != nil {
+				return nil, fmt.Errorf("failed to trigger Dynatrace GCP Principal creation. Please reach out to support: %w", err)
```
Kirdock: Based on https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/1131#discussion_r3362856966
should we remove all occurrences of `Please reach out to support`?

dynatrace/api/builtin/hyperscalerauthentication/connections/gcpdynatraceprincipal/service.go
```diff
@@ -0,0 +1,142 @@
...
+// returns the existing stub; otherwise creation is triggered and ensurePrincipalExists polls until
+// the principal appears (or principalCreationTimeout elapses).
+func (me *service) Create(ctx context.Context, _ *gcpprincipalsettings.Settings) (*api.Stub, error) {
+	stub, err := me.ensurePrincipalExists(ctx)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return stub, nil
```
Kirdock:
```suggestion
	return me.ensurePrincipalExists(ctx)
```

dynatrace/api/builtin/hyperscalerauthentication/connections/gcpdynatraceprincipal/service.go
d0weinberger: Up for debate, but here I think there is genuinely no way how users could reconcile this

arthurpitman: I think users should *always*  reach out to support to help with errors, but that should not be part of the message.

d0weinberger: Okay, I removed it from the messages

Kirdock (approved):

PR #1141, opened 2026-06-11
feat: platform slo data source

Kirdock:
#### **Why** this PR?
Users had no way to reference an existing platform SLO that wasn't managed by Terraform, forcing them to hardcode IDs or duplicate some logic (e.g. DQLs).

#### **What** has changed?
Adds a new `dynatrace_platform_slo` data source (plus docs and an integration test).

#### **How** does it do it?
It reuses the `dynatrace_platform_slo` resource schema (set to computed) with `name` as the lookup key.
On read, it lists all SLOs, matches by name, fetches the full object, and marshals every property into state.

#### How is it **tested**?
Integration test (`TestAccDataSource`) backed by an example that creates an SLO and reads it back, asserting on its `id` and `custom_sli.indicator`.

Users can now look up an existing platform SLO by name and consume all of its properties.

**Issue:** CA-19712

### DIFF
--- a/datasources/slo/platform/data_source.go
+++ b/datasources/slo/platform/data_source.go
@@ -1,0 +1,95 @@
+/**
+* @license
+* Copyright 2026 Dynatrace LLC
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+ */
+
+package platformslo
+
+import (
+	"context"
+
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api"
+	sloservice "github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/slo"
+	slosettings "github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/slo/settings"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider/config"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider/logging"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/terraform/hcl"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func DataSource() *schema.Resource {
+	scm := hcl.SetComputedSchema(new(slosettings.SLO).Schema())
+	scm["name"] = &schema.Schema{
+		Type:        schema.TypeString,
+		Description: "Name of the SLO",
+		Required:    true,
+	}
+
+	return &schema.Resource{
+		ReadContext: logging.EnableDSCtx(DataSourceRead),
+		Schema:      scm,
+	}
+}
+
+func DataSourceRead(ctx context.Context, d *schema.ResourceData, m any) diag.Diagnostics {
+	name := d.Get("name").(string)
+
+	creds, err := config.Credentials(m, config.CredValDefault)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	service := sloservice.Service(creds)
+	var stubs api.Stubs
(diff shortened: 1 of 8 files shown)


datasources/slo/platform/data_source.go
```diff
@@ -0,0 +1,95 @@
...
+	slosettings "github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/slo/settings"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider/config"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider/logging"
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/terraform/hcl"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func DataSource() *schema.Resource {
+	scm := hcl.SetComputedSchema(new(slosettings.SLO).Schema())
```
d0weinberger, line 34: I would honestly prefer to just have the schema statically defined here. You see at the first glance what you can expect. In my opinion, we do not gain much from introducing `SetComputedSchema` and it increases complexity reading the code

templates/data-sources/platform_slo.md.tmpl
```diff
@@ -0,0 +1,40 @@
...
+
+# dynatrace_platform_slo (Data Source)
+
+-> **Dynatrace SaaS only**
+
+-> This data source requires the OAuth permission **View SLOs** (`slo:slos:read`)
+
+The `dynatrace_platform_slo` data source allows to get an SLO and all its data by its name.
+
+If multiple SLOs match the given criteria, the first result will be retrieved.
```
d0weinberger: That is a weird behaviour, don't you think? Why wouldn't we return a list, in this case?

datasources/slo/platform/data_source.go
Kirdock: I would say we greatly reduce the redundancy and improve maintenance with it, as this matches exactly the SLO resource

templates/data-sources/platform_slo.md.tmpl
Kirdock: We don't return a list, we filter the result and take the first one that matches.
The requirement is to return one item, not a list

d0weinberger: Which is weird. Why don't we return all matches?

datasources/slo/platform/data_source.go
Kirdock: In the resource there is also the MarshalHCL defined, which we need for setting the retrieved data

d0weinberger: Should we use this for all datasources then?

Kirdock: Yes and no. This is the first data source where we actually have the whole payload of a resource, so this use case doesn't fit the others.

templates/data-sources/platform_slo.md.tmpl
Kirdock: The defined use case is to fetch a single SLO.
Also, I can actually remove this line as the name of an SLO is unique

d0weinberger: That would make it clearer, yes. The line makes it seem as if the name wasn't unique

Kirdock: Removed the line 👍

datasources/slo/platform/data_source.go
d0weinberger: I see the benefit now, because the SLO resource is not small and having the whole schema duplicated here would not be great, especially in case of changes to the original SLO resource schema

templates/data-sources/platform_slo.md.tmpl
```diff
@@ -0,0 +1,38 @@
+---
+layout: ""
+page_title: "dynatrace_platform_slo Data Source - terraform-provider-dynatrace"
+subcategory: "Service-level Objective"
```
arthurpitman, line 4: nit: Is it "Service-Level Objective" or "Service-level Objective"?

arthurpitman: see: https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/1141/changes#diff-e66d35f28c510e93d15a7bb34ba725448519aad7c5cd66b213ac535a060b4824R19

Kirdock: It actually is `Service-Level Objectives`, but currently we are using `Service-level Objectives` for every SLO resource/data-source, so I kept it the same

arthurpitman (approved):

d0weinberger (approved):

PR #1146, opened 2026-06-12
fix: AddInsertAfterWeakIDDependency adds self dependency to resources with insert after

arthurpitman:
#### **Why** this PR?
`AddInsertAfterWeakIDDependencies` was intended to prevent exported ordered Settings 2.0 resources from keeping hardcoded IDs in `insert_after`, but the updated `ResourceDescriptor` values were not written back to the resource map. As a result, the weak self-dependency was not actually applied during export.

- `dynatrace/export/resource_descriptor.go` updates `AddInsertAfterWeakIDDependencies` to accept the resource map as an argument and write each modified `ResourceDescriptor` back into that map after appending `Dependencies.WeakID(resType)`, so the weak self-dependency is persisted.
- `dynatrace/export.go` now passes `export.AllResources` into `AddInsertAfterWeakIDDependencies` explicitly instead of relying on hidden access to the global resource registry.

#### How is it **tested**?
- Unit tests added in `dynatrace/export/resource_descriptor_test.go` for Settings 2.0 resources with `insert_after`, non-Settings 2.0 resources, resources without `insert_after`, existing weak dependencies, and mixed resource sets.

Users exporting ordered Settings 2.0 resources no longer get hardcoded IDs left in `insert_after` when the reference points to the same resource type.

**Issue:** CA-19762

### DIFF
--- a/dynatrace/export/resource_descriptor_test.go
+++ b/dynatrace/export/resource_descriptor_test.go
@@ -1,0 +1,447 @@
+//go:build unit
+
+/*
+ * @license
+ * Copyright 2026 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


dynatrace/export/resource_descriptor_test.go
Kirdock, line 1: [nit] We usually don't use snake_case for t.Run names

Kirdock (dismissed):

dynatrace/export/resource_descriptor_test.go
arthurpitman: Now fixed!

Kirdock (approved):

PR #1132, opened 2026-06-04
feat: wire and document gcp connection feature

d0weinberger:
#### Why this PR? 

- The `dynatrace_gcp_connection` and `dynatrace_gcp_principal` resources existed in the codebase but were not yet registered with the provider, not exportable, had their acceptance tests skipped, and lacked documentation. 
- This PR wires them up end-to-end so users can actually configure GCP connections via service account impersonation. 

#### What has changed? 

- Provider registration: `dynatrace_gcp_principal` and `dynatrace_gcp_connection` resources added to provider.go. 
- Export support: new `GCPConnection` and `GCPPrincipal` resource types in `export/enums.go`, descriptors wired in `resource_descriptor.go`, and `GCPConnection` registered as a dependency of the settings-permissions resource. `documentation/supported-resources.md` adapted to list both.
- Docs: new `templates/resources/gcp_connection.md.tmpl` and `templates/resources/gcp_principal.md.tmpl`
- Examples/tests: new `examples/full-usage-example.tf` (referenced by both resource doc templates) and `testdata/terraform/e2e-test.tf`; previously-skipped acceptance tests (`TestGcpConnection`, `TestAccGcpConnectionAuthenticationFailure`, `TestAccResourceGcpPrincipal`) un-skipped. 
- CI: new `TF_VAR_DT_GCP_TEST_IMPERSONABLE_SERVICE_ACCOUNT` env var added to the integration test job. 

#### **How** does it do it?

  - The resources follow the existing generic-resource and `settings20` registration patterns already used by the AWS/Azure connections.
  - The connection is added as an export dependency of the settings-permissions resource so exported configurations preserve ordering.
  - The `dynatrace_gcp_principal` resource takes no input: it triggers provisioning of the Dynatrace-managed singleton principal and exposes it via the computed `principal` attribute; `update` and `delete` are no-ops.
  - The resource doc templates embed the full-usage example via `tffile` and document the IAM-propagation retry (2-minute default create timeout, adjustable via a `timeouts` block) and the `dynatrace_gcp_principal` grant requirement.

  #### How is it **tested**?

  - Integration/acceptance tests un-skipped: `TestGcpConnection` (now using `api.TestAcc(t)` directly, since its e2e config no longer creates the GCP-side service account itself), `TestAccGcpConnectionAuthenticationFailure`, and
  `TestAccResourceGcpPrincipal`.
[...]

PR #1147, opened 2026-06-15
refactor: GCP principal test output var renaming

d0weinberger:
#### **Why** this PR?
This is a follow-up PR to #1132  to address minor issues.

#### **What** has changed?
The output variable in the GCP principal test has been renamed from `records` to `principal_output`

#### **How** does it do it?
N/A

#### How is it **tested**?
N/A

It doesn't

**Issue:** CA-19464


Kirdock (commented):
We should also update the output name in the service_test

Kirdock (approved):

PR #1151, opened 2026-06-18
refactor: IAM client uses standard request processing

arthurpitman:
#### **Why** this PR?
- The IAM client had its own response-code validation and error handling, which duplicated the standard request-processing behavior used elsewhere.
- The code was brittle when callers had to inspect specific API errors.

- Refactored `IAMClient` so methods return `api.Response` instead of raw response bytes, and removed the custom `handleResponse` plus the exact-status-code method parameters.
- The client now relies on `api.NewResponseFromHTTPResponse(...)` for standard success and error handling.
- Updated IAM callers to unmarshal from `response.Data` and use the simplified request interface.
- Tightened IAM-specific error handling where callers need to branch on API errors: `dynatrace/api/iam/policies/policy_level.go` now uses `coreapi.IsNotFoundError(err)` instead of string-matching `404`, and `dynatrace/api/iam/bindings/service.go` now tolerates `400 Bad Request` when deleting bindings that no longer exist.
- Removed now-unnecessary multi-response helpers and related call-site handling

#### How is it **tested**?
- Existing tests
- Updated IAM client mocks in `dynatrace/testing/mockiamclient.go`.
- Unit tests updated in `dynatrace/api/iam/boundaries/service_unit_test.go`.
- Unit tests updated in `dynatrace/api/iam/serviceusers/service_test.go`.

Error messages contain more detail. Previously:
```
╷
│ Error: API error: response code 400 (expected: [201])
│
│   with dynatrace_iam_user.user1_test_com,
│   on main.tf line 11, in resource "dynatrace_iam_user" "user1_test_com":
│   11: resource "dynatrace_iam_user" "user1_test_com" {
│
╵
```

Now:
```
╷
│ Error: API error: API request HTTP POST https://api.dynatace.com/iam/v1/accounts/12345678-1234-1234-1234-123456789012/users failed with status code 400: {"error":true,"message":"Following users: user1@test.com have already been invited into the account: 12345678-1234-1234-1234-123456789012","payload":null}
│
│   with dynatrace_iam_user.user1_test_com,
│   on main.tf line 11, in resource "dynatrace_iam_user" "user1_test_com":
│   11: resource "dynatrace_iam_user" "user1_test_com" {
│
╵
```
**Issue:** PS-45554, PS-45615

### DIFF
--- a/dynatrace/api/iam/bindings/service.go
+++ b/dynatrace/api/iam/bindings/service.go
@@ -111,8 +111,9 @@ func (me *BindingServiceClient) Get(ctx context.Context, id string, v *bindings.
 		v.Environment = levelID
 	}
 	v.GroupID = groupID
+
 	for idx, policyID := range v.PolicyIDs {
-		v.PolicyIDs[idx] = fmt.Sprintf("%s#-#%s#-#%s", policyID, levelType, levelID)
+		v.PolicyIDs[idx] = policies.Join(policyID, levelType, levelID)
 	}
 	return nil
 }
@@ -207,7 +208,7 @@ func (me *BindingServiceClient) List(ctx context.Context) (api.Stubs, error) {
 	for _, policy := range policyBindingsResponse.PolicyBindings {
 		for _, group := range policy.Groups {
 			if _, exists := groupIds[group]; !exists {
-				id := fmt.Sprintf("%s#-#%s#-#%s", group, "account", me.AccountID())
+				id := join(group, "account", me.AccountID())
 				stubs = append(stubs, &api.Stub{ID: id, Name: "PolicyBindings-" + id})
 				groupIds[group] = true
 			}
@@ -229,7 +230,7 @@ func (me *BindingServiceClient) List(ctx context.Context) (api.Stubs, error) {
 		for _, policy := range policyBindingsResponse.PolicyBindings {
 			for _, group := range policy.Groups {
 				if _, exists := groupIds[group]; !exists {
-					id := fmt.Sprintf("%s#-#%s#-#%s", group, "environment", environment.ID)
+					id := join(group, "environment", environment.ID)
 					stubs = append(stubs, &api.Stub{ID: id, Name: "PolicyBindings-" + id})
 					groupIds[group] = true
 				}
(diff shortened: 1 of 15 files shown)


dynatrace/api/iam/bindings/service.go
```diff
@@ -249,7 +251,12 @@ func (me *BindingServiceClient) Delete(ctx context.Context, id string) error {
 		return err
 	}
 	for _, policyID := range binding.PolicyIDs {
-		if _, err = iam.NewIAMClient(ctx, me).DELETE_MULTI_RESPONSE(ctx, fmt.Sprintf("/iam/v1/repo/%s/%s/bindings/%s/%s", levelType, levelID, policyID, groupID), rest2.RequestOptions{}, []int{204, 400}); err != nil {
+		if _, err = iam.NewIAMClient(ctx, me).DELETE(ctx, fmt.Sprintf("/iam/v1/repo/%s/%s/bindings/%s/%s", levelType, levelID, policyID, groupID), rest2.RequestOptions{}); err != nil {
+			// A 400 Bad Request is tolerated here: it indicates the binding no longer exists
+			var apiErr coreapi.APIError
+			if errors.As(err, &apiErr) && apiErr.StatusCode == http.StatusBadRequest {
+				continue
+			}
```
Kirdock: We really get a 400 instead of a 404?

dynatrace/api/iam/groups/service.go
```diff
@@ -95,22 +95,22 @@ func (me *GroupServiceClient) Name() string {
 // Permissions              groups.Permissions `json:"permissions"`
 func (me *GroupServiceClient) Create(ctx context.Context, group *groups.Group) (*api.Stub, error) {
 	var err error
```
Kirdock: [nit] this one can be removed too

dynatrace/api/iam/permissions/service.go
```diff
@@ -109,7 +109,6 @@ type GetGroupPermissionsResponse struct {
 
 func (me *PermissionServiceClient) Get(ctx context.Context, id string, v *permissions.Permission) error {
 	var err error
```
Kirdock: [nit] this one can be removed too

dynatrace/api/iam/policies/policy_level.go
```diff
@@ -320,11 +318,11 @@ func getEnvironmentIDs(ctx context.Context, auth iam.Authenticator) ([]string, e
 	var err error
```
Kirdock: [nit] this one can be removed too

dynatrace/api/iam/policies/service.go
```diff
@@ -206,16 +203,16 @@ func listForEnvironment(ctx context.Context, auth iam.Authenticator, environment
 		defer close(results)
 		client := iam.NewIAMClient(ctx, auth)
 
-		var response ListPoliciesResponse
-		var responseBytes []byte
-		if responseBytes, err = client.GET(ctx, fmt.Sprintf("/iam/v1/repo/environment/%s/policies", environmentID), rest2.RequestOptions{}, 200); err != nil {
+		var policiesResponse ListPoliciesResponse
+		response, err := client.GET(ctx, fmt.Sprintf("/iam/v1/repo/environment/%s/policies", environmentID), rest2.RequestOptions{})
```
Kirdock: It doesn't matter here, I guess that this is now a scoped variable, as the named return parameter would never be set, right?

dynatrace/api/iam/bindings/service.go
arthurpitman: Yes, we have verified that is what the API actually returns.

dynatrace/api/iam/policies/service.go
arthurpitman: Good catch, you're right, it breaks the error handling below

arthurpitman: The unintended change has been reverted and instead we can improve this area in a follow up.

Kirdock (dismissed):

dynatrace/api/iam/bindings/service.go
Kirdock: [nit] because we are now using policies.SplitID(policyID, levelType, levelID) for splitting the ID, should we also use policies.Join?

arthurpitman: Good point, I have improved the formatting.

Kirdock (approved):

PR #1157, opened 2026-06-24
refactor(iam): Each service uses a single IAM client

arthurpitman:
Build one iam.IAMClient per service at construction (NewIAMClientForCredentials) and reuse it; hold the account ID in the client rather than each service.

**Issue:**


arthurpitman:
Not needed anymore

PR #1167, opened 2026-07-01
feat: Remove legacy HTTP client

arthurpitman:
#### **Why** this PR?
The legacy HTTP client path is no longer needed, so this change removes the fallback implementation and the environment variables that only existed to control that legacy behavior.

- `dynatrace/rest/api_token_client.go`, `dynatrace/rest/cluster_v1_client.go`, `dynatrace/rest/cluster_v2_client.go`, and `dynatrace/rest/hybrid_client.go` no longer branch to `legacy_request` when `DYNATRACE_HTTP_LEGACY` is set. Their `Finish` implementations now always use the current request flow.
- `dynatrace/rest/legacy_request.go` is deleted, removing the unused legacy HTTP implementation entirely.
- `dynatrace/rest/error.go` and `dynatrace/rest/request.go` keep the shared `Envelope` and `statuscodes` helpers.
- `provider/envutils/env_vars.go` removes `DYNATRACE_HTTP_LEGACY` and `DYNATRACE_HTTP_RESPONSE`

#### How is it **tested**?
N/A

Users no longer have the option to switch back to the legacy HTTP client through `DYNATRACE_HTTP_LEGACY`.

**Issue:** PS-46388


dynatrace/rest/cluster_v2_client.go
```diff
@@ -111,20 +109,6 @@ func clusterV2Client(baseURL string, apiToken string) (*rest.Client, error) {
 }
 
 func (me *cluster_v2_request) Finish(optionalTarget ...any) error {
-	credentials := me.client.Credentials()
-	if envutils.DynatraceHTTPLegacy.Get() {
-		if !credentials.ContainsClusterURL() {
-			return NoClusterURLError
-		}
-		if !credentials.ContainsClusterToken() {
-			return NoClusterTokenError
```
Kirdock, line 120: Now we can also get rid of the `NoClusterTokenError` and `NoClusterURLError` error in `dynatrace/rest/error.go`, right?

arthurpitman: And these two functions too, `credentials.ContainsClusterURL()` and `credentials.ContainsClusterToken()`.

Kirdock (approved):

PR #1168, opened 2026-07-01
docs: `README.md` links directly to Dynatrace docs for export functionality

arthurpitman:
#### **Why** this PR?
Currently the README implies that detailed information is on the Terraform Registry, infact it just links to the Dynatrace documentation.

Links directly to the Dynatrace documentation via an existing short URL.

#### How is it **tested**?
NA

Lowers frustration by saving them a click.

**Issue:** NA

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,9 +21,12 @@ Please check the documentation within the [Terraform Registry](https://registry.
 
 Additional information can be found under [Dynatrace Configuration as Code via Terraform](https://www.dynatrace.com/support/help/manage/configuration-as-code/terraform).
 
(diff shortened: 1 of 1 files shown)


Kirdock:
Not sure about that, because it also mentions the envs needed to debug it on your own, by looking at the requests

arthurpitman:
> Not sure about that, because it also mentions the envs needed to debug it on your own, by looking at the requests

That's actually the following subsection, but I've now added an explicit link for that too.

Kirdock (approved):

PR #1172, opened 2026-07-07
refactor: Align clients

arthurpitman:
#### **Why** this PR?
Platform and IAM clients were maintaining their own client construction logic, which duplicated setup already available in the shared core client factory. The IAM OAuth path also did not consistently carry `InsecureSkipVerify` into the underlying `tls.Config`.

- `dynatrace/rest/platform_request.go` now creates platform clients through `dynatrace-configuration-as-code-core/clients.Factory()`, centralizing OAuth and platform-token authentication, retry handling, rate limiting, HTTP listeners, and the `User-Agent` setup.
- `dynatrace/api/iam/iam_client.go` switches IAM client creation to the same shared client factory and uses `rest.NewContextWithOAuthRetryClient(ctx)` when creating the account REST client so the existing TLS configuration, including `InsecureSkipVerify`, is preserved.

#### How is it **tested**?
N/A

Users configuring IAM access against environments that require `insecure_skip_verify` now get consistent TLS handling there as well. Other changes are internal refactoring with no intended user-facing behavior change.

**Issue:** PS-45605

### DIFF
--- a/dynatrace/rest/platform_request.go
+++ b/dynatrace/rest/platform_request.go
@@ -20,9 +20,6 @@ package rest
 import (
 	"context"
 	"errors"
-	"fmt"
-	"maps"
-	"net/http"
 	"net/url"
 	"regexp"
 	"strings"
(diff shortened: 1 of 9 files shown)


dynatrace/rest/platform_request.go
```diff
@@ -44,64 +41,34 @@ type platform_request request
...
+			WithOAuthCredentials(clientcredentials.Config{
+				ClientID:     credentials.OAuth.ClientID,
+				ClientSecret: credentials.OAuth.ClientSecret,
+				TokenURL:     evalTokenURL(platformURL),
+			}).
+			CreatePlatformClient(NewContextWithOAuthRetryClient(ctx))
+
 	} else {
 		return nil, NoPlatformCredentialsErr
 	}
```
Kirdock: [nit] With the early `return` inside, we don't need any `else`, just `if`.

Kirdock (dismissed):

Kirdock (approved):

PR #1175, opened 2026-07-09
refactor: Clean up provider configuration

arthurpitman:
#### **Why** this PR?
This PR cleans up the provider configuration. While not _strictly_ necessary, it helps to prepare it for subsequent refactoring.

- Unused fields are removed from `config.ProviderConfiguration`.
- The term `Platform` is used to group platform credentials, rather than `OAuth` or `Automation`
- Restructuring of rules used to build the credentials from the provider block with the aim of increasing readability and consistency:
  - When no token URL is explicitly configured, ProviderConfigureGeneric now derives the SSO token endpoint from the environment URL (Sprint→Sprint SSO, Dev→Dev SSO, otherwise Prod) instead of defaulting to production in cases like a Sprint/Dev tenant with automation_env_url set. All explicit precedence rules for token URLs, client IDs/secrets, account ID, and IAM endpoint URL remain unchanged.

#### How is it **tested**?
- Existing tests,
- Extended `provider/config/config_test.go`

For production users, very little effect.  On Sprint/Dev environments, the token URL now defaults to the corresponding non-production SSO endpoint instead of production. URL cleaning is now consistent (whitespace and trailing `/` trimmed).

**Issue:** PS-45615

### DIFF
--- a/provider/config/config.go
+++ b/provider/config/config.go
@@ -18,17 +18,17 @@
 package config
 
 import (
-	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider/envutils"
 	"context"
 	"fmt"
 	"regexp"
 	"strings"
 
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider/envutils"
+
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/rest"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
-
 
 type IAM struct {
 	ClientID     string
@@ -134,15 +134,12 @@ func Credentials(m any, CredentialValidation int) (*rest.Credentials, error) {
 
 // ProviderConfiguration contains the initialized API clients to communicate with the Dynatrace API
 type ProviderConfiguration struct {
-	EnvironmentURL    string
-	DTenvURL          string
-	DTNonConfigEnvURL string
-	DTApiV2URL        string
-	ClusterAPIV2URL   string
-	ClusterAPIToken   string
-	APIToken          string
-	IAM               IAM
-	Automation        rest.OAuthCredentials
+	EnvironmentURL  string
+	ClusterAPIV2URL string
+	ClusterAPIToken string
+	APIToken        string
+	IAM             IAM
+	Automation      rest.OAuthCredentials
 }
 
 type Getter interface {
@@ -177,10 +174,6 @@ func ProviderConfigureGeneric(ctx context.Context, d Getter) (any, diag.Diagnost
 	}
 
 	clusterURL = strings.TrimSuffix(strings.TrimSuffix(clusterURL, " "), "/")
-
-	fullURL := dtEnvURL + "/api/config/v1"
-	fullNonConfigURL := dtEnvURL + "/api/v1"
-	fullApiV2URL := dtEnvURL + "/api/v2"
 
 	automation_environment_url := getString(d, "automation_env_url")
 	automation_token_url := getString(d, "automation_token_url")
@@ -244,13 +237,10 @@ func ProviderConfigureGeneric(ctx context.Context, d Getter) (any, diag.Diagnost
 	var diags diag.Diagnostics
 
 	pc := &ProviderConfiguration{
-		EnvironmentURL:    dtEnvURL,
-		DTenvURL:          fullURL,
-		DTApiV2URL:        fullApiV2URL,
-		DTNonConfigEnvURL: fullNonConfigURL,
-		APIToken:          apiToken,
-		ClusterAPIToken:   clusterAPIToken,
-		ClusterAPIV2URL:   clusterURL,
+		EnvironmentURL:  dtEnvURL,
+		APIToken:        apiToken,
+		ClusterAPIToken: clusterAPIToken,
+		ClusterAPIV2URL: clusterURL,
 		IAM: IAM{
 			ClientID:     iam_client_id,
 			AccountID:    iam_account_id,
--- a/provider/config/config.go
+++ b/provider/config/config.go
@@ -38,6 +38,24 @@ type IAM struct {
 	EndpointURL  string
 }
 
+// ProviderConfiguration contains credentials to communicate with the Dynatrace API
+type ProviderConfiguration struct {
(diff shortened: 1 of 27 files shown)


provider/config/config.go
```diff
@@ -132,153 +194,231 @@ func Credentials(m any, CredentialValidation int) (*rest.Credentials, error) {
...
+	return getTokenURLForEnvironment(getClassicEnvironmentURL(d))
+}
+
+// getIAMClientID retrieves the OAuth client ID for IAM from the provided configuration.
+// It first checks for the "iam_client_id" key, and if not found, it checks for the "client_id" and "automation_client_id" keys.
+func getIAMClientID(d Getter) string {
+	iamClientID := getString(d, "iam_client_id")
+	if iamClientID == "" {
+		iamClientID = getString(d, "client_id")
+	}
```
Kirdock: [nit] just about consistency. In some functions we use `if val := ...; val != "" { return val}` and for others we have this approach

provider/config/config.go
```diff
@@ -38,16 +38,78 @@ type IAM struct {
...
+	ProdTokenURL   = "https://sso.dynatrace.com/sso/oauth2/token"
+	SprintTokenURL = "https://sso-sprint.dynatracelabs.com/sso/oauth2/token"
+	DevTokenURL    = "https://sso-dev.dynatracelabs.com/sso/oauth2/token"
+
+	ProdIAMEndpointURL   = "https://api.dynatrace.com"
+	SprintIAMEndpointURL = "https://api-hardening.internal.dynatracelabs.com"
+	DevIAMEndpointURL    = "https://api-dev.internal.dynatracelabs.com"
+)
+
+var regexpSaasTenant = regexp.MustCompile(`https:\/\/(.*).(live|apps).dynatrace.com`)
```
Kirdock: [nit] We call the other ones "dev" and "hard", so we should align here with "prod". Same for the `extractSaasEnvironmentID` function then

```suggestion
var regexpProdTenant = regexp.MustCompile(`https:\/\/(.*).(live|apps).dynatrace.com`)
```

provider/config/config.go
```diff
@@ -132,153 +194,231 @@ func Credentials(m any, CredentialValidation int) (*rest.Credentials, error) {
...
-		oauth_endpoint_url = rest.ProdIAMEndpointURL
-	} else if strings.Contains(dtEnvURL, ".sprint.dynatracelabs.com") || strings.Contains(dtEnvURL, ".sprint.apps.dynatracelabs.com") {
-		oauth_endpoint_url = rest.SprintIAMEndpointURL
-	} else if strings.Contains(dtEnvURL, ".dev.dynatracelabs.com") || strings.Contains(dtEnvURL, ".dev.apps.dynatracelabs.com") {
-		oauth_endpoint_url = rest.DevIAMEndpointURL
+// getPlatformTokenURL returns the SSO token URL for platform based on the provided configuration.
+// It first checks if a custom token URL is specified in the configuration using the "automation_token_url" key.
+// If not, it checks the "token_url" and "iam_token_url" keys.
+// If none are found, it derives the token URL based on the environment URL.
+func getPlatformTokenURL(d Getter) string {
```
Kirdock, line 304: Note: This is slightly different than before.
Before the value was set/overwritten by the computed token endpoint through `dt_env_url` if `automation_env_url` wasn't set.

So, it was:
-> `automation_token_url` -> if `automation_env_url` isn't set, override with extracted stage of `dt_env_url` (no fallback to prod) -> if not set `token_url` -> if not set `iam_token_url` -> if not set prod URL

provider/config/config_test.go
```diff
@@ -32,16 +32,196 @@ func (mrd mockResourceData) Get(k string) any {
...
-	configuration := result.(*config.ProviderConfiguration)
-	if configuration.DTenvURL != "https://something.live.dynatrace.com/api/config/v1" {
-		t.Fail()
-	}
+func configure(t *testing.T, d mockResourceData) *config.ProviderConfiguration {
+	t.Helper()
+	result, diags := config.ProviderConfigureGeneric(t.Context(), d)
+	assert.Empty(t, diags)
+	configuration, ok := result.(*config.ProviderConfiguration)
+	assert.True(t, ok)
```
Kirdock: Shouldn't we use require here?

provider/config/config_test.go
```diff
@@ -32,16 +32,196 @@ func (mrd mockResourceData) Get(k string) any {
...
+			{name: "Specific keys win per audience",
+				data:             mockResourceData{"iam_client_id": "iam", "client_id": "generic", "automation_client_id": "automation"},
+				expectedIAM:      "iam",
+				expectedPlatform: "automation"},
+			{name: "Generic client_id shared by both",
+				data: mockResourceData{"client_id": "generic"}, expectedIAM: "generic", expectedPlatform: "generic"},
+			{name: "Only iam_client_id falls through to platform",
+				data: mockResourceData{"iam_client_id": "iam"}, expectedIAM: "iam", expectedPlatform: "iam"},
+			{name: "Only automation_client_id falls through to IAM",
+				data: mockResourceData{"automation_client_id": "automation"}, expectedIAM: "automation", expectedPlatform: "automation"},
```
Kirdock: [nit] The formatting is a bit off here and other places. Can we align that it's either a single line or one line per field? So, line-break after `name` and before the last closing `}`?

provider/config/config.go
arthurpitman: Fixed 👍

provider/config/config.go
arthurpitman: Fixed 👍

provider/config/config_test.go
```diff
@@ -32,16 +33,343 @@ func (mrd mockResourceData) Get(k string) any {
...
+				name: "Generic client_id shared by both",
+				data: mockResourceData{
+					"client_id": "generic"},
+				expectedIAM:      "generic",
+				expectedPlatform: "generic",
+			},
+			{
+				name: "Only iam_client_id falls through to platform",
+				data: mockResourceData{
+					"iam_client_id": "iam"},
```
Kirdock: [nit] formatting

Kirdock (dismissed):

provider/config/config.go
Kirdock: Discussed: Fine with the change as the impact should not be big or even possible to achieve a different behavior

Kirdock (approved):

PR #1176, opened 2026-07-10
refactor: Introduce `rest.ClientSet`

arthurpitman:
#### **Why** this PR?
We would like services to use premade REST clients rather than recreating them for every operation.

This PR contains a large number of highly similar small changes, intended to make subsequent PRs easier to review.

Services and data sources are updated to accept `rest.ClientSet` as the new dependency shape. As an intermediate step, `rest.ClientSet` still exposes credentials so that clients can be added one by one, and callers can be migrated gradually.

#### How is it **tested**?
Existing tests

NA

**Issue:** PS-45615


datasources/credentials/vault/data_source.go
```diff
@@ -112,14 +112,14 @@ func DataSourceRead(ctx context.Context, d *schema.ResourceData, m any) diag.Dia
...
 			continue
 		}
-		if typ != "" && string(credentials.Type) != typ {
+		if typ != "" && string(clientSet.Type) != typ {
 			continue
 		}
-		d.Set("scope", string(credentials.Scope))
-		d.Set("type", string(credentials.Type))
+		d.Set("scope", string(clientSet.Scope))
+		d.Set("type", string(clientSet.Type))
```
Kirdock: I would guess that this rename was not intended

datasources/tenant/data_source.go
```diff
@@ -39,24 +39,24 @@ func DataSource() *schema.Resource {
...
+	if strings.Contains(clientSet.Credentials().URL, "/e/") {
+		idx := strings.Index(clientSet.Credentials().URL, "/e/")
+		tenant = strings.TrimSuffix(strings.TrimPrefix(clientSet.Credentials().URL[idx:], "/e/"), "/")
+	} else if after, ok := strings.CutPrefix(clientSet.Credentials().URL, "http://"); ok {
 		tenant = after
 		if idx := strings.Index(tenant, "."); idx != -1 {
 			tenant = tenant[:idx]
 		}
-	} else if after, ok := strings.CutPrefix(creds.URL, "https://"); ok {
+	} else if after, ok := strings.CutPrefix(clientSet.Credentials().URL, "https://"); ok {
```
Kirdock: Should we restructure this one so that the function is not called every time we want to access the URL?

dynatrace/api/builtin/host/monitoring/mode/service.go
```diff
@@ -35,18 +35,18 @@ import (
...
+		client:    rest.APITokenClient(clientSet.Credentials()),
 	}
 }
 
 type service struct {
-	service     settings.CRUDService[*mode.Settings]
-	credentials *rest.Credentials
-	client      rest.Client
+	service   settings.CRUDService[*mode.Settings]
+	clientSet rest.ClientSet
```
Kirdock: Do we even need the clientSet here?

dynatrace/api/builtin/managementzones/service.go
```diff
@@ -36,18 +36,18 @@ import (
...
 
-func Service(credentials *rest.Credentials) settings.CRUDService[*managementzones.Settings] {
+func Service(clientSet rest.ClientSet) settings.CRUDService[*managementzones.Settings] {
 	return &service{
-		service:     settings20.Service(credentials, SchemaID, SchemaVersion, &settings20.ServiceOptions[*managementzones.Settings]{LegacyID: settings.LegacyLongDecode}),
-		client:      rest.HybridClient(credentials),
-		credentials: credentials,
+		service:   settings20.Service(clientSet, SchemaID, SchemaVersion, &settings20.ServiceOptions[*managementzones.Settings]{LegacyID: settings.LegacyLongDecode}),
+		client:    rest.HybridClient(clientSet.Credentials()),
+		clientSet: clientSet,
```
Kirdock: [nit] should we also put the SLO service in here? Then we wouldn't need the clientSet

dynatrace/api/grail/segments/service_unit_test.go
```diff
@@ -26,13 +26,16 @@ import (
...
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider/config"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
 func TestList(t *testing.T) {
 	t.Run("Returns an error if the client creation fails", func(t *testing.T) {
-		_, err := segments.Service(&rest.Credentials{}).List(t.Context())
+		_, err := segments.Service(&config.ProviderConfiguration{
+			Platform: rest.PlatformCredentials{},
```
Kirdock: [nit] Is the addition here actually needed?

dynatrace/export/environment.go
```diff
@@ -65,15 +65,15 @@ type Environment struct {
 
 func (me *Environment) TenantID() string {
 	var tenant string
-	if strings.Contains(me.Credentials.URL, "/e/") {
-		idx := strings.Index(me.Credentials.URL, "/e/")
-		tenant = strings.TrimSuffix(strings.TrimPrefix(me.Credentials.URL[idx:], "/e/"), "/")
-	} else if after, ok := strings.CutPrefix(me.Credentials.URL, "http://"); ok {
+	if strings.Contains(me.ClientSet.Credentials().URL, "/e/") {
```
Kirdock: Can we use a variable instead of calling the credential function every time?

Kirdock (commented):
Because this is just about the services and not the clients, we have to touch a lot of the same files again... so another huge PR 😅

datasources/credentials/vault/data_source.go
arthurpitman: Fixed

datasources/tenant/data_source.go
arthurpitman: Why not - it looks nicer :)

That being said, this also identifies that this data source will be a little problematic, as it just uses the URL and no client. This would mean that in the future, the client set would still need to provide the URL.

dynatrace/export/environment.go
arthurpitman: Done

dynatrace/api/builtin/managementzones/service.go
arthurpitman: Fixed

dynatrace/api/builtin/host/monitoring/mode/service.go
arthurpitman: Fixed

dynatrace/api/grail/segments/service_unit_test.go
arthurpitman: We need to pass a `rest.ClientSet`, so passing a ref to an empty `config.ProviderConfiguration` is enough.

arthurpitman:
> Because this is just about the services and not the clients, we have to touch a lot of the same files again... so another huge PR 😅

Actually the follow up PRs are much smaller and focused. It is about keeping the brain load as low as possible :)

Kirdock (approved):

PR #1178, opened 2026-07-13
refactor: Add shared IAM client to clientset

arthurpitman:
#### **Why** this PR?
As IAM always requires OAuth, a single IAMClient instance reduces the number of calls made to the SSO. As this is a bottleneck, this PR improves the stability and performance of IAM resources and data sources.

- The IAM client moves to the `rest` package and gains the `AccountID()` method that provides the fixed account ID it was created with.
- Updates provider and request plumbing in `provider/config/config.go`, `dynatrace/rest/client_set.go`, and `dynatrace/rest/request.go` so IAM client initialization happens once during provider configuration and `DYNATRACE_HTTP_INSECURE` is applied by setting `InsecureSkipVerify` on the default transport up front.
- IAM services are refactored to use a shared IAM client via `clientSet.IAMClient()` instead of building clients or reading `Credentials().IAM.AccountID` directly.
- `dynatrace/api/iam/policies/policy_level.go` updates policy and environment lookup helpers to work with `rest.IAMClient`.
- Follow-on refactoring details by fixing `ServiceWithGloabals` to `ServiceWithGlobals`, handling constructor errors in IAM data sources, and removing unused IAM-related code such as `policies.FetchPolicyLevel(...)` and `ListUsersResponse.Count`.

#### How is it **tested**?
Existing tests are updated.

No effect rather than improved performance of IAM resources and datasources

**Issue:** PS-45615

### DIFF
--- a/datasources/iam/environments/service.go
+++ b/datasources/iam/environments/service.go
@@ -28,12 +28,12 @@ import (
 const baseURL = "/env/v2/accounts"
 
 type environmentsService struct {
-	credentials *rest.Credentials
+	client rest.IAMClient
 }
 
-func newEnvironmentService(clientSet rest.ClientSet) *environmentsService {
+func newEnvironmentService(ctx context.Context, clientSet rest.ClientSet) *environmentsService {
 	return &environmentsService{
-		credentials: clientSet.Credentials(),
+		client: rest.NewIAMClient(ctx, clientSet.Credentials()),
 	}
 }
 
@@ -49,13 +49,12 @@ type Response struct {
 }
 
 func (ec *environmentsService) Get(ctx context.Context) ([]Environment, error) {
-	u, err := url.JoinPath(baseURL, ec.credentials.IAM.AccountID, "environments")
+	u, err := url.JoinPath(baseURL, ec.client.AccountID(), "environments")
 	if err != nil {
 		return nil, err
 	}
 
-	client := rest.NewIAMClient(ctx, ec.credentials)
-	response, err := client.GET(ctx, u, rest2.RequestOptions{})
+	response, err := ec.client.GET(ctx, u, rest2.RequestOptions{})
 	if err != nil {
 		return nil, err
 	}
--- a/datasources/iam/environments/service.go
+++ b/datasources/iam/environments/service.go
@@ -21,7 +21,6 @@ import (
 	"encoding/json"
 	"net/url"
 
-	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/iam"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/rest"
 	rest2 "github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
 )
@@ -55,7 +54,7 @@ func (ec *environmentsService) Get(ctx context.Context) ([]Environment, error) {
 		return nil, err
 	}
 
-	client := iam.NewIAMClient(ctx, ec.credentials)
+	client := rest.NewIAMClient(ctx, ec.credentials)
 	response, err := client.GET(ctx, u, rest2.RequestOptions{})
 	if err != nil {
 		return nil, err
--- a/datasources/iam/environments/service.go
+++ b/datasources/iam/environments/service.go
@@ -31,10 +31,15 @@ type environmentsService struct {
 	client rest.IAMClient
 }
 
-func newEnvironmentService(ctx context.Context, clientSet rest.ClientSet) *environmentsService {
-	return &environmentsService{
-		client: rest.NewIAMClient(ctx, clientSet.Credentials()),
+func newEnvironmentService(clientSet rest.ClientSet) (*environmentsService, error) {
+	iamClient, err := clientSet.IAMClient()
+	if err != nil {
+		return nil, err
 	}
+
+	return &environmentsService{
+		client: iamClient,
+	}, nil
 }
 
 type Environment struct {
--- a/dynatrace/api/iam/policies/service.go
+++ b/dynatrace/api/iam/policies/service.go
@@ -230,7 +230,7 @@ func listForAccount(ctx context.Context, credentials *rest.Credentials) (results
 		defer close(results)
 
 		var response api2.Response
-		if response, err = client.GET(ctx, fmt.Sprintf("/iam/v1/repo/account/%s/policies", credentials.IAM.AccountID), rest2.RequestOptions{}); err != nil {
+		if response, err = client.GET(ctx, fmt.Sprintf("/iam/v1/repo/account/%s/policies", client.AccountID()), rest2.RequestOptions{}); err != nil {
 			return
 		}
 
@@ -240,7 +240,7 @@ func listForAccount(ctx context.Context, credentials *rest.Credentials) (results
 		}
 
 		for _, policy := range policiesResponse.Policies {
-			results <- &api.Stub{ID: Join(policy.UUID, "account", credentials.IAM.AccountID), Name: policy.Name}
+			results <- &api.Stub{ID: Join(policy.UUID, "account", client.AccountID()), Name: policy.Name}
 		}
 	}()
 	return results, nil
--- a/dynatrace/api/iam/policies/service.go
+++ b/dynatrace/api/iam/policies/service.go
@@ -41,7 +41,7 @@ func Service(clientSet rest.ClientSet) (settings.CRUDService[*policies.Policy], 
 	return &PolicyServiceClient{credentials: clientSet.Credentials()}, nil
 }
 
-func ServiceWithGloabals(clientSet rest.ClientSet) *PolicyServiceClient {
+func ServiceWithGlobals(clientSet rest.ClientSet) *PolicyServiceClient {
 	return &PolicyServiceClient{credentials: clientSet.Credentials()}
 }
 
--- a/dynatrace/api/iam/policies/service.go
+++ b/dynatrace/api/iam/policies/service.go
@@ -34,15 +34,19 @@ import (
 )
 
 type PolicyServiceClient struct {
-	credentials *rest.Credentials
+	client rest.IAMClient
 }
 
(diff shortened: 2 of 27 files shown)


dynatrace/api/iam/serviceusers/service_test.go
```diff
@@ -36,12 +36,8 @@ import (
 const testAccountID = "test-account-id"
 
 func createTestServiceUserServiceClient(client *testing2.MockIAMClient) *serviceUserServiceClient {
-	return &serviceUserServiceClient{
-		iamClientGetter: &testing2.MockIAMClientGetter{
-			Client: client,
-		},
-		accountID: testAccountID,
-	}
+	client.AccountIDValue = testAccountID
```
Kirdock: It's a bit weird that the account ID is set/overwritten here and not during initialization, but I get the point why

dynatrace/api/iam/policies/service.go
```diff
@@ -35,15 +34,15 @@ import (
...
 
 func Service(clientSet rest.ClientSet) settings.CRUDService[*policies.Policy] {
-	return &PolicyServiceClient{credentials: clientSet.Credentials()}
+	return &PolicyServiceClient{client: clientSet.IAMClient()}
 }
 
 func ServiceWithGloabals(clientSet rest.ClientSet) *PolicyServiceClient {
-	return &PolicyServiceClient{credentials: clientSet.Credentials()}
+	return &PolicyServiceClient{client: clientSet.IAMClient()}
 }
```
Kirdock: So, the whole point of the second function is the return type? 😅

dynatrace/rest/logging/logger_test.go
```diff
@@ -119,15 +120,14 @@ func TestProviderLogging(t *testing.T) {
...
-				ClientSecret: "secret",
-				TokenURL:     httpServer.URL + "/sso",
-				EndpointURL:  httpServer.URL,
-			},
+		credentials := &rest.Credentials{}
+		credentials.IAM.ClientID = "id"
+		credentials.IAM.AccountID = "account-id"
+		credentials.IAM.ClientSecret = "secret"
+		credentials.IAM.TokenURL = httpServer.URL + "/sso"
+		credentials.IAM.EndpointURL = httpServer.URL
```
Kirdock: [nit] Why not in the struct creation?

datasources/iam/environments/service.go
```diff
@@ -21,23 +21,26 @@ import (
...
+func newEnvironmentService(ctx context.Context, clientSet rest.ClientSet) *environmentsService {
 	return &environmentsService{
-		credentials: clientSet.Credentials(),
+		client: clientSet.IAMClient(),
 	}
 }
 
+func (ec *environmentsService) AccountID() string {
+	return ec.client.AccountID()
+}
```
Kirdock: [nit] Not sure if I like it or not, but without it, it would mean calling `clientSet.IAMClient().AccountID()`, which is not too bad and wouldn't require the function here 🤔

datasources/iam/environments/service.go
```diff
@@ -21,23 +21,26 @@ import (
...
 
 const baseURL = "/env/v2/accounts"
 
 type environmentsService struct {
-	credentials *rest.Credentials
+	client rest.IAMClient
 }
 
-func newEnvironmentService(clientSet rest.ClientSet) *environmentsService {
+func newEnvironmentService(ctx context.Context, clientSet rest.ClientSet) *environmentsService {
```
Kirdock: We actually don't need the context here, right?

dynatrace/rest/iam_client.go
```diff
@@ -1,6 +1,6 @@
 /**
 * @license
-* Copyright 2025 Dynatrace LLC
+* Copyright 2026 Dynatrace LLC
```
Kirdock, line 3: [nit] I still don't quite get why we always change the license header year, as this just produces noise 🤔

dynatrace/rest/iam_client.go
```diff
@@ -43,14 +42,18 @@ type IAMClient interface {
...
+	client    *rest2.Client
+	accountID string
 }
 
-func NewIAMClient(ctx context.Context, credentials *rest.Credentials) IAMClient {
-	rest.PreRequest()
+func NewIAMClient(ctx context.Context, credentials *Credentials) IAMClient {
+	PreRequest()
 
 	client, _ := clients.Factory().
```
Kirdock: Will there be any follow-up to properly check the error? e.g., the "IAMClient" of the clientSet also returns the error returned here and/or by `validateCredentials(pc, CredValIAM)`
Also, how does the current error handling look like now, if some fields are not provided (e.g., accountId or clientId)?

dynatrace/rest/iam_client.go
```diff
@@ -43,14 +42,18 @@ type IAMClient interface {
...
 type iamClient struct {
-	client *rest2.Client
+	client    *rest2.Client
+	accountID string
 }
 
-func NewIAMClient(ctx context.Context, credentials *rest.Credentials) IAMClient {
-	rest.PreRequest()
+func NewIAMClient(ctx context.Context, credentials *Credentials) IAMClient {
+	PreRequest()
```
Kirdock: Thinking about this one: Does it make sense to have it in the clients, or should we just call it once in the provider before the clients are initialized? This would also mean that we don't need the mutex inside the `PreRequest`.

dynatrace/rest/iam_client.go
```diff
@@ -43,14 +42,18 @@ type IAMClient interface {
 	PUT(ctx context.Context, url string, payload any, options rest2.RequestOptions) (api.Response, error)
 	GET(ctx context.Context, url string, options rest2.RequestOptions) (api.Response, error)
 	DELETE(ctx context.Context, url string, options rest2.RequestOptions) (api.Response, error)
+	// AccountID returns the account ID associated with the client. It is a constant fixed when the
+	// client is created.
```
Kirdock, line 46: I'm just thinking about this comment like "a pen is used for writing" 😅 . We can keep the comment, of course.

provider/config/config.go
```diff
@@ -119,6 +125,14 @@ func ProviderConfigureGeneric(ctx context.Context, d Getter) (any, diag.Diagnost
...
 		},
 	}
+
+	// As the context passed to this function does not live long enough, use context.Background() instead.
+	// The context is used to retrieve and refresh the bearer token but not for the actual API calls.
+	clientCtx := context.Background()
+
+	if validateCredentials(pc, CredValIAM) == nil {
+		pc.iamClient = rest.NewIAMClient(clientCtx, pc.Credentials())
+	}
```
Kirdock: If this is only set if the credentials are valid, this would mean a panic because of a nil value for all the resources, right?

dynatrace/api/iam/policies/service.go
arthurpitman: Yes, I believe so, to provide access to `ListWithGlobals(...)`, but I'll at least fix the typo in the `ServiceWithGloabals(...)` function name

dynatrace/api/iam/serviceusers/service_test.go
arthurpitman: I have removed `createTestServiceUserServiceClient(...)` because it doesn't really do much anymore after the refactoring

dynatrace/rest/logging/logger_test.go
arthurpitman: I have introduced `rest.IAMCredentials`, similar to the existing `rest.PlatformCredentials`. This makes this more straightforward.

datasources/iam/environments/service.go
arthurpitman: Fixed 👍

datasources/iam/environments/service.go
arthurpitman: 👍

dynatrace/rest/iam_client.go
arthurpitman: This was a bit of a mashup. Now, after the refactoring we have done in the meantime, this error here is actually returned.

dynatrace/rest/iam_client.go
arthurpitman: This is true, I'll add it 👍

provider/config/config.go
arthurpitman: 👍

arthurpitman: `NewIAMClient(...)` now also returns an error, which is recorded and returned by the client set's `IAMClient(...)` method, allowing services to check this, and with the most recent refactoring, return this and error out early, avoiding any panic because of a nil value.

dynatrace/api/iam/policies/service.go
```diff
@@ -35,15 +34,23 @@ import (
...
+	client rest.IAMClient
 }
 
 func Service(clientSet rest.ClientSet) (settings.CRUDService[*policies.Policy], error) {
-	return &PolicyServiceClient{credentials: clientSet.Credentials()}, nil
+	iamClient, err := clientSet.IAMClient()
+	if err != nil {
+		return nil, err
+	}
+	return &PolicyServiceClient{client: iamClient}, nil
```
Kirdock: [nit]
```suggestion
	return ServiceWithGlobals(clientSet)
```
or we simply return `*PolicyServiceClient` instead of the interface and remove the `ServiceWithGlobals` function

dynatrace/rest/client_set.go
```diff
@@ -19,9 +19,10 @@ package rest
 
 // ClientSet is the credential/client provider handed to services, resources and datasources.
 //
-// For now it only exposes the raw Credentials so the existing client factories can be reused
-// unchanged. Concrete clients (e.g. an IAM client, an API-token client) will be added over time and
-// services migrated onto them, at which point Credentials can eventually be retired.
+// It exposes the raw Credentials (so the existing client factories can be reused unchanged) plus
+// concrete clients that are built once and reused. More clients will be  added over time and services
```
Kirdock: [nit]
```suggestion
// concrete clients that are built once and reused. More clients will be added over time and services
```

Kirdock (dismissed):
Nice!

Kirdock (approved):

PR #1179, opened 2026-07-13
refactor: Add shared platform client to clientset

arthurpitman:
#### **Why** this PR?
Platform-backed resources and data sources were repeatedly creating platform clients on demand, which increased load on the SSO as each operation requested a new bearer token.

- Adds a shared `PlatformClient()` to `rest.ClientSet` so platform client initialization happens once and can be reused across the provider.
- Updates platform-backed services and data sources to request the shared client from `clientSet` instead of calling `rest.CreatePlatformClient(...)` directly.
- Simplifies test wiring by updating `dynatrace/testing/mockclientset.go` and replacing callback-based injection with direct client injection.

#### How is it **tested**?
- Existing tests cover most of the changes.

NA

**Issue:** PS-45605

### DIFF
--- a/dynatrace/api/documents/directshares/service_unit_test.go
+++ b/dynatrace/api/documents/directshares/service_unit_test.go
@@ -28,10 +28,10 @@ import (
 	coreapi "github.com/dynatrace/dynatrace-configuration-as-code-core/api"
 
 	serviceSettings "github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/documents/directshares/settings"
-	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/rest"
-	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider/config"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
+	testing2 "github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/testing"
 )
 
 type mockDirectSharesClient struct {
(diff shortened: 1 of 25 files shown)


dynatrace/api/documents/directshares/service_unit_test.go
```diff
@@ -66,41 +67,28 @@ func (m *mockDirectSharesClient) RemoveRecipients(ctx context.Context, id string
...
-		svc := ServiceWithClientGetter(failingClientGetter(clientErr), &config.ProviderConfiguration{})
-		v := &serviceSettings.DirectShare{}
-		err := svc.Get(t.Context(), "id-1", v)
-		assert.ErrorIs(t, err, clientErr)
-	})
+// TestServiceCreationFailsIfMissingClient tests that the service creation fails if the platform client is missing.
+func TestServiceCreationFailsIfClientUnavailable(t *testing.T) {
+	service, err := Service(&testing2.MockClientSet{PlatformClientErr: rest.NoPlatformCredentialsErr})
+	require.Nil(t, service)
+	require.Error(t, err)
```
Kirdock: [nit] technically we could use `require.ErrorIs(t, rest.NoPlatformCredentialsErr)` or instead of `rest.NoPlatformCredentialsErr` use `assert.AnError` (also for other similar tests)

dynatrace/api/documents/document/service.go
```diff
@@ -101,18 +94,11 @@ func (me *service) SchemaID() string {
 }
 
 func (me *service) List(ctx context.Context) (api.Stubs, error) {
-	if me == nil {
+	if me == nil || me.client == nil {
```
Kirdock: [note] just about consistency. This is the only place where we validate that the client is not nil

dynatrace/api/v2/settings/objects/permissions/service.go
```diff
@@ -33,39 +33,22 @@ import (
...
-	}
-	restClient, err := rest.CreatePlatformClient(ctx, me.ClientSet.Credentials().Platform.EnvironmentURL, me.ClientSet.Credentials())
-	if err != nil {
-		return nil, err
-	}
-
-	me.SettingsClient = NewPlatformSettingsClient(restClient)
-	return me.SettingsClient, nil
+type ServiceImpl struct {
+	ClientSet      rest.ClientSet
```
Kirdock: Do we need/even use the clientSet in here?

provider/config/config.go
```diff
@@ -41,6 +42,9 @@ type ProviderConfiguration struct {
 
 	iamClient    rest.IAMClient
 	iamClientErr error
+
+	platformClient    *rest2.Client
```
Kirdock, line 46: [nit/note] just an observation. While we use an interface for the iamClient, we use a struct for the platformClient. Even though this shouldn't be an issue, I would say.

Kirdock (dismissed):

provider/config/config.go
arthurpitman: We need the struct as the clients in the core library that we use require it. If we wanted to change this, we'd have to update the clients there first.

Kirdock (approved):

PR #1182, opened 2026-07-16
refactor: Service function should also be able to return an error

arthurpitman:
#### **Why** this PR?
Service construction can fail in a number of places, but many `Service(...)` helpers could not return that failure to their callers. This refactor makes initialization errors explicit so they can be handled immediately instead of surfacing later through follow-up CRUD operations.

- Service constructors across the provider now return `(service, error)` instead of only a service instance, so initialization failures can propagate through the API boundary.
- Call sites across roughly 300 files in `datasources/*`, `resources/*`, `dynatrace/api/app/*`, `dynatrace/api/automation/*`, and `dynatrace/api/builtin/*` now create services with `service, err := Service(...)` and return early when construction fails.
- Terraform-facing entry points such as `datasources/alerting/data_source.go`, `datasources/iam/policies/data_source.go`, and `datasources/v2alerting/resource.go` convert those constructor failures into diagnostics with `diag.FromErr(err)`.
- Composed services such as `dynatrace/api/builtin/monitoring/slo/service.go`, `dynatrace/api/builtin/managementzones/service.go`, and `dynatrace/api/builtin/hyperscalerauthentication/connections/azure/authentication/service.go` are updated so dependent service creation can fail fast instead of continuing with a partially initialized dependency.
 
#### How is it **tested**?
N/A

N/A

**Issue:** PS-45615


dynatrace/api/automation/business_calendars/service.go
```diff
@@ -30,8 +30,8 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/clients/automation"
 )
 
-func Service(clientSet tfrest.ClientSet) settings.CRUDService[*business_calendars.Settings] {
-	return &service{clientSet}
+func Service(clientSet tfrest.ClientSet) (settings.CRUDService[*business_calendars.Settings], error) {
+	return &service{clientSet}, nil
```
Kirdock, line 34: [nit] (probably follow-up) with this, we could move the "client" creation to the Service.
Also for all other files like
- (all?) IAM services
- everything in cluster
- basically everything with a specific core client
- api V1, V2

dynatrace/api/v1/config/alerting/service.go
```diff
@@ -26,10 +26,10 @@ import (
 const SchemaID = "v1:config:alerting"
 const BasePath = "/api/config/v1/alertingProfiles"
 
-func Service(clientSet rest.ClientSet) settings.CRUDService[*alerting.Profile] {
+func Service(clientSet rest.ClientSet) (settings.CRUDService[*alerting.Profile], error) {
 	return settings.NewAPITokenService(
```
Kirdock, line 30: Shouldn't this one return an error? (and also `APITokenStaticService`)
The service and/or the client

When we touch the service later on, we have to touch this file (and all other V1 and V2 files again)

But I guess this is a follow-up

dynatrace/api/v1/config/anomalies/processgroups/service.go
```diff
@@ -101,7 +101,10 @@ func (me *service) Get(ctx context.Context, id string, v *processgroups.AnomalyD
 }
 
 func (me *service) List(ctx context.Context) (api.Stubs, error) {
-	srv := entities.Service("PROCESS_GROUP", "", "", "", "", me.clientSet)
+	srv, err := entities.Service("PROCESS_GROUP", "", "", "", "", me.clientSet)
```
Kirdock, line 104: [note]: this one is actually an interesting use. If we put it into the service, we could technically get an error during the client creation that may not have happened during terraform apply/plan/destroy.
Also happens in `dynatrace/api/iam/permissions/service.go`

dynatrace/api/iam/serviceusers/service.go
```diff
@@ -62,8 +62,8 @@ func NewService(clientSet rest.ClientSet) ServiceUserService {
 	}
 }
 
-func Service(clientSet rest.ClientSet) settings.CRUDService[*serviceusers.ServiceUser] {
-	return NewService(clientSet)
+func Service(clientSet rest.ClientSet) (settings.CRUDService[*serviceusers.ServiceUser], error) {
```
Kirdock, line 65: [nit] Shouldn't the NewService return an error here?

dynatrace/export/resource_descriptor.go
```diff
@@ -563,7 +575,11 @@ func ContainsInsertAfterAttribute(protoType settings.Settings, schemaID string)
 // `Dependencies.WeakID` takes care of that.
 func AddInsertAfterWeakIDDependencies(resources map[ResourceType]ResourceDescriptor) {
 	for resType, descriptor := range resources {
-		schemaID := descriptor.Service(&config.ProviderConfiguration{}).SchemaID()
+		service, err := descriptor.Service(&config.ProviderConfiguration{})
+		if err != nil {
+			continue
+		}
+		schemaID := service.SchemaID()
```
Kirdock: This is a small behavior change but I guess/hope this isn't really an issue (we only add the dependency if the credentials are given)

Kirdock (dismissed):

dynatrace/api/iam/serviceusers/service.go
arthurpitman: Yes, you are correct, but I'll fix this ~~in the immediately following IAM PR~~ now

dynatrace/api/automation/business_calendars/service.go
arthurpitman: yes, but this is in every follow up as things are added to the clientset

dynatrace/api/v1/config/alerting/service.go
arthurpitman: No, you are right, this belongs here - I will fix this

dynatrace/export/resource_descriptor.go
arthurpitman: I've done a bit more refactoring because I'd rather not have the change

Kirdock (dismissed):

Kirdock (approved):

PR #1183, opened 2026-07-18
fix: avoid panic parsing array-shaped REST error bodies

knQzx:
#### **Why** this PR?

any non-2xx classic-API response whose body is a JSON array without an `error` field (e.g. `[{"code":400}]`) panics the provider with a nil-pointer dereference

#### **What** has changed?

`Envelope` and `unmarshalError` now guard `envs[0].Error != nil` on the array branch, like the existing single-object branch already does

#### **How** does it do it?

adds `&& envs[0].Error != nil` to the array-branch condition in `dynatrace/rest/error.go` and `dynatrace/rest/request.go`, so a shape-drifted array body falls through instead of dereferencing nil

#### How is it **tested**?

added a table test in `error_test.go` for array/object bodies with and without the `error` field; it panics on the old code and passes on the new

a malformed or array-shaped error response no longer crashes the provider mid-operation

**Issue:** N/A

### DIFF
--- a/dynatrace/rest/error.go
+++ b/dynatrace/rest/error.go
@@ -38,7 +38,7 @@ func Envelope(data []byte, url string, method string) error {
 	}
 
 	var envs []errorEnvelope
-	if err := json.Unmarshal(data, &envs); err == nil && len(envs) > 0 {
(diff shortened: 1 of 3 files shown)


Kirdock (approved):
Hi, thanks for your contribution!
LGTM

PR #1191, opened 2026-07-22
feat: update `dynatrace_activegate_updates` to v1.0.4

Kirdock:
BREAKING CHANGE

#### **Why** this PR?
The resource `dynatrace_activegate_updates` always shows a non-empty plan. Turns out, the schema was completely changed. The `auto_update` field no longer exists.

#### **What** has changed?
The `dynatrace_activegate_updates` resource is up to date with the schema.

#### **How** does it do it?
The field `auto_update` has been removed.
New required fields `target_version` and `update_mode`.
New conditional field `update_windows`.

#### How is it **tested**?
Example updated. All fields are being used.
New testcase tests added, as the field `update_windows` is of type set.

They can use the resource again

**Issue:** PS-48255

### DIFF
--- a/dynatrace/api/builtin/deployment/activegate/updates/settings/settings.go
+++ b/dynatrace/api/builtin/deployment/activegate/updates/settings/settings.go
@@ -23,8 +23,10 @@ import (
 )
 
 type Settings struct {
-	AutoUpdate bool    `json:"autoUpdate"`      // Automatic updates at earliest convenience
-	Scope      *string `json:"-" scope:"scope"` // The scope of this setting (ENVIRONMENT_ACTIVE_GATE). Omit this property if you want to cover the whole environment.
+	Scope         *string       `json:"-" scope:"scope"`         // The scope of this setting (ENVIRONMENT_ACTIVE_GATE). Omit this property if you want to cover the whole environment.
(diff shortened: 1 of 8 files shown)


dynatrace/api/builtin/deployment/activegate/updates/settings/settings.go
```diff
@@ -33,31 +35,53 @@ func (me *Settings) Name() string {
...
-		"scope":       me.Scope,
+		"scope":          me.Scope,
+		"target_version": me.TargetVersion,
+		"update_mode":    me.UpdateMode,
+		"update_windows": me.UpdateWindows,
 	})
 }
 
+func (me *Settings) HandlePreconditions() error {
+	// ---- UpdateWindows UpdateWindows -> {"expectedValue":"AUTOMATIC_DURING_UW","property":"updateMode","type":"EQUALS"}
```
d0weinberger, line 76: I wonder why this one could not get generated. Did you check why it failed?

arthurpitman (approved):

dynatrace/api/builtin/deployment/activegate/updates/settings/settings.go
Kirdock: We don't check arrays as far as I know.

PR #1196, opened 2026-07-24
test: add filter_segment to SLO test file

d0weinberger:
#### **Why** this PR?
We recently had a bug in `dynatrace_platform_slo` which we would have caught earlier, had we tested the use of the filter segments field for SLOs. See: #1195 

An E2E test example has been extended to make use of filter segments.

#### How is it **tested**?
This is the test.

It doesn't

**Issue:** NOISSUE

### DIFF
--- a/dynatrace/api/slo/testdata/terraform/example-b.tf
+++ b/dynatrace/api/slo/testdata/terraform/example-b.tf
@@ -19,5 +19,17 @@ resource "dynatrace_platform_slo" "#name#" {
       | fieldsAdd sli=(((total[]-failures[])/total[])*(100))
       | fieldsRemove total, failures, tags
     EOT
+
+    filter_segments {
+      filter_segment {
+        id = "segment_1"
+        variables {
+          filter_segment_variable {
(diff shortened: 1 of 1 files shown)


dynatrace/api/slo/testdata/terraform/example-b.tf
```diff
@@ -19,5 +19,17 @@ resource "dynatrace_platform_slo" "#name#" {
       | fieldsAdd sli=(((total[]-failures[])/total[])*(100))
       | fieldsRemove total, failures, tags
     EOT
+
+    filter_segments {
+      filter_segment {
+        id = "segment_1"
```
Kirdock, line 25: So, the SLO endpoint doesn't validate the segment ID as it seems 😅

Kirdock (approved):

dynatrace/api/slo/testdata/terraform/example-b.tf
d0weinberger: Exactly

PR #1203, opened 2026-08-03
feat(document) - support labels

tomazpu:
#### **Why** this PR?

The core lib reworked its documents client and added three new document fields. This PR adopts the new client and exposes those fields on both the resource and the data source.

#### **What** has changed?

- Pinned `dynatrace-configuration-as-code-core` to latest `main` (`22208c619fec`)
- Added `description`, `labels`, `is_reshareable` to the `dynatrace_document` resource
- Added the same three fields to the `dynatrace_documents` data source
- Removed a two-step create workaround the lib now handles internally
- Regenerated docs for both pages; fixed a stale `type` description missing `launchpad`

#### **How** does it do it?

`Create`/`Update` take a single `Metadata` struct instead of individual parameters, so the service builds and passes that. For the data source, `service.List` now carries the three fields in `api.Stub.Extra` — no extra API calls, since the lib's `ListResponse` entries already embed the full `Metadata`. Nil `Description`/`IsReshareable` pointers are dereferenced to zero values so the keys are always present.

#### How is it **tested**?

Extended the existing `TestAccDocuments` e2e fixtures — `example-a.tf` creates with all three fields set, `example-b.tf` updates them including removing a label. `go build ./...` and `go vet` clean under both `unit` and `integration` tags.

`description`, `labels`, and `is_reshareable` are now settable on `dynatrace_document` and readable from `dynatrace_documents`. All three are optional, so existing configs are unaffected.

**Issue:** PS-47481

### DIFF
--- a/datasources/documents/document/data_source.go
+++ b/datasources/documents/document/data_source.go
@@ -37,7 +37,7 @@ func DataSource() *schema.Resource {
 			"type": {
 				Type:        schema.TypeString,
 				Optional:    true,
-				Description: "The type of documents to query for. Leave empty if you want to query for all kinds of documents. Possible values are `dashboard` or `notebook`",
+				Description: "The type of documents to query for. Leave empty if you want to query for all kinds of documents. Possible values are `dashboard`, `notebook` or `launchpad`",
 			},
 			"values": {
 				Type:     schema.TypeList,
@@ -63,6 +63,22 @@ func DataSource() *schema.Resource {
 							Type:        schema.TypeString,
 							Computed:    true,
 							Description: "The owner of the document. This could be a user or a group that has ownership rights over the document.",
+						},
+						"description": {
+							Type:        schema.TypeString,
+							Computed:    true,
+							Description: "A short description of the document.",
+						},
+						"labels": {
+							Type:        schema.TypeSet,
+							Computed:    true,
+							Elem:        &schema.Schema{Type: schema.TypeString},
+							Description: "Labels attached to the document.",
+						},
+						"is_reshareable": {
+							Type:        schema.TypeBool,
+							Computed:    true,
+							Description: "Specifies whether recipients of a direct share can share the document further.",
 						},
 					},
 				},
@@ -103,10 +119,13 @@ func DataSourceRead(ctx context.Context, d *schema.ResourceData, m any) diag.Dia
 				}
 			}
 			m := map[string]any{
-				"id":    stub.ID,
-				"name":  stub.Name,
-				"type":  stub.Extra["type"],
-				"owner": stub.Extra["owner"],
+				"id":             stub.ID,
+				"name":           stub.Name,
+				"type":           stub.Extra["type"],
+				"owner":          stub.Extra["owner"],
+				"description":    stub.Extra["description"],
+				"labels":         stub.Extra["labels"],
+				"is_reshareable": stub.Extra["isReshareable"],
 			}
 
 			values = append(values, m)
--- a/datasources/documents/document/data_source.go
+++ b/datasources/documents/document/data_source.go
@@ -22,7 +22,8 @@ import (
 	"fmt"
 
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api"
-	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/export"
+	documents "github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/documents/document"
+	settings "github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/api/documents/document/settings"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider/config"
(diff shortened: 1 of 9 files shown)


datasources/documents/document/data_source.go
Kirdock, line 1: Seems like the new values are actually not set in the data source

Image

config used:
```tf
resource "dynatrace_document" "this" {
  type           = "dashboard"
  name           = "Example Dashboard (local test)"
  description    = "Initial description"
  labels = ["monitoring", "cloud", "draft"]
  is_reshareable = true
  content = jsonencode({})
}

data "dynatrace_documents" "example" {
  type = "dashboard"
  depends_on = [
    dynatrace_document.this
  ]
}

output "documents" {
  value = [for v in data.dynatrace_documents.example.values : v if v.id == dynatrace_document.this.id]
}
```

Seems like the list endpoint is not returning these values and at least `isReshareable` can't be added via the `add-fields` query parameter. So I would suggest just returning the IDs in the List for the documents and iterating over them as there doesn't seem to be another way

dynatrace/api/documents/document/settings/document.go
```diff
@@ -90,45 +109,57 @@ func (me *Document) MarshalHCL(properties hcl.Properties) error {
...
+		"type":           &me.Type,
+		"owner":          &me.Owner,
+		"version":        &me.Version,
+		"description":    &me.Description,
+		"labels":         &me.Labels,
+		"is_reshareable": &me.IsReshareable,
 	})
 }
 
 func (me *Document) MarshalJSON() ([]byte, error) {
```
Kirdock: Because we are assigning the values 1:1 instead of marshaling, we can actually get rid of this one, right?

dynatrace/api/documents/document/service.go
```diff
@@ -131,21 +147,22 @@ func (me *service) Validate(_ *documents.Document) error {
...
-	if !v.IsPrivate {
-		if err = me.update(ctx, stub.ID, v); err != nil {
-			return nil, err
-		}
-	}
-	return stub, nil
+	return me.createPrivate(ctx, v)
 }
 
 func (me *service) createPrivate(ctx context.Context, v *documents.Document) (stub *api.Stub, err error) {
```
arthurpitman: Now that this doesn't always create a private document, we should rename this:

```suggestion
func (me *service) create(ctx context.Context, v *documents.Document) (stub *api.Stub, err error) {
```

go.sum
```diff
@@ -104,8 +104,12 @@ github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSs
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc h1:U9qPSI2PIWSS1VwoXQT9A3Wy9MM3WgvqSxFWenqJduM=
 github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
-github.com/dynatrace/dynatrace-configuration-as-code-core v0.9.1-0.20260707095607-540d9090df78 h1:Blet3gMWA58QgUKFx9gSh//wUBgMX4oO3lCpvf7X91I=
-github.com/dynatrace/dynatrace-configuration-as-code-core v0.9.1-0.20260707095607-540d9090df78/go.mod h1:KhaE60mnB3XYk2dlMnHzvvJ2OYXSiaMXX4BLZkWuwLo=
+github.com/dynatrace/dynatrace-configuration-as-code-core v0.9.1-0.20260804134354-22208c619fec h1:uLv9i3twuTGXlXsVBPmbW1mYIdol1CRZifZ2yPf4u8U=
```
d0weinberger: There a many corelib entries here. 
I think you should run `go mod tidy`

dynatrace/api/documents/document/service.go
```diff
@@ -154,15 +177,25 @@ func (me *service) createPrivate(ctx context.Context, v *documents.Document) (st
...
 func (me *service) update(ctx context.Context, id string, v *documents.Document) (err error) {
-	_, err = me.client.Update(ctx, id, v.Name, v.IsPrivate, []byte(v.Content), docclient.DocumentType(v.Type))
+	meta := docclient.Metadata{
+		ID:        id,
+		Name:      v.Name,
+		IsPrivate: v.IsPrivate,
+		Type:      v.Type,
+		Labels:    v.Labels,
+	}
+	if v.Description != "" {
```
d0weinberger: Can you still delete a description then? I.e. is it deleted because we send null (or don't send it at all) or would you need to allow an empty string to be set here to delete it?
It might make sense to have a test for this scenario

d0weinberger (approved):

PR #1213, opened 2026-08-14
feat: support platform credentials for several classic resources

Kirdock:
#### **Why** this PR?
Some resources on config V1 and environment API V1 support OAuth and platform tokens. We should support that.

#### **What** has changed?
Adds OAuth and platform token support for the following resources
- dynatrace_ag_token
- dynatrace_custom_service, dynatrace_custom_service_order
- dynatrace_request_naming, dynatrace_request_namings
- dynatrace_request_attribute

and data sources
- ~~dynatrace_api_token / dynatrace_api_tokens~~ support not added for now
- dynatrace_lambda_agent_version
- dynatrace_request_attribute
- dynatrace_request_naming

#### **How** does it do it?
By making use of the hybrid client but by disabling the environment variable `DYNATRACE_HTTP_OAUTH_PREFERENCE`, so that previous configurations are working as before.

#### How is it **tested**?
New integration tests added that test the affected resources with platform and classic credentials.

They can use platform credentials to manage the mentioned resources

**Issue:** PS-49501

### DIFF
--- a/dynatrace/rest/classic_hybrid_client.go
+++ b/dynatrace/rest/classic_hybrid_client.go
@@ -19,6 +19,8 @@ package rest
 
 import (
 	"context"
+
+	"github.com/dynatrace-oss/terraform-provider-dynatrace/provider/envutils"
 )
 
 // ClassicHybridClient prefers classic over platform requests and doesn't consider the DYNATRACE_HTTP_OAUTH_PREFERENCE env variable.
@@ -53,5 +55,5 @@ func (me *classic_hybrid_client) Delete(ctx context.Context, url string, expecte
 }
 
 func newDisableOAuthPreferenceContext(ctx context.Context) context.Context {
-	return context.WithValue(ctx, "DYNATRACE_HTTP_OAUTH_PREFERENCE", false)
+	return context.WithValue(ctx, envutils.DynatraceHTTPOAuthPreference.Key, false)
 }
--- a/dynatrace/rest/classic_hybrid_client.go
+++ b/dynatrace/rest/classic_hybrid_client.go
@@ -1,0 +1,57 @@
+/**
+* @license
+* Copyright 2026 Dynatrace LLC
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+ */
+
+package rest
+
+import (
+	"context"
+)
+
+// ClassicHybridClient prefers classic over platform requests and doesn't consider the DYNATRACE_HTTP_OAUTH_PREFERENCE env variable.
+func ClassicHybridClient(clientSet ClientSet) Client {
+	credentials := clientSet.Credentials()
+	return &classic_hybrid_client{client: HybridClient(credentials), credentials: credentials}
+}
+
+type classic_hybrid_client struct {
+	client      Client
+	credentials *Credentials
+}
+
+func (me *classic_hybrid_client) Credentials() *Credentials {
+	return me.credentials
+}
+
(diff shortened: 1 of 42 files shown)


Kirdock:
*Converting to draft. Credentials need to be updated with new scopes

dynatrace/rest/classic_hybrid_client.go
```diff
@@ -0,0 +1,57 @@
...
+func (me *classic_hybrid_client) Put(ctx context.Context, url string, payload any, expectedStatusCodes ...int) Request {
+	return me.client.Put(newDisableOAuthPreferenceContext(ctx), url, payload, expectedStatusCodes...)
+}
+
+func (me *classic_hybrid_client) Delete(ctx context.Context, url string, expectedStatusCodes ...int) Request {
+	return me.client.Delete(newDisableOAuthPreferenceContext(ctx), url, expectedStatusCodes...)
+}
+
+func newDisableOAuthPreferenceContext(ctx context.Context) context.Context {
+	return context.WithValue(ctx, "DYNATRACE_HTTP_OAUTH_PREFERENCE", false)
```
d0weinberger: 1. I would use `envutils.DynatraceHTTPOAuthPreference.Key`  instead of the hardcoded string and 2. I would combine this helper with `NewPreferOAuthContext` in `hybrid_client` to one helper which gets provided a boolean var and sets http oauth preference accordingly

dynatrace/settings/service_options.go
```diff
@@ -25,26 +25,27 @@ import (
...
+	CreateConfirm    int
+	OnChanged        func(context.Context, rest.Client, string, T) error
+	OnBeforeUpdate   func(id string, v T) error
+	HasNoValidator   bool
+	Name             func(id string, v T) (string, error)
+	HijackOnCreate   func(ctx context.Context, err error, service RService[T], v T) (*api.Stub, error)
+	Lock             func()
+	Unlock           func()
+	Duplicates       func(ctx context.Context, service RService[T], v T) (*api.Stub, error)
+	UseClassicHybrid bool // Whether to use classic hybrid client for this service. If false, the default ApiTokenClient will be used.
```
d0weinberger: This is only needed to differentiate in `NewAPITokenService` whether an APITokenClient or a ClassicHybridClient should be used. 

I think the cleaner change would be to remove this `UseClassicHybrid` flag again and offer a `NewClassicHybridService` func in `default_service.go`

dynatrace/settings/default_service.go
```diff
@@ -33,7 +33,13 @@ import (
 var ExportRunning = false
 
 func NewAPITokenService[T Settings](clientSet rest.ClientSet, schemaID string, options *ServiceOptions[T]) (CRUDService[T], error) {
```
d0weinberger, line 35: Suggestion:


```suggestion
func NewAPITokenService[T Settings](clientSet rest.ClientSet, schemaID string, options *ServiceOptions[T]) (CRUDService[T], error) {
	return NewServiceWithClient(rest.APITokenClient(clientSet.Credentials()), schemaID, options)
}

func NewClassicHybridService[T Settings](clientSet rest.ClientSet, schemaID string, options *ServiceOptions[T]) (CRUDService[T], error) {
	return NewServiceWithClient(rest.ClassicHybridClient(clientSet), schemaID, options)
}

func NewServiceWithClient[T Settings](client rest.Client, schemaID string, options *ServiceOptions[T]) (CRUDService[T], error) {
	return &defaultService[T]{schemaID: schemaID, client: client, options: options}, nil
}
```

because then you can get rid of `options.UseClassicHybrid` and in `requestnaming.Service` you could potentially reuse the client, e.g.

```
client := rest.ClassicHybridClient(clientSet)

svc, err := settings.NewServiceWithClient(
	client,
	SchemaID,
	settings.DefaultServiceOptions[*requestnaming.RequestNaming](BasePath),
)
if err != nil {
	return nil, err
}

return &service{service: svc, client: client}, nil
```

dynatrace/settings/service_options.go
Kirdock: ~~But this would remove all the logic that we have in the API token service, which I want to prevent.~~
nvm, got the intention with the example below 👍

dynatrace/rest/classic_hybrid_client.go
Kirdock: Theoretically, it's mixing two different things here (env variables and context keys; but it's against the separation of concerns), but I get the point.
Adjusted for all (also old) usages

d0weinberger (approved):

PR #1215, opened 2026-08-14
feat(rest): add package for Workload Identity Federation tokens

d0weinberger:
#### **Why** this PR?

Preparing WIF support

#### **What** has changed?
Added`rest/wif` package containing:
- `jwt` for extracting the expiry from a token
- `config` defines a config containig WIF vendor, audience, and optionally a pre-minted token for debugging purposes.
- `validate` for validating a config
- `minter` for defining a minter interface and as entry point for creating minters for different vendors
- `github` implements the OIDC token minter for GitHub
- `token_source` mints tokens on demand, handles expiry
- `cache` caches token sources. Necessary, since every `platform_request` creates its own platform client, but we do not want to create a new token source every time this happens.
- `errors` containing `ConfigError` marking errors the user can only fix by changing the provider config.

#### **How** does it do it?
- Added two new entries in `env_vars.go` for GitHub Action-provided vars`ActionsIDTokenRequestURL` and `ActionsIDTokenRequestToken`
- `TokenSource` extends `oauth2.TokenSource` with `Replace(rejected string) (*oauth2.Token, error)`, so a caller that has a token turned down can ask for a different one.
- No existing behaviour is touched - apart from `env_vars.go` all the files in the diff are new.
- Sources are cached per `Config.CacheKey()` (vendor + audience) for the life of the process, since a platform client is built per request.
- The package deliberately imports neither the surrounding `rest` package nor `http.DefaultTransport`: the first would cycle and would put token-service responses - which are the tokens themselves - into the provider's HTTP logging; the second would let `DYNATRACE_HTTP_INSECURE` disable certificate verification on the call carrying the minting credentials.

#### How is it **tested**?
- unit tests
- Concurrency is tested in `TestTokenSourceMintsOnceForConcurrentCallers` and `TestTokenSourceReplacesOnceForConcurrentRejections` 

- No user-facing change. Nothing imports the package yet, no provider attribute or environment variable is read during a plan or apply, and no existing code path is modified.

**Issue:** PS-49502


dynatrace/rest/wif/http_client.go
```diff
@@ -0,0 +1,87 @@
...
+		},
+	}
+}
+
+// retryTransport is local rather than the provider's RetryTransport because this package must not
+// import the rest package. Requests to a token service are bodyless, so there is nothing to replay.
+type retryTransport struct {
+	base    http.RoundTripper
+	backoff time.Duration
+}
```
Kirdock: Why can't we just reuse our existing retryTransport? The body argument doesn't hold as this is catched in the retry_transport (only replay if set).
I also don't get the "must not import the rest package". If this is a cyclic dependency issue then we can just move it to another package.

I would prefer DRY here. Aligned error handling, logging, etc.

dynatrace/rest/wif/errors.go
```diff
@@ -0,0 +1,29 @@
...
+// ConfigError marks a failure the user can only resolve by changing the provider configuration or
+// the CI job, as opposed to a token service being temporarily unreachable. That distinction is what
+// lets the provider report these while it is being configured rather than on the first request.
+type ConfigError struct {
+	Message string
+}
+
+func (err ConfigError) Error() string {
+	return err.Message
+}
```
Kirdock: Do we have a use case for this one? If we don't make any `errors.As`, there is no need for a different struct compared to a simple `errors.New`.

dynatrace/rest/wif/validate.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Validate reports the configuration problems that can be judged without contacting a token service.
+// Requesting no Workload Identity Federation at all is not one of them: only a caller that requires
+// platform credentials can decide whether its absence is a problem.
+func Validate(config Config) error {
```
Kirdock, line 29: Shouldn't some validation happen on the provider level? Like the vendor that should definitely be handled already in the provider schema, right?

dynatrace/rest/wif/config.go
```diff
@@ -0,0 +1,44 @@
...
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+ */
+
+// This package must not import the surrounding rest package: that would put the responses of a token
+// service into the provider's HTTP logging, and those responses are the tokens themselves.
```
Kirdock: I mean, the logging is only applied when calling the factory "of the core client" with `WithHTTPListener`, right?

dynatrace/rest/wif/cache.go
```diff
@@ -0,0 +1,67 @@
...
+// - the day those clients outlive the request they were built for, it can go.
+var (
+	tokenSourceCacheMutex sync.Mutex
+	tokenSourceCache      = map[string]TokenSource{}
+)
+
+// TokenSourceFor returns the token source for the given configuration, creating it on first use.
+//
+// Callers must not wrap the result in an [oauth2.ReuseTokenSource]: that caches until ten seconds
+// before the token's stated expiry and would override the refresh margin the source keeps.
```
Kirdock, line 42: Is this the reason why `ReuseTokenSource` can't be used? If yes, there would be `ReuseTokenSourceWithExpiry` that would handle it, right?

Kirdock: - wif staticTokenSource (struct + token) could be the oauth2.StaticTokenSource.
- wif tokenSource mutex + token + refreshAt: handled by reuseTokenSource
- wif token (cache, single-flight, etc.): ReuseTokenSourceWithExpiry(nil, mintingSrc, refreshMargin)

With this, we could remove a lot of the implementation in token_source.go, right?

Kirdock: On the other hand, the "Replace" in the token_source.go might be the only thing why the oauth2 lib can't be used, right?
Might be a thing to double-check, if we really need it or if it's supported.

dynatrace/rest/wif/cache.go
```diff
@@ -0,0 +1,67 @@
...
+		return source, nil
+	}
+
+	minter, err := newMinter(config, mintingHTTPClient())
+	if err != nil {
+		// Not cached: a cached one would outlive a fix to the environment.
+		return nil, err
+	}
+
+	source := &tokenSource{mintContext: context.Background(), minter: minter}
```
Kirdock: I assume the context.Background is used here because the one provided at provider config level doesn't last long, right?

dynatrace/rest/wif/minter.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// newMinter builds the minter for the configured vendor, without network access, so that missing
+// credentials are reported while the provider is configured rather than on the first request.
+func newMinter(config Config, httpClient *http.Client) (minter, error) {
+	switch config.Vendor {
+	case VendorGitHub:
+		return newGitHubMinter(config.Audience, httpClient)
+	default:
+		return nil, ConfigError{fmt.Sprintf("`%s` is not a supported Workload Identity Federation vendor. The only supported value for `wif_vendor` (`DYNATRACE_WIF_VENDOR`) is `%s`", config.Vendor, VendorGitHub)}
```
Kirdock: So this (and other places) already tells how the provider schema should look like.
I would propose to have a `wif` block instead.
```hcl
wif {
  vendor =...
  // github {...}
}
```

dynatrace/rest/wif/http_client.go
d0weinberger: > also don't get the "must not import the rest package". If this is a cyclic dependency issue then we can just move it to another package.

Yeah, it's going to be a cyclic dependency issue with subsequent changes. I will use the existing retryTransport and move it to a different package, and adapt it a bit.

dynatrace/rest/wif/errors.go
d0weinberger: Fair. I'll get rid of it

dynatrace/rest/wif/config.go
d0weinberger: True, the comment doesn't fit. Removed it

dynatrace/rest/wif/cache.go
d0weinberger: There might be a way, though. I will for now just optimistically add a ctx parameter to the `TokenSourceFor` func and we can later see what to pass. If we don't find anything fitting, it's back to the solution above.

dynatrace/rest/wif/validate.go
d0weinberger: Yes, it will be handled there as well. There still needs to be this validation because export doesn't run the schema validation. There is credentials validation also inlined in `config.go` for other credential types. Here it's just not inlined.

d0weinberger:
https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/1229

PR #1222, opened 2026-08-19
Retry transient settings requests safely

viasnake:
#### **Why** this PR?

Creating Settings 2.0 objects can fail intermittently under parallel apply.

When API token authentication is used, the provider first sends the request to the classic Settings API. A `400` response indicating that OAuth is required is expected, so the provider retries the request against the Platform Settings endpoint.

Some Platform POST requests terminate with `unexpected EOF` before an HTTP response is received. The transport currently treats this as a hard failure, although a later retry often succeeds.

#### **What** has changed?

- Retry transient `io.EOF`, `io.ErrUnexpectedEOF`, and connection-reset errors.
- Retry these errors automatically for idempotent HTTP methods.
- Explicitly allow retries for Settings 2.0 object POST requests.
- Buffer and replay request bodies across retry attempts.
- Apply the retry-enabled HTTP client to platform-token clients as well as OAuth clients.
- Reuse the same `externalId` during a logical Settings create and its internal retries.
- Preserve the same `externalId` when the generic Settings service falls back from the classic API to the Platform API.

#### **How** does it do it?

`RetryTransport` classifies EOF and connection-reset errors as transient transport failures while excluding timeout errors and unrelated failures.

Non-idempotent requests are not retried by default. The Platform Settings object endpoint explicitly opts in because its create operation is designed to reuse the same `externalId` when replayed.

The request body is buffered before the first attempt so that every retry sends the same payload.

The existing OAuth fallback behavior remains unchanged: the expected classic API `400` causes the generic Settings service to retry using the Platform endpoint.

#### How is it **tested**?

- `go test -tags=unit ./...`
- `go vet -tags=unit ./dynatrace/rest ./dynatrace/api/builtin/generic ./dynatrace/settings/services/settings20`
- `git diff --check`

Unit tests cover transient transport errors, retry limits, context cancellation, non-idempotent request opt-in, and request body replay.

No live Dynatrace acceptance test was run.

Transient connection failures during Settings 2.0 creation can now recover automatically instead of failing immediately.

Normal POST requests remain non-retriable unless they explicitly opt in, reducing the risk of duplicate side effects.

The generated `externalId` is reused within one logical Create operation but is not persisted across separate Terraform runs. This change also does not introduce a provider-wide HTTP timeout.

The `externalId` behavior follows the Settings API’s caller-provided object replacement model:

- [Dynatrace Settings API key concepts](https://docs.dynatrace.com/docs/dynatrace-api/environment-api/settings/key-concepts)
- [POST Settings object](https://docs.dynatrace.com/docs/dynatrace-api/environment-api/settings/objects/post-object)

**Issue:** N/A

### DIFF
--- a/dynatrace/api/builtin/generic/service.go
+++ b/dynatrace/api/builtin/generic/service.go
@@ -31,6 +31,7 @@ import (
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/settings"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/settings/services/settings20"
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/shutdown"
+	"github.com/google/uuid"
 )
 
 const settingsObjectEndpoint = "/api/v2/settings/objects"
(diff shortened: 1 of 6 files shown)


arthurpitman:
Thank you @viasnake for bringing this serious issue to our attention.

I think, however, we need to fix this at the source (presumably the underlying API), so could I ask you to reach out to [Dynatrace Support](https://support.dynatrace.com/) and work with them to reproduce it?

While I really appreciate the effort you've put in to handling it here, if we can fix the root cause, we actually can avoid the complexity of the retry logic, which would be my preferred option.

viasnake:
Thanks for the feedback. I agree that fixing the root cause would be preferable.
I'll reach out to Dynatrace Support and work with them to reproduce the issue.

PR #1224, opened 2026-08-24
feat: resource updates for 345

Kirdock:
#### **Why** this PR?
Some APIs have changed with Dynatrace version 343. Therefore, the Terraform resources also need updating.

#### **What** has changed?
- `dynatrace_davis_copilot`: New fields: `pii_blocking_enabled` (default: false) and `pii_blocking_types` (if `pii_blocking_enabled` is set to `true`)
- `dynatrace_maintenance_windows`: New optional field "object_scopes", which contains settings for pausing specific synthetic monitors during the maintenance window
- `dynatrace_kubernetes_enrichment`: New optional "use_ingest_enrichment_config_schema" field.
- `dynatrace_azure_connection` and `dynatrace_aws_connection`: Remove the deprecated enums for the consumers field: "NONE" and "DA" (Breaking change)

Documentation and fixes for conditional field handling

#### **How** does it do it?

#### How is it **tested**?

**Issue:** PS-49923

### DIFF
--- a/dynatrace/api/builtin/hyperscalerauthentication/connections/aws/settings/enums.go
+++ b/dynatrace/api/builtin/hyperscalerauthentication/connections/aws/settings/enums.go
@@ -20,15 +20,11 @@ package aws
 type ConsumersOfAwsRoleBasedAuthentication string
 
 var ConsumersOfAwsRoleBasedAuthentications = struct {
-	Da                          ConsumersOfAwsRoleBasedAuthentication
-	None                        ConsumersOfAwsRoleBasedAuthentication
 	SvcComDynatraceBo           ConsumersOfAwsRoleBasedAuthentication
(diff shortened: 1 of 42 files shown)


dynatrace/api/builtin/hyperscalerauthentication/connections/aws/settings/enums.go
```diff
@@ -1,6 +1,6 @@
 /**
 * @license
-* Copyright 2025 Dynatrace LLC
+* Copyright 2020 Dynatrace LLC
```
tomazpu: 2020?

Kirdock: Fixed 👍

tomazpu (approved):

PR #1227, opened 2026-08-25
refactor: Add remaining clients to clientset

arthurpitman:
#### **Why** this PR?
The provider still had clients being created from raw credentials in many places, and tests were often constructing `config.ProviderConfiguration` directly in incomplete ways. This refactor finishes moving client creation behind the shared `ClientSet`, clarifies which URL is the classic environment URL, and removes request plumbing that only existed to support test-specific behavior.

- `provider/config` and `dynatrace/rest` now build and reuse the full client set through `config.ProviderConfigureGeneric(...)`, so API token, hybrid, classic platform, and cluster requests are all created from `rest.ClientSet` instead of re-deriving clients from credentials.
- `rest.Credentials.URL` was renamed to `ClassicEnvironmentURL`, and callers that depend on the classic base URL were updated accordingly.
- The remaining services that still called `rest.APITokenClient(clientSet.Credentials())`, `rest.HybridClient(clientSet.Credentials())`, `rest.ClusterV1Client(clientSet.Credentials())`, or `rest.ClusterV2Client(clientSet.Credentials())` were migrated to pass the shared `ClientSet` directly.
- Request handling in `dynatrace/rest` was simplified by removing unnecessary header plumbing and test-only behavior, including the old `rest.classic_request` path and explicit header mutation in extension requests.
- Tests and test helpers were aligned with the new configuration path so they use the same provider setup as production code instead of manually instantiating partial provider configs.

#### How is it **tested**?
- Validation and service tests were updated to use `config.ProviderConfigureGeneric(...)`
- Client behavior tests were reworked in `dynatrace/rest/hybrid_client_test.go`

- This PR is mainly refactoring. However, performance will be improved by potentially re-using the platform OAuth client, rather than re-creating it for each operation.

**Issue:** PS-48500

### DIFF
--- a/dynatrace/api/v2/hub/extension/config/service.go
+++ b/dynatrace/api/v2/hub/extension/config/service.go
@@ -48,7 +48,7 @@ func (me *service) Get(ctx context.Context, id string, v *extension_config.Setti
 	name, configurationID := splitID(id)
 
 	var response GetMonitoringConfigurationResponse
-	client := rest.APITokenClient(me.clientSet.Credentials())
+	client := rest.APITokenClient(me.clientSet)
 
 	urlPath := fmt.Sprintf(
 		"/api/v2/extensions/%s/monitoringConfigurations/%s",
@@ -95,7 +95,7 @@ func (me *service) List(ctx context.Context) (api.Stubs, error) {
 	var stubs api.Stubs
 
 	var extensionsList ExtensionsList
-	client := rest.APITokenClient(me.clientSet.Credentials())
+	client := rest.APITokenClient(me.clientSet)
 
 	nextPageKey := "first"
 	for len(nextPageKey) > 0 {
@@ -143,7 +143,7 @@ func (me *service) Create(ctx context.Context, v *extension_config.Settings) (*a
 	if err := me.ensureInstalled(ctx, name, version); err != nil {
 		return nil, err
 	}
-	client := rest.APITokenClient(me.clientSet.Credentials())
+	client := rest.APITokenClient(me.clientSet)
 	createResponse := []CreateMonitoringConfigResponse{}
 	retry := 10
 	for retry > 0 {
@@ -166,7 +166,7 @@ func (me *service) Create(ctx context.Context, v *extension_config.Settings) (*a
 }
 
 func (me *service) ensureInstalled(ctx context.Context, name string, version string) error {
-	client := rest.APITokenClient(me.clientSet.Credentials())
+	client := rest.APITokenClient(me.clientSet)
 	if strings.HasPrefix(name, "custom:") {
 		return client.Get(ctx, fmt.Sprintf("/api/v2/extensions/%s/%s", url.PathEscape(name), url.QueryEscape(version)), 200).Finish()
 	}
@@ -195,7 +195,7 @@ func (me *service) Update(ctx context.Context, id string, v *extension_config.Se
 	if err := me.ensureInstalled(ctx, name, version); err != nil {
 		return err
 	}
-	client := rest.APITokenClient(me.clientSet.Credentials())
+	client := rest.APITokenClient(me.clientSet)
 	createResponse := CreateMonitoringConfigResponse{}
 	payload := MonitoringConfigCreateDto{Value: []byte(v.Value)}
 	if err := client.Put(ctx, fmt.Sprintf("/api/v2/extensions/%s/monitoringConfigurations/%s", url.PathEscape(name), url.PathEscape(configID)), &payload, 200).Finish(&createResponse); err != nil {
@@ -209,7 +209,7 @@ func (me *service) Update(ctx context.Context, id string, v *extension_config.Se
 
 func (me *service) Delete(ctx context.Context, id string) error {
 	name, configID := splitID(id)
-	client := rest.APITokenClient(me.clientSet.Credentials())
+	client := rest.APITokenClient(me.clientSet)
 	if err := client.Delete(ctx, fmt.Sprintf("/api/v2/extensions/%s/monitoringConfigurations/%s", url.PathEscape(name), url.PathEscape(configID)), 200).Finish(nil); err != nil {
 		// Potential response when the configuration contains
 		//    import {
--- a/dynatrace/api/v2/hub/extension/config/service.go
+++ b/dynatrace/api/v2/hub/extension/config/service.go
@@ -168,9 +168,7 @@ func (me *service) Create(ctx context.Context, v *extension_config.Settings) (*a
 func (me *service) ensureInstalled(ctx context.Context, name string, version string) error {
 	client := rest.APITokenClient(me.clientSet.Credentials())
 	if strings.HasPrefix(name, "custom:") {
-		request := client.Get(ctx, fmt.Sprintf("/api/v2/extensions/%s/%s", url.PathEscape(name), url.QueryEscape(version)), 200)
-		request.SetHeader("Accept", "application/json; charset=utf-8")
-		return request.Finish()
+		return client.Get(ctx, fmt.Sprintf("/api/v2/extensions/%s/%s", url.PathEscape(name), url.QueryEscape(version)), 200).Finish()
 	}
 	response := struct {
 		Name    string `json:"extensionName"`
--- a/dynatrace/rest/api_token_client.go
+++ b/dynatrace/rest/api_token_client.go
@@ -22,7 +22,6 @@ import (
 	"errors"
 	"net/http"
 	"net/url"
-	"strings"
 	"sync"
 
 	"github.com/dynatrace-oss/terraform-provider-dynatrace/dynatrace/rest/logging"
@@ -56,7 +55,7 @@ func (me *api_token_client) Get(ctx context.Context, url string, expectedStatusC
 }
 
 func (me *api_token_client) Post(ctx context.Context, url string, payload any, expectedStatusCodes ...int) Request {
-	req := &api_token_request{id: uuid.NewString(), ctx: ctx, client: me, url: url, method: http.MethodPost, payload: payload, headers: map[string]string{"Content-Type": "application/json"}}
+	req := &api_token_request{id: uuid.NewString(), ctx: ctx, client: me, url: url, method: http.MethodPost, payload: payload}
 	if len(expectedStatusCodes) > 0 {
 		req.expect = statuscodes(expectedStatusCodes)
 	}
@@ -64,7 +63,7 @@ func (me *api_token_client) Post(ctx context.Context, url string, payload any, e
 }
 
 func (me *api_token_client) Put(ctx context.Context, url string, payload any, expectedStatusCodes ...int) Request {
-	req := &api_token_request{id: uuid.NewString(), ctx: ctx, client: me, url: url, method: http.MethodPut, payload: payload, headers: map[string]string{"Content-Type": "application/json"}}
+	req := &api_token_request{id: uuid.NewString(), ctx: ctx, client: me, url: url, method: http.MethodPut, payload: payload}
 	if len(expectedStatusCodes) > 0 {
 		req.expect = statuscodes(expectedStatusCodes)
 	}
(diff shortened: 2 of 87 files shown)


dynatrace/api/v2/hub/extension/config/service.go
```diff
@@ -166,11 +166,9 @@ func (me *service) Create(ctx context.Context, v *extension_config.Settings) (*a
 }
 
 func (me *service) ensureInstalled(ctx context.Context, name string, version string) error {
-	client := rest.APITokenClient(me.clientSet.Credentials())
+	client := rest.APITokenClient(me.clientSet)
 	if strings.HasPrefix(name, "custom:") {
-		request := client.Get(ctx, fmt.Sprintf("/api/v2/extensions/%s/%s", url.PathEscape(name), url.QueryEscape(version)), 200)
-		request.SetHeader("Accept", "application/json; charset=utf-8")
```
d0weinberger, line 172: Why is this line not needed anymore?
The content-type header is removed from the clients because it is added to every request in the corelib anyways. But with the `Accept` header it's different, right?

dynatrace/rest/hybrid_client_test.go
```diff
@@ -19,10 +19,9 @@
 
 package rest
 
+/*
```
d0weinberger: Why is this all commented out?

dynatrace/testing/environmentgetter.go
```diff
@@ -0,0 +1,57 @@
...
+var _ config.Getter = &environmentGetter{}
+
+type environmentGetter struct {
+	envURL   string
+	apiToken string
+}
+
+// EnvironmentGetter returns a config.Getter that retrieves the Dynatrace environment URL and API token from environment variables.
+func EnvironmentGetter(t *testing.T) *environmentGetter {
+	envURL := os.Getenv("DYNATRACE_ENV_URL")
```
Kirdock: Should we use `DynatraceEnvURL.Get` from `env_vars.go` instead?
Also for the api token

provider/config/config.go
```diff
@@ -45,12 +45,24 @@ type ProviderConfiguration struct {
...
+
+	clusterV2Client    *rest2.Client
+	clusterV2ClientErr error
 }
 
 func (c *ProviderConfiguration) Credentials() *rest.Credentials {
 	credentials := &rest.Credentials{
 		Token:    c.APIToken,
-		URL:      c.EnvironmentURL,
+		ClassicEnvironmentURL:      c.EnvironmentURL,
```
Kirdock: Because of this, could you run go fmt on all the files?

dynatrace/rest/api_token_client.go
```diff
@@ -56,15 +54,15 @@ func (me *api_token_client) Get(ctx context.Context, url string, expectedStatusC
...
+	req := &api_token_request{id: uuid.NewString(), ctx: ctx, client: me, url: url, method: http.MethodPost, payload: payload}
 	if len(expectedStatusCodes) > 0 {
 		req.expect = statuscodes(expectedStatusCodes)
 	}
 	return req
 }
 
 func (me *api_token_client) Put(ctx context.Context, url string, payload any, expectedStatusCodes ...int) Request {
-	req := &api_token_request{id: uuid.NewString(), ctx: ctx, client: me, url: url, method: http.MethodPut, payload: payload, headers: map[string]string{"Content-Type": "application/json"}}
+	req := &api_token_request{id: uuid.NewString(), ctx: ctx, client: me, url: url, method: http.MethodPut, payload: payload}
```
d0weinberger, line 62: It's safe to remove headers because in the corelib, in `setHeadersOnRequest` all request without content-type would anyways get `content-type application/json`right?

dynatrace/rest/hybrid_client_test.go
```diff
@@ -186,81 +254,30 @@ func TestApiTokenClient(t *testing.T) {
...
 	})
-
-	t.Run("Sets correct headers", func(t *testing.T) {
-		customHeaderName := "my-header"
-		customHeaderValue := "my-value"
-		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			expectedURL, err := url.JoinPath(activeGatePostfix, endpoint)
-			require.NoError(t, err)
-			require.Equal(t, expectedURL, r.URL.Path)
-			require.Equal(t, fmt.Sprintf("Api-Token %s", mockToken), r.Header.Get("Authorization"))
```
d0weinberger, line 223: This was the only assertion that the `Api-Token` header is sent. Maybe you could re-add a test or extend one of the existing tests to reintroduce such a check

dynatrace/rest/hybrid_client_test.go
d0weinberger, line 136: duplicate line underneath

provider/config/config.go
```diff
@@ -45,12 +45,24 @@ type ProviderConfiguration struct {
 
 	platformClient    *rest2.Client
 	platformClientErr error
+
+	classicPlatformClient    *rest2.Client
```
Kirdock, line 49: Can we add a comment here and/or to the `client_set.go`?
This client uses OAuth or platform token authentication to make requests on classic URLs.

dynatrace/rest/hybrid_client_test.go
```diff
@@ -21,151 +21,219 @@ package rest
...
-var credential_repo = map[string]Credentials{
-	"unconfigured":                           {URL: TestCaseEnvURL},
-	"api-token":                              {URL: TestCaseEnvURL, Token: mockToken},
-	"oauth":                                  {URL: TestCaseEnvURL, Platform: PlatformCredentials{ClientID: mockToken, ClientSecret: mockToken}},
-	"platform-token":                         {URL: TestCaseEnvURL, Platform: PlatformCredentials{PlatformToken: mockToken}},
-	"api-token-and-oauth":                    {URL: TestCaseEnvURL, Token: mockToken, Platform: PlatformCredentials{ClientID: mockToken, ClientSecret: mockToken}},
-	"api-token-and-platform-token":           {URL: TestCaseEnvURL, Token: mockToken, Platform: PlatformCredentials{PlatformToken: mockToken}},
-	"oauth-and-platform-token":               {URL: TestCaseEnvURL, Platform: PlatformCredentials{ClientID: mockToken, ClientSecret: mockToken, PlatformToken: mockToken}},
-	"api-token-and-oauth-and-platform-token": {URL: TestCaseEnvURL, Token: mockToken, Platform: PlatformCredentials{PlatformToken: mockToken, ClientID: mockToken, ClientSecret: mockToken}},
+type mockClientSet struct {
```
Kirdock: Can't we reuse the already existing mock from `mockclientset.go`?

dynatrace/rest/api_token_client.go
```diff
@@ -35,16 +33,16 @@ import (
 
 var NoClassicURLDefinedErr = errors.New("no Environment URL has been specified. Use either the environment variable `DYNATRACE_ENV_URL` or the configuration attribute `dt_env_url` of the provider for that")
```
Kirdock: This is a not used variable now.

dynatrace/rest/hybrid_client_test.go
```diff
@@ -186,81 +254,30 @@ func TestApiTokenClient(t *testing.T) {
...
 	t.Run("Errors on empty env URL", func(t *testing.T) {
-		cred := Credentials{URL: "", Token: mockToken}
-		client := HybridClient(&cred)
+		cred := Credentials{ClassicEnvironmentURL: "", Token: mockToken}
+		client := HybridClient(createMockClientSet(t, &cred))
 
 		req := client.Get(t.Context(), endpoint)
 		err := req.Finish()
-		assert.ErrorIs(t, err, NoClassicURLDefinedErr)
+		assert.ErrorContains(t, err, "no classic API URL provided")
```
Kirdock, line 157: I guess this one is coming from the core-lib? Just asking because `NoClassicURLDefinedErr` is now not used anymore

dynatrace/rest/hybrid_client_test.go
```diff
@@ -21,151 +21,219 @@ package rest
...
+	t.Run("unconfigured with OAuth preferred", func(t *testing.T) {
+		ctx := context.WithValue(t.Context(), envutils.DynatraceHTTPOAuthPreference.Key, true)
+		err := HybridClient(createMockClientSet(t, &Credentials{})).Get(ctx, "").Finish()
+		assert.ErrorIs(t, err, NoOAuthCredentialsError)
+	})
+
+	t.Run("API token configured without OAuth preferred", func(t *testing.T) {
+		v := struct {
+			Classic bool `json:"classic"`
+		}{}
```
Kirdock: As all the structs for the responses here are reused we should define them once above.

Still, I'm also thinking about the implementation.
We could check the auth header in the server and respond with the used auth.

e.g., 

```go
func newHybridTestServer(t *testing.T) (baseURL, tokenURL string) {
	writeJSON := func(w http.ResponseWriter, body string) {
		w.Header().Set("Content-Type", "application/json")
		_, err := w.Write([]byte(body))
		require.NoError(t, err)
	}
	mux := http.NewServeMux()
	mux.HandleFunc("/sso/oauth2/token", func(w http.ResponseWriter, r *http.Request) {
		writeJSON(w, `{"access_token":"tok","token_type":"Bearer","expires_in":3600}`)
	})
	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		if strings.HasPrefix(r.Header.Get("Authorization"), "Api-Token") {
			writeJSON(w, `{"classic": true}`)
			return
		}
		writeJSON(w, `{"platform": true}`)
	})
	srv := httptest.NewServer(mux)
	t.Cleanup(srv.Close)
	return srv.URL, srv.URL + "/sso/oauth2/token"
}
```

Then we can have the table test again

```go
			ctx := t.Context()
			if tt.oauthPreferred {
				ctx = context.WithValue(ctx, envutils.DynatraceHTTPOAuthPreference.Key, true)
			}
			var v struct {
				Classic  bool `json:"classic"`
				Platform bool `json:"platform"`
			}
			err := HybridClient(createMockClientSet(t, tt.creds)).Get(ctx, "").Finish(&v)

			if tt.wantErr != nil {
				assert.ErrorIs(t, err, tt.wantErr)
				return
			}
			require.NoError(t, err)
			assert.Equal(t, tt.want == "classic", v.Classic)
			assert.Equal(t, tt.want == "platform", v.Platform)
```

dynatrace/rest/hybrid_client_test.go
arthurpitman: Was a rebase gone bad, but already fixed 👍

dynatrace/testing/environmentgetter.go
d0weinberger: +1

dynatrace/api/v2/hub/extension/config/service.go
arthurpitman: So I tested this manually on DEV and PROD: the API returns JSON by default, so I'd say it is not needed, unless you need the actual extension, which we don't. That is the only instance where the functionality is even used. With respect to content type, correct, that is done by the core lib.

dynatrace/testing/environmentgetter.go
arthurpitman: 👍

dynatrace/rest/hybrid_client_test.go
arthurpitman: I have reworked it now, and it does look fine as a table test 😄

dynatrace/rest/hybrid_client_test.go
arthurpitman: True, this works if we move it to `rest_test`

dynatrace/rest/api_token_client.go
arthurpitman: `NoPlatformCredentialsErr` stays for now, but let's rework how credentials are validated in a follow up

provider/config/config.go
arthurpitman: Added to `client_set.go` 👍

dynatrace/rest/hybrid_client_test.go
arthurpitman: Correct

Kirdock (approved):

d0weinberger (approved):

PR #1229, opened 2026-08-26
feat: add wif package

d0weinberger:
#### **Why** this PR?
Preparing WIF support

#### **What** has changed?
Moved `RetryTransport`:
- `dynatrace/rest/retry_transport.go` moved to the leaf package `dynatrace/rest/retry`; `RetryTransport` became `retry.Transport`. `NewContextWithOAuthRetryClient` stayed in `rest`, now in its own `oauth_retry_client.go`.
- `retry.Transport` gained a `ShouldRetry func(*http.Response, error) bool` field, defaulting to `IfTooManyRequestsOrServiceUnavailable` — the previous 429/503-only behaviour.

Added `rest/wif` package containing:
- `jwt` for extracting the expiry from a token
- `config` defines a config containing WIF vendor, audience, and optionally a pre-minted token as a non-refreshable escape hatch for workloads that cannot mint one.
- `validate` for validating a config
- `minter` for defining a minter interface and as entry point for creating minters for different vendors
- `github` implements the OIDC token minter for GitHub
- `http_client` builds the client used to talk to a token service: a `retry.Transport` widened to retry any 5xx and a round trip that never completed, over an explicit `http.Transport` rather than `http.DefaultTransport`.
- `token_source` mints tokens on demand, reads the expiry from the token and lets `oauth2.ReuseTokenSourceWithExpiry` decide when to re-mint. Contains a cache, which is necessary, since every `platform_request` creates its own platform client, but we do not want to create a new token source every time this happens.

- Added two new entries in `env_vars.go` for GitHub Action-provided vars `ActionsIDTokenRequestURL` and `ActionsIDTokenRequestToken`

#### **How** does it do it?

- Sources are cached per `Config.cacheKey()` (vendor + audience) for the life of the process, since a platform client is built per request. A pre-minted token is never cached, so a credential cannot become a key in a process-long map.
- Caching, the two-minute refresh margin and single-flight minting all come from `oauth2.ReuseTokenSourceWithExpiry`; the package only mints and reads the `exp` claim. A pre-minted token becomes an `oauth2.StaticTokenSource` with no expiry, since reporting one would make `oauth2` discard a token that cannot be re-minted.
- `mintingHTTPClient` must not use `http.DefaultTransport`: the provider mutates that one when `DYNATRACE_HTTP_INSECURE` is set, and that escape hatch must not disable certificate verification on the call carrying the minting credentials.
- No signature is verified when reading the `exp` claim, and none is missing — the provider is not the relying party for these tokens, the Dynatrace platform is. The package also imports neither `rest` nor its HTTP logging, since the responses of a token service are the tokens themselves. Error strings name no provider attribute; which attribute sets a value is the provider layer's business.

#### How is it **tested**?
Unit tests 

- No user-facing change. Nothing imports `rest/wif` yet, and no provider attribute or environment variable is read during a plan or apply.
- `retry.Transport` behaves as `RetryTransport` did for existing callers: `ShouldRetry` defaults to the 429/503 check that used to be hardcoded.

**Issue:** PS-49502

### DIFF
--- a/dynatrace/rest/wif/config.go
+++ b/dynatrace/rest/wif/config.go
@@ -1,0 +1,82 @@
+/**
+* @license
+* Copyright 2026 Dynatrace LLC
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+ */
+
+package wif
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+)
+
+var (
+	errNotConfigured        = errors.New("no Workload Identity Federation has been configured")
+	errVendorAndStaticToken = errors.New("a vendor and a pre-minted OIDC token are mutually exclusive, specify only one of them")
+	errStaticTokenNotAJWT   = errors.New("the pre-minted OIDC token is not a JWT: expected three dot-separated segments")
+	errNoAudience           = errors.New("no audience has been specified")
+)
+
+// Vendor identifies the workload identity provider that issues the OIDC token.
+type Vendor = string
+
+// VendorGitHub obtains tokens from the GitHub Actions OIDC token service.
+const VendorGitHub Vendor = "github"
+
+type Config struct {
+	Vendor      Vendor
+	Audience    string
+	StaticToken string
+}
+
+// Configured reports whether any form of Workload Identity Federation was requested.
+func (config Config) Configured() bool {
+	return len(config.Vendor) > 0 || len(config.StaticToken) > 0
+}
+
+// Validate reports whether the configuration describes a usable setup, as far as that can be judged
+// without contacting a token service. Callers for which Workload Identity Federation is optional have
+// to guard with [Config.Configured] - only they can decide whether its absence is a problem.
+func (config Config) Validate() error {
+	if !config.Configured() {
+		return errNotConfigured
+	}
+
+	if len(config.Vendor) > 0 && len(config.StaticToken) > 0 {
+		return errVendorAndStaticToken
+	}
+
+	if len(config.StaticToken) > 0 {
+		if strings.Count(config.StaticToken, ".") != jwtSegmentCount-1 {
+			return errStaticTokenNotAJWT
+		}
+		return nil
+	}
+
+	if config.Vendor != VendorGitHub {
+		return unsupportedVendorError(config.Vendor)
+	}
+
+	if len(config.Audience) == 0 {
+		return errNoAudience
+	}
+
+	return nil
+}
+
+func unsupportedVendorError(vendor Vendor) error {
+	return fmt.Errorf("`%s` is not a supported vendor, the only supported one is `%s`", vendor, VendorGitHub)
+}
--- a/dynatrace/rest/wif/github.go
+++ b/dynatrace/rest/wif/github.go
@@ -1,0 +1,113 @@
+/**
+* @license
+* Copyright 2026 Dynatrace LLC
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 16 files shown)


dynatrace/rest/wif/config.go
```diff
@@ -0,0 +1,40 @@
...
+* limitations under the License.
+ */
+
+package wif
+
+// Vendor identifies the workload identity provider that issues the OIDC token.
+type Vendor string
+
+// VendorGitHub obtains tokens from the GitHub Actions OIDC token service.
+const VendorGitHub Vendor = "github"
```
Kirdock: Some time ago, we aligned that we should use aliases instead of a new string type.
So, would it be possible to use `type Vendor = string` instead? This would also remove the need for "string(myVendor)" and "myVendor" could directly be used

dynatrace/rest/wif/github.go
```diff
@@ -0,0 +1,108 @@
...
+
+	response, err := minter.httpClient.Do(request)
+	if err != nil {
+		return "", fmt.Errorf("failed to get ID token: %w", err)
+	}
+	defer func() { _ = response.Body.Close() }()
+
+	if response.StatusCode != http.StatusOK {
+		// The body is left out: on success it is the ID token itself.
+		return "", fmt.Errorf("failed to get ID token: %s responded with HTTP %d", envutils.ActionsIDTokenRequestURL.Key, response.StatusCode)
```
Kirdock: The key, not the value here?

dynatrace/rest/wif/validate.go
```diff
@@ -0,0 +1,58 @@
...
+	"strings"
+)
+
+// Validate reports whether the configuration describes a usable setup, as far as that can be judged
+// without contacting a token service. Callers for which Workload Identity Federation is optional have
+// to guard with [Config.Configured] - only they can decide whether its absence is a problem.
+func Validate(config Config) error {
+	if !config.Configured() {
+		return errors.New("no Workload Identity Federation has been configured")
+	}
```
Kirdock: What's the reason why this check (if vendor and static token aren't set) is inside the config, and the rest of the validation is here?
As far as I can see, the Configured function is only invoked here.

dynatrace/rest/wif/validate_test.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// Callers for which federation is optional guard with Configured, so absence is a problem by the time
+// Validate is asked.
+func TestValidateRejectsUnconfigured(t *testing.T) {
+	assert.EqualError(t, Validate(Config{}), "no Workload Identity Federation has been configured")
```
Kirdock: [nit] We could put the returned errors into global variables/constants and use them here for the check. This would remove the need for the duplicated error string and we can just use `assert.ErrorIs`.

The same for `dynatrace/rest/wif/github_test.go`

dynatrace/rest/wif/github_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+// A stub rather than a closed port, so the asserted error is the one given to the transport rather
+// than however the operating system words a refused connection.
+type unreachableTransport struct {
+	err error
+}
+
+func (transport *unreachableTransport) RoundTrip(*http.Request) (*http.Response, error) {
+	return nil, transport.err
+}
```
Kirdock: [nit] This is probably to check the error returned for a request. Technically, we could also just provide a canceled context. Then we would not need the transport helper.
```go
	ctx, cancel := context.WithCancel(t.Context())
	cancel()
	_, err = minter.mint(ctx)
    assert.ErrorIs(t, err, context.Canceled)
```

dynatrace/rest/wif/token_source.go
```diff
@@ -0,0 +1,96 @@
...
+// out just before it is due cannot expire in flight, and wide enough to absorb a clock differing from
+// the platform's.
+const refreshMargin = 2 * time.Minute
+
+// A platform request builds its own client, and with it a token source, inside every single call, so
+// without this cache every API call would mint a token of its own.
+var (
+	tokenSourceCacheMutex sync.Mutex
+	tokenSourceCache      = map[string]oauth2.TokenSource{}
+)
```
Kirdock: With the latest changes coming from https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/1227 we can actually get rid of this cache and mutex because we create the clients once, right?

dynatrace/rest/wif/github.go
d0weinberger: actually, the concrete URL is not interesting ... I will just replace it with something generic like `"the token service responded with ..."`

dynatrace/rest/wif/validate.go
d0weinberger: I will merge validate.go and config.go actually

dynatrace/rest/wif/token_source.go
d0weinberger: yes, that's true. I'll remove the cache

Kirdock (approved):

PR #1238, opened 2026-09-09
fix: revert to previous state on error during update

Kirdock:
#### **Why** this PR?
On update the plugin SDK sets the planned value as the "state" value. So, failing an update will change the state, which isn't a problem because you would do a plan (refresh) anyway the next time. BUT if the value that changed is not checked against the API, like secrets, the wrong (not updated) value is in the state and a second apply which would succeed wouldn't even show a state drift (needs update).
https://github.com/hashicorp/terraform-plugin-sdk/issues/476
Why this exists or how it's different to the plugin framework: https://developer.hashicorp.com/terraform/plugin/sdkv2/guides/v2-upgrade-guide#removal-of-helper-schema-resourcedata-setpartial

#### **What** has changed?
The state is not set to the planned state during update if there is an error.
Limitation: The plugin SDK ignores any set values for TypeSet during this, so it's not the whole old state, but it should be enough to tackle the secret limitation.

#### **How** does it do it?
Read and set the old state if there is an error during apply

#### How is it **tested**?
New unit test added.
Manual tests with the credentials resource.

create resource -> apply (update secret) error -> apply => now correctly shows that there is an update needed instead of "no changes found"

**Issue:** NOISSUE


resources/generic.go
```diff
@@ -305,7 +303,18 @@ func (me *Generic) Update(ctx context.Context, d *schema.ResourceData, m any) di
 		if restWarning, ok := err.(rest.Warning); ok {
 			return diag.Diagnostics{diag.Diagnostic{Severity: diag.Warning, Summary: restWarning.Message}}
 		}
-		return toDiagError(err)
+		// On a failed update the SDK would otherwise persist the (rejected)
+		// planned values into state. Restore the prior state so state isn't corrupted.
+		revertErr := me.revertToState(d)
+		diags := diag.Diagnostics{toDiagError(err)}
```
arthurpitman: nit: Anything against swapping these two lines?

arthurpitman (approved):

PR #1239, opened 2026-09-10
feat: ingest schema support as SDLC event 

Kirdock:
#### **Why** this PR?
For easier and nicer statistics, the ingestion of the schema support file as SDLC event would really be benefitial.

#### **What** has changed?
Schema support JSON file is now ingested as SDLC event

#### **How** does it do it?
By updating the release pipeline to ingest the generated Schema support file as SDLC event.

#### How is it **tested**?
Unit tests added.
Local test with a schema support file of the latest release.

N/A

**Issue:** PS-49938


.github/workflows/release.yml
```diff
@@ -7,26 +7,40 @@ jobs:
...
-      -
-        name: Import GPG key
+
+      - name: Generate schema support file
+        run: go run ./tools/schemasupport/
+
+      - name: Generate schema support event file
+        run: go run ./tools/schemaingest/ --file schemaSupport.json --out event.json --release
+
+      - name: 📡 Send test results to Dynatrace
```
d0weinberger:
```suggestion
      - name: 📡 Send schema support events to Dynatrace
```

d0weinberger (approved):

PR #1243, opened 2026-09-11
fix: additional new line when using EOT in export

Kirdock:
#### **Why** this PR?
EOT always adds a newline to a string, which ends in a non-empty plan.

#### **What** has changed?
No non-empty plan anymore (no additional new lines)

#### **How** does it do it?
By adding a `trimsuffix` to the generated HCL that removes the newline introduced via EOT.

#### How is it **tested**?
New tests added

No non-empty plan anymore when exporting configs with strings that contain new lines or are heavily quoted.

**Issue:** PS-50332


terraform/hclgen/primitive_entry.go
```diff
@@ -78,6 +78,26 @@ func finalizeString(s string, indent string) string {
 	return finalString
 }
 
+// wantHeredoc returns true if EOT should be used (newline or heavily quoted string)
+func wantHeredoc(s string) bool {
+	return envutils.DynatraceHeredoc.Get() && (strings.Contains(s, "\n") || strings.Count(s, `"`) > 3)
+}
+
+func heredocBody(body string, indent string) string {
+	return "<<-EOT\n" + indent + "  " + finalizeString(body, indent) + "\n" + indent + "EOT"
```
arthurpitman, line 87: nit: I think we should take the opportunity to rename `finalizeString` and document what it does.

Kirdock: Done and renamed to `escapeHCLTemplate`

arthurpitman (approved):

PR #1245, opened 2026-09-11
fix(synthetic): Tighten dynatrace_browser_monitor validation

d0weinberger:
#### **Why** this PR?
`dynatrace_browser_monitor` resources with an invalid `validation` block are not directly rejected by the API. They get internally migrated to use Settings 2.0, and this process could fail. Failure is signaled by a tag added to the resource, as you can see in the following `terraform plan` output which aims to remove this tag again.

image

We can already know that the block is going to be invalid in the Terraform provider, and therefore we should validate and fail early.

#### **What** has changed?
When validation type is `text_match` or `content_type`, then `match` must be set. This is now validated. Also, `text_match` validation type is now documented.

#### **How** does it do it?
Added a precondition check, which fails the plan if `match` is not set when required.

image

#### How is it **tested**?
Manual tests

Fail early on invalid configurations instead of finding out later that the migration to Settings 2.0 failed.

**Issue:** PS-50447


dynatrace/api/v1/config/synthetic/monitors/browser/settings/event/validation.go
```diff
@@ -127,14 +129,23 @@ func (me *Validation) UnmarshalHCL(decoder hcl.Decoder) error {
 	return nil
 }
 
-// ValidationType The goal of the validation. `content_match` (check page for the specific content. Not allowed for validation inside of wait condition), `element_match` (check page for the specific element)
+func (me *Validation) HandlePreconditions() error {
+	if (me.Type == ValidationTypes.ContentMatch || me.Type == ValidationTypes.TextMatch) && len(me.Match) == 0 {
+		return fmt.Errorf("'match' must be a non-empty string when 'type' is '%s'", me.Type)
+	}
+	return nil
+}
```
Kirdock, line 137: Would empty actually work on the API side?
If nil, we could set it to an empty string if not set, when the condition for `match` is met.

d0weinberger: No, I tried it, an empty string leads to the same error.

Kirdock (approved):

PR #1247, opened 2026-09-15
feat: export OpenPipeline resources with admin-access

Kirdock:
#### **Why** this PR?
OpenPipeline resources in Terraform currently operate within strict owner-scoped boundaries, preventing administrators from reading or managing configurations across different owners.

#### **What** has changed?
OpenPipeline configs of different owners can now be exported

#### **How** does it do it?
By introducing an `-admin-access` flag to the export functionality.
By setting an `SupportsAdminAccess` option for all related resources. Only OpenPipeline configs (except routing) should be exported via adminAccess.
Also slightly changed the `List` logic to not consider the insertAfter and insertBefore fields for adminAccess resources, since they don't support ordering.

Regarding the implementation of using the context for the admin access flag:
Admin-access-during-export is request-scoped, optional, cross-cutting, and irrelevant to nearly the entire stack. That is the textbook legitimate use of a context value. That's the reason for not adjusting and updating the GET and LIST signature, as this would lead to modifying each and every existing resource

#### How is it **tested**?
Manual tests

They can now export OpenPipeline configs of different owners.

**Issue:** PS-45537

Note: Current failing tests are fixed in https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/1248 and not related to this PR


dynatrace/settings/services/settings20/service.go
```diff
@@ -82,8 +84,9 @@ func (me *service[T]) LegacyID() func(id string) string {
 func (me *service[T]) Get(ctx context.Context, id string, v T) error {
 	var err error
 	var settingsObject SettingsObject
+	adminAccess := me.options != nil && me.options.SupportsAdminAccess && context2.GetAdminAccess(ctx)
 
-	req := me.client.Get(ctx, fmt.Sprintf("/api/v2/settings/objects/%s", url.PathEscape(id))).Expect(200)
+	req := me.client.Get(ctx, fmt.Sprintf("/api/v2/settings/objects/%s?adminAccess=%s", url.PathEscape(id), strconv.FormatBool(adminAccess))).Expect(200)
```
arthurpitman: I guess this would also work:
```suggestion
	req := me.client.Get(ctx, fmt.Sprintf("/api/v2/settings/objects/%s?adminAccess=%t", url.PathEscape(id), adminAccess)).Expect(200)
```

arthurpitman (dismissed):

arthurpitman (approved):

PR #1248, opened 2026-09-15
test: make integration tests more stable

Kirdock:
#### **Why** this PR?
We have some flaky tests due to uniqueness constraints

#### **What** has changed?
Tests should not fail anymore if they run in parallel or if there is a missing cleanup

#### **How** does it do it?
Due to concurrency issues or missing cleanups, some configs can't be redeployed. Therefore, we add a unique identifier to the affected tests, so that they won't run into uniqueness constraints. Affected resources
- dynatrace_user_session_metrics
- dynatrace_user_action_metrics
- dynatrace_custom_units
- dynatrace_remote_environments
- dynatrace_web_app_beacon_origins

Also:
- `dynatrace_gcp_connection`: Uses `SVC:com.dynatrace.openpipeline` instead of `SVC:com.dynatrace.da`, as this is the supported use case via the UI. Other consumers aren't displayed

#### How is it **tested**?
Tests are less flaky

N/A

**Issue:** NOISSUE


dynatrace/api/builtin/remote/environment/testdata/terraform/example_a.tf
```diff
@@ -1,6 +1,6 @@
-resource "dynatrace_remote_environments" "#name#" {
-  name          = "TerraformExample"
+resource "dynatrace_remote_environments" "env" {
+  name          = "#name#"
   network_scope = "EXTERNAL"
   token         = "################"
-  uri           = "https://terraformexample.live.dynatrace.com"
+  uri           = "https://example_#name#.live.dynatrace.com"
```
arthurpitman: nit:
```suggestion
  uri           = "https://example-#name#.live.dynatrace.com"
```

arthurpitman (dismissed):

arthurpitman (approved):

PR #1251, opened 2026-09-22
feat: resource updates 346

Kirdock:
#### **Why** this PR?
Some APIs have changed with Dynatrace version 346. Therefore, the Terraform resources also need updating.

#### **What** has changed?
dynatrace_openpipeline_v2_*_dataforwarding:
- New in-development vendor type "otlp".
- New field "otlp_connection", containing a "connection_id" field.
- Fields that can't be set for "otlp": bulk_pattern, bulk_size
- Fields that must be set for "otlp": otlp_connection

Documentation and fixes for conditional field handling

#### **How** does it do it?

#### How is it **tested**?

**Issue:** PS-51053


dynatrace/api/builtin/openpipeline/bizevents/dataforwarding/settings/settings.go
```diff
@@ -175,12 +185,27 @@ func (me *Settings) HandlePreconditions() error {
...
 	if (me.GcpConnection == nil) && (string(me.CloudVendorType) == "gcp") {
 		return fmt.Errorf("'gcp_connection' must be specified when 'cloud_vendor_type' is set to 'gcp'; got 'cloud_vendor_type'='%v'", me.CloudVendorType)
 	}
+	if (me.OtlpConnection != nil) && (string(me.CloudVendorType) != "otlp") {
+		return fmt.Errorf("'otlp_connection' must not be specified unless 'cloud_vendor_type' is set to 'otlp'; got 'cloud_vendor_type'='%v'", me.CloudVendorType)
+	}
+	if (me.OtlpConnection == nil) && (string(me.CloudVendorType) == "otlp") {
+		return fmt.Errorf("'otlp_connection' must be specified when 'cloud_vendor_type' is set to 'otlp'; got 'cloud_vendor_type'='%v'", me.CloudVendorType)
+	}
 	// ---- BuiltinIngestSources []string -> {"preconditions":[{"expectedValue":"raw","property":"dataForwardingType","type":"EQUALS"}],"type":"AND"}
```
d0weinberger, line 209: Would be worth investigating why these preconditions are not generated

d0weinberger (approved):

dynatrace/api/builtin/openpipeline/bizevents/dataforwarding/settings/settings.go
Kirdock: I mean, the why, as we know, is that this is about a conditional array/slice, which we don't support yet

PR #1254, opened 2026-09-22
fix(iam): keep exporting environment policy bindings when one environment is not accessible

thiagogabriell:
#### **Why** this PR?

Exporting `dynatrace_iam_policy_bindings_v2` silently returns only the account level bindings when the account contains an environment the configured credentials are not allowed to read. Every environment level binding is lost, for all environments, and nothing is logged or returned as an error, so the export looks successful and the result is quietly incomplete.

This is easy to hit in practice: accounts often hold environments that a given OAuth client is not entitled to, and the export walks all of them regardless of which environment is being exported.

#### **What** has changed?

`FetchEnvironmentBindings` no longer aborts the whole iteration when a single environment cannot be read. It skips that environment and keeps going.

#### **How** does it do it?

The function loops over all environment IDs of the account and calls `/iam/v1/repo/environment/{id}/bindings` for each one. Both the request error and the unmarshal error used `return`, which ends the goroutine and therefore the entire loop, including the environments that had not been visited yet. Those two statements are now `continue`, so a single unreadable environment only costs that environment.

#### How is it **tested**?

Built two binaries from the same commit, one with and one without the change, and ran both against the same account, which has five environments, one of them not readable by the credentials in use:

```
without the change:  Downloading "dynatrace_iam_policy_bindings_v2" Count:  16
with the change:     Downloading "dynatrace_iam_policy_bindings_v2" Count:  51
```

The 35 additional stubs are the environment level bindings that were previously dropped. The exported files were then compared against the bindings reported by `/iam/v1/repo/environment/{id}/bindings/groups/{group}` for every group in the account, and they match exactly, with no missing and no extra group.

`go build ./...` and `go vet ./dynatrace/api/iam/...` both pass. The package has no unit tests to extend.

Exports that previously produced only account level policy bindings now also produce the environment level ones. Users who relied on the incomplete output will see additional `dynatrace_iam_policy_bindings_v2` resources appear, which is the configuration that already exists in their account. No behaviour changes for accounts where every environment is readable.

One related observation, left as is because it looks like a deliberate scope decision rather than a defect: even with this fix, the export returns bindings for every environment of the account, not only for the environment the export is pointed at. Consumers that keep one folder per environment will want to filter afterwards.

**Issue:** not filed, issues are disabled on this repository.


Kirdock:
Hi, thanks for the contribution.
Unfortunately, this is not the whole fix for this problem. The List call may show the correct amount of policy bindings, but the response is only used for getting the IDs of the related data, which are provided to the `Get` function of the policy_bindings resource. The initial count may show e.g., "5", but only 1 or no bindings are downloaded. In case there are binding details like parameters, boundaries, or metadata, this one is still not downloaded and errors out.

In here, there is another point where it errors if the permissions for a single environment are missing
https://github.com/dynatrace-oss/terraform-provider-dynatrace/blob/4d07463eef765542f42e2073878dbe3ba3ef5656/dynatrace/api/iam/v2bindings/service.go#L149-L152

Could you also take care of that? Otherwise, it's just partly fixed.

thiagogabriell:
Thanks for the review, that makes sense. I pushed a second commit that tries to cover the two cases you pointed at, but please read it as a proposal rather than a finished answer - you know the intended behaviour of these resources better than I do.

What it does:

- `Get` no longer skips a policy that appears in `policyUuids` without an entry in `bindingsDetails`. The policy is kept with its uuid.
- `resolvePolicies` no longer fails the whole binding when the level of a policy cannot be determined. The level lookup skips an environment it is not allowed to read, and a policy that stays unresolved falls back to its bare uuid. If the configuration already refers to it in the concatenated form, that spelling is kept, so the fallback does not introduce a permanent diff.

How I checked it, in case it is useful:

- Unit test with a stubbed IAM client for both cases. Without the change it fails with the 403 you described.
- End to end with the provider binary against a stub API: one account, two environments, one of them answering 403, and a group bound to three policies (one resolvable carrying boundaries, one unresolvable carrying parameters, one listed without details). Before the change the export reports the bindings, writes no file at all and still exits successfully. After it, both bindings are written with all three policies, boundaries and parameters intact.
- Against a real account with five environments, one of them not readable: 38 -> 45 bindings and 78 -> 88 policies, and the bindings that already worked come out identical.

Two things I would rather leave to you:

- The bare uuid fallback is the least surprising option I could find, but if you prefer an explicit error, or a warning, for a policy whose level cannot be resolved, I am happy to change it.
- `FetchAccountBindings` has the same shape of silent return on error. I left it out to keep this PR focused - say the word if you want it covered here too.

dynatrace/api/iam/v2bindings/service.go
```diff
@@ -143,7 +156,9 @@ func (me *BindingServiceClient) Get(ctx context.Context, id string, v *bindings.
 		filteredBindings := getFilteredBindings(policyBindings.BindingsDetails, policyID)
 
 		if len(filteredBindings) == 0 {
-			continue
+			// The policy is bound to the group, the response just carries no details
+			// for it. Dropping it here would remove the binding from the state.
+			filteredBindings = []BindingDetails{{PolicyUUID: policyID}}
```
Kirdock: This case would also mean that the API returned an incorrect result.
The `policyUuids` array contains a unique set of policy UUIDs in `bindingsDetails`. If something is not in `bindingsDetails` but in `policyUuids`, this means the API returned an incorrect result, which I would say we can ignore.

Kirdock:
Actually, maybe we can get rid of the whole resolvePolicy thing. Then we wouldn't need all the error handling, because we should already have all the needed information. I'll try something out and keep you updated

dynatrace/api/iam/policies/policy_level.go
Kirdock: This would completely hide permission errors, which should not happen, in my opinion.
Could you maybe do it like this?
```go
var errs []error
	for _, environmentID := range environmentIDs {
		if exists, name, err = CheckPolicyExists(ctx, client, "environment", environmentID, uuid); err != nil {
			errs = append(errs, err)
			continue
		}
		if exists {
			return "environment", environmentID, name, nil
		}
	}
	err = errors.Join(errs...)
	if err != nil {
		return "", "", name, err
	}
```

With this change, we also don't need any other modifications in the other files, other than the `continue`.

And also: The documentation/comments are a bit overkill. Can you reduce them a bit?

dynatrace/api/iam/v2bindings/service.go
Kirdock: With this, it's actually bug on the API side, as the policyUUIDs returned from the API are a distinct set of the used policies in the policy bindings, so skipping that would be ok in my opinion and we can keep the previous implementation.

dynatrace/api/iam/v2bindings/service.go
```diff
@@ -174,11 +189,15 @@ func (me *BindingServiceClient) resolvePolicies(ctx context.Context, uuid string
...
+		// A policy whose level cannot be determined - it may live in an environment
+		// these credentials cannot read - is still bound to the group. The resource
+		// accepts the bare UUID as well, so fall back to it instead of failing the
+		// whole binding.
+		policyID := unresolvedPolicyID(uuid, existingPolicies)
+		if levelType, levelID, _, err := policies.ResolvePolicyLevel(ctx, me.client, uuid); err == nil {
+			policyID = deductPolicyID(uuid, levelType, levelID, existingPolicies)
 		}
-		policy := &bindings.Policy{ID: deductPolicyID(uuid, levelType, levelID, existingPolicies)}
+		policy := &bindings.Policy{ID: policyID}
```
Kirdock: With the suggestion above, we don't need to handle this case.
I would prefer no resource to a resource that is "partly" correct/filled.

dynatrace/api/iam/v2bindings/service_unit_test.go
```diff
@@ -0,0 +1,174 @@
...
+	policyNoLevel   = "33333333-3333-3333-3333-333333333333"
+	policyNoDetails = "55555555-5555-5555-5555-555555555555"
+)
+
+// fakeIAMClient answers the handful of endpoints the binding service calls.
+// Any environment other than readableEnv responds 403, which is what happens
+// when the configured credentials do not cover every environment of an account.
+type fakeIAMClient struct {
+	calls []string
+}
```
Kirdock: We don't need any new fakeIAMClient here, as this is a duplicated implementation as we already have in `dynatrace/testing/mockiamclient.go`. Could you use this one instead?
Also for the body: no need to do the whole JSON payload manually here, as we already have the Go struct for them. We can just use the structs and marshal them.

dynatrace/api/iam/v2bindings/service_unit_test.go
Kirdock: Could you align with our testing tools and use testify?

Kirdock (changes requested):

PR #1255, opened 2026-09-23
fix(test): frequency = 5 for all dynatrace_http_monitor examples

d0weinberger:
#### **Why** this PR?
We received the feedback that a frequency value of 1 is misleading in the examples. The documentation for the endpoint even states:

> You can use one of the following values: 5, 10, 15, 30, and 60

#### **What** has changed?
Changed the frequency in the examples to a value that makes sense, i.e. 5

#### How is it **tested**?
The tests are the point.

The documentation will reflect appropriate values

**Issue:** N/A


Kirdock (approved):
`dynatrace_browser_monitor` also seems to have correct values in the examples, so also fine for this 👍

